PR #28, opened 2015-12-15
first stab at putting protocol handlers in all processes

the8472:
it's not quite done, but i'd like some feedback. i had to untangle the gui and configuration concerns a bit to make it work in the child process.

PR #29, opened 2015-12-15
Add pin function

jedahan:
Please do not merge this, I just opened this to start a discussion. We could come up with a much better ux for pinning / unpinning I think:

1. make the current button act as pin/unpin, with ON/OFF being replaced with a pin icon. Right click the button to turn on and off the redirect, which would color/uncolor the logo.

2. create a second button just dedicated to pinning

3. make the button just pin/unpin, and ask people to enable/disable the addon like they do all other addons, in the addons preference pane.

All of these require better code:

1. Convert open new tab to ajax call that handles success/failure

2. Add unpin functionality

And probably want to not hard-code the :5001 port for the daemon (make it the default, allow for override like for the gateway)

Anyway, is this a feature people would like me to work on more?

### DIFF
--- a/lib/gateways.js
+++ b/lib/gateways.js
@@ -8,6 +8,7 @@ var cm = require('sdk/context-menu');
 var prefs = require('sdk/simple-prefs').prefs;
 var ioservice = Cc['@mozilla.org/network/io-service;1'].getService(Ci.nsIIOService);
 var gui = require('./gui.js');
+var tabs = require("sdk/tabs");
 
 const IPFS_RESOURCE = /^https?:\/\/[^\/]+\/ip(f|n)s\//;
 var PUBLIC_GATEWAY_HOSTS; // getPublicGatewayHostsRegex()
@@ -95,6 +96,13 @@ function disableHttpGatewayRedirect(button) {
   if (button) button.state(button, gui.toggleStateDisabled);
 }
 
+function pin(address) {
+  let IPFS_API_URI = 'http://' + prefs.customGatewayHost + ':5001/api/v0';
+  let uri = ioservice.newURI(IPFS_API_URI+'/pin/add?arg='+address, null, null);
+  tabs.open(uri.spec);
+}
+
+exports.pin = pin;
 exports.ipfsRequestObserver = ipfsRequestObserver;
 exports.enableHttpGatewayRedirect = enableHttpGatewayRedirect;
 exports.disableHttpGatewayRedirect = disableHttpGatewayRedirect;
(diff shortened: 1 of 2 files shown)


lib/gateways.js
```diff
@@ -8,6 +8,7 @@ var cm = require('sdk/context-menu');
 var prefs = require('sdk/simple-prefs').prefs;
 var ioservice = Cc['@mozilla.org/network/io-service;1'].getService(Ci.nsIIOService);
 var gui = require('./gui.js');
+var tabs = require("sdk/tabs");
```
lidel: We probably want single quotes here:
https://travis-ci.org/lidel/ipfs-firefox-addon/jobs/96923547#L790 
:)

lidel:
Thanks! Looks very good, feel free to work on it :+1: 

My thoughts: 
- API port should be configurable (just in case someone runs it on something else than `5001`)
- I think pinning via right-click context menu is fine for now (we can improve it later, see below)
  - Is it possible to pin recursively? 
- I agree we should improve UX related to the toolbar button:
  - Right now it simply enables/disables forwarding to local gateway. 
  - My plan was to change it to display a menu (on click) with multiple actions instead (perhaps we could simply attach items from context menu?)

lidel:
FYI I merged other pull request (#28) and you probably will need to rebase your changes on top of current master.

lib/gateways.js
```diff
@@ -95,6 +96,13 @@ function disableHttpGatewayRedirect(button) {
   if (button) button.state(button, gui.toggleStateDisabled);
 }
 
+function pin(address) {
+  let IPFS_API_URI = 'http://' + prefs.customGatewayHost + ':5001/api/v0';
+  let uri = ioservice.newURI(IPFS_API_URI+'/pin/add?arg='+address, null, null);
+  tabs.open(uri.spec);
```
lidel: I think opening this in a [Sidebar](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/ui#Sidebar) would be less-intrusive for a user (as a quick UX fix). 

Ideally, we would want to parse the response and display some kind of [notification](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/notifications) instead.

lidel:
As for the ToggleButton improvements, [this](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/panel#Attaching_panels_to_buttons) is what I have in mind:



The popup panel would have all items from the context menu (copy link, pin resource etc) + on/off switch for entire plugin (current onClick behaviour).

lidel:
Closing this one, will continue in #30

PR #30, opened 2015-12-18
Add pin function

jedahan:
Continuation of  #29 in a new branch. This one uses the notifications api, and should be safe to merge after some testing. I think gui improvements should be in a separate pull request/issue...


lidel:
Looks good, will merge and do some tweaks.

PR #35, opened 2015-12-21
WIP: rewriting ipfs uris to fs: prefix and linkifying inline urls

the8472:
Very experimental for now, but I can rewrite pretty much everything to `fs:/` URIs while doing http in the background.

### DIFF
--- a/lib/redirects.js
+++ b/lib/redirects.js
@@ -2,58 +2,90 @@
 
 const gw = require('./gateways');
 
-var {
-  Cc, Ci
-} = require('chrome');
+var { Cc, Ci, Cr } = require('chrome');
+
+const protocols = require("./protocols.js")
 
 var ioservice = Cc['@mozilla.org/network/io-service;1'].getService(Ci.nsIIOService);
+var events = require("sdk/system/events");
 
-var ipfsRequestObserver = {
-  observe: function(subject, topic, data) { // jshint unused:false
-    if (topic == 'http-on-modify-request') {
-      let channel = subject.QueryInterface(Ci.nsIHttpChannel);
-      let httpUrl = channel.URI.spec;
-      if (httpUrl.match(gw.publicHosts()) && httpUrl.match(gw.IPFS_RESOURCE)) {
-        channel.setRequestHeader('x-ipfs-firefox-addon', 'true', false);
-        if (gw.isEnabled()) {
-          //console.info('Detected HTTP request to the public gateway: ' + channel.URI.spec);
-          let uri = ioservice.newURI(httpUrl.replace(gw.publicHosts(), gw.customUri().spec), null, null);
-          //console.info('Redirecting to custom gateway: ' + uri.spec);
-          channel.redirectTo(uri);
-        }
+const listener = function(event) {
+  //console.log("obs", event);
+  let {subject, type, data} = event;
+
+  if (type == 'http-on-modify-request') {
+    if(!gw.isEnabled() && prefs.rewriteAs == "gateway")
+      return;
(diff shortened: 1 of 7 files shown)


lib/redirects.js
```diff
@@ -2,58 +2,90 @@
...
+    let isGatewayRequest = !!(httpUrl.match(gw.publicHosts()) && httpUrl.match(gw.IPFS_RESOURCE));
+    // TODO: handle gateway -> local redirects
+    if(!isGatewayRequest)
+      return;
+    let ipfsURI = ioservice.newURI(protocols.fsScheme + ":" + channel.URI.path, null, null);
+    channel.setRequestHeader('x-ipfs-firefox-addon', 'true', false);
+
+    //console.log("comp", ipfsURI.spec, channel.originalURI.spec, ipfsURI.spec == channel.originalURI.spec)
+
+    if(prefs.useCustomGateway || !channel.originalURI || channel.originalURI.spec != ipfsURI.spec ) {
```
lidel: `redirects.js:30`: we probably want to use `gw.isEnabled()` here as  `ReferenceError: prefs is not defined` (I got this error while testing b5d785b)

package.json
```diff
@@ -40,6 +40,22 @@
...
       "value": 5001
+    },
+    {
+      "name": "rewriteAs",
+      "type": "menulist",
+      "title": "Rewrite IPFS URIs in the UI as",
+      "value": "gateway",
+      "options": [
+        {
+          "value": "gateway",
```
lidel: I feel a better value here would be "publicGateway" so that there is no overlap between preference name and its value.

lidel:
I understand it is work in progress and some things do not work yet, but it looks quite cool.
I really like the canonization idea (forwarding `ipfs:`, `ipns:` and other custom protocol handlers to `fs:` one).

Keep up good work :+1: 

Some random bugs:
- when `rewriteAs=fs`:
  - `ipfs:///QmTAsnXoWmLZQEpvyZscrReFzqxP3pvULfGVgpJuayrp1w` redirects to `fs:/ipfs/QmTAsnXoWmLZQEpvyZscrReFzqxP3pvULfGVgpJuayrp1w` (OK) but  `ipfs://QmTAsnXoWmLZQEpvyZscrReFzqxP3pvULfGVgpJuayrp1w` redirects to `fs:/ipfs/` (BUG)
- when `rewriteAs=gateway` (Rewrite IPFS URIs in the UI as Public Gateway) the redirect to the custom gateway does not seem to work  -- it is probably caused by https://github.com/lidel/ipfs-firefox-addon/pull/35#discussion-diff-48364269

lidel:
@the8472   [noot noot :penguin:](https://yt.aergia.eu/#v=a4VvRWTD3Ok), are you still working on this? A lot of code changed since December :cry:
I'll understand if you lost motivation, just let me know so that I can move https://github.com/lidel/ipfs-firefox-addon/issues/48  forward without duplicating our efforts.

the8472:
Yeah, still on it.
Specifically I have a cut-down version that only works in non-e10s, doesn't try to do any of the URI fixups and doesn't rewrite http -> fs either. In other words it only keeps the `fs:` scheme if you're opening a `fs:` uri.

If you want to I can make a new PR for that. e10s support and the other things could be built on top of it.

lidel:
I could merge new PR with cut-down version, but only if it comes with e10 support (otherwise we have a regression).

If it is not something you can easily add, then we can wait for the full solution, no problem. :+1:

the8472:
Obsoleted by #70

Issue #27, opened 2015-12-14
Linkify ipfs-ish text


the8472:
options to turn plain /ipfs/ and ipfs:// text into links and context menu entries to open any highlighted multihash as ipfs links would be nice.

lidel:
Sounds useful  :+1: 

My thinking:
- A new context menu item "Open IPFS resource" visible when there is a text selection that match IPFS resource regex 
- Not sure about turning text into links. I think it could be done, but modifying pages on the fly should be configurable via checkbox on the Preferences screen and probably disabled by default.

If there is no PR with this I'll try to add it in my spare time.. some day. ;-)

the8472:
Yeah, modification of the DOM should be optional.

PR #39, opened 2015-12-28
plaintext linkification

the8472:
splitting out the linkification work from the url canonization since this pretty much works while the latter is a hairy mess.

### DIFF
--- a/data/test.html
+++ b/data/test.html
@@ -1,0 +1,15 @@
+<!doctype html>
+<!-- open as resource://ipfs-firefox-addon-at-lidel-dot-org/data/test.html -->
+<html>
+<body>
+  <p>
+    ipfs:/QmTAsnXoWmLZQEpvyZscrReFzqxP3pvULfGVgpJuayrp1w<br>
+    ipfs:/ipfs/QmTAsnXoWmLZQEpvyZscrReFzqxP3pvULfGVgpJuayrp1w<br>
+    fs:/ipfs/QmTAsnXoWmLZQEpvyZscrReFzqxP3pvULfGVgpJuayrp1w<br>
+    /ipfs/QmTAsnXoWmLZQEpvyZscrReFzqxP3pvULfGVgpJuayrp1w<br>
+  </p>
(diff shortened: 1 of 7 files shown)


the8472:
implements #27

data/test.html
```diff
@@ -0,0 +1,15 @@
+<!doctype html>
+<!-- open as resource://ipfs-firefox-addon-at-lidel-dot-org/data/test.html -->
```
lidel: Cool! It probably should be used in a dedicated  test suite (`tests/test-linkify.js`).

So far I did manual test and got weird results: 

0. Run via `jpm -b vendor/firefox-45.0a2/firefox run`
1. I enabled at addon's `about:addons` screen: 
2016-01-01-182616_256x49_scrot
2. Visited `resource://ipfs-firefox-addon-at-lidel-dot-org/data/test.html`
3. Text links were not converted into `<a>` tags. The only change I was able to spot was the `href` of the "abs link", which changed from 
`fs:/` to `https://ipfs.io/`

I did not dig into the underlying code: perhaps it is still work in progress?
Or did I found a bug? :-)

lib/rewrite-pages.js
```diff
@@ -0,0 +1,214 @@
...
+  let obs = new window.MutationObserver(mutationCallback);
+  obs.observe(document, {
+    characterData: true,
+    subtree: true,
+    childList: true
+  });
+
+};
+
+events.on('content-document-global-created', documentObserver);
```
lidel: It may be a better idea to on/off entire listener based on checkbox state.
Perhaps something along these lines:

```javascript
gw.onChange(() => {
  if (gw.linkify) {
    events.on('content-document-global-created', documentObserver);
  } else {
    events.off('content-document-global-created', documentObserver);
  }
});
gw.reload();
```

the8472:
>  Or did I found a bug? :-)

yeah, missed a change when rebasing. should work now.

I also added a test.

> It may be a better idea to on/off entire listener based on checkbox state.

Eh, it bails out fairly early: https://github.com/lidel/ipfs-firefox-addon/pull/39/files#diff-d5184922475dad24022cebfa449bba33R166

lidel:
Thanks. Now works as advertised! :ok_hand: 

(did some cleanup in a837132 & a74d4f7)

PR #70, opened 2016-02-11
support for fs: URIs without redirecting to http:

the8472:
It works with and without e10s for basic resource loading and I also tested it against the case in #3. But it most likely needs some further testing to weed out edge-cases, so i've marked it as default-off and experimental.

Basically, from firefox's view this all runs under CORS policy because we're running the HTTP requests with the principals of the fs:/ document instead of implementing our own channels which then delegate to a privileged HTTP channel
Getting the URLs a tiny bit wrong causes redirects from the gateways which then can then void the CORS permissions under circumstances.

Implementing as a custom channel wrapper around http channels might be more robust wrt. CORS but currently is way too troublesome under e10s until the FF devs improve support (or documentation) there.

### DIFF
--- a/lib/request-fix.js
+++ b/lib/request-fix.js
@@ -1,0 +1,49 @@
+'use strict'
+
+const {Ci} = require('chrome')
+const events = require('sdk/system/events')
+const gw = require('./gateways.js')
+
+/*
+const catman = Cc['@mozilla.org/categorymanager;1'].getService(CI.nsICategoryManager)
+// category ("net-channel-event-sinks")
+catman.addCategoryEntry(in string aCategory, in string aEntry, in string aValue, in boolean aPersist, in boolean aReplace)
+*/
+
+function fixupRequest (e) {
+  if (e.type !== 'http-on-modify-request') {
+    return
+  }
+  if (!gw.fsUris) {
+    return
+  }
+  const channel = e.subject.QueryInterface(Ci.nsIHttpChannel)
+
+  channel.QueryInterface(Ci.nsIHttpChannelInternal)
+  channel.QueryInterface(Ci.nsIPropertyBag2)
+  channel.QueryInterface(Ci.nsIPropertyBag)
+
+  let isIpfsReq = null
+  try {
+    isIpfsReq = channel.hasKey('ipfs-uri')
+  } catch (e) {
+    // console.log(e)
+  }
+
+  if (isIpfsReq && channel.originalURI.scheme === 'fs') {
+    /*
+    // TODO: investigate effects of the following flags
+    // cookies make no sense in the ipfs context, we don't want to carry different gateway cookies into the page
+    channel.loadFlags |= Ci.nsIRequest.LOAD_ANONYMOUS
+    // should only do that for /ipfs/ paths since those are stable
+    channel.loadFlags |= Ci.nsIRequest.VALIDATE_NEVER
+    */
+
+    // prevent redirects from replacing the effective URI
+    channel.loadFlags &= ~Ci.nsIChannel.LOAD_REPLACE
+
+  }
+
+}
+
+events.on('http-on-modify-request', fixupRequest)
--- a/test/test-canonical-fs-uris.js
+++ b/test/test-canonical-fs-uris.js
@@ -1,0 +1,79 @@
+'use strict'
+
+const tabs = require('sdk/tabs')
+const protocols = require('../lib/protocols.js')
+const fsFactory = protocols.fs
+
+protocols.register()
+
+const fs = fsFactory.createInstance()
+const gw = require('../lib/gateways.js')
+const self = require('sdk/self')
+const testpage = self.data.url('linkify-demo.html')
+const sripage = 'fs:/ipfs/QmSrCRJmzE4zE1nAfWPbzVfanKQNBhp7ZWmMnEdbiLvYNh/mdown#sample.md'
+const parent = require('sdk/remote/parent')
+
+const childMain = require.resolve('../lib/child-main.js')
+
+parent.remoteRequire(childMain)
+
+const {Cc, Ci} = require('chrome')
+const ioservice = Cc['@mozilla.org/network/io-service;1'].getService(Ci.nsIIOService)
+
+ioservice.newURI('fs:/ipns/foo', null, null)
+
+exports['test newURI'] = function (assert) {
+  require('sdk/simple-prefs').prefs.fsUris = true
+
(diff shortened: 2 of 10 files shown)


the8472:
Idk why the tests fail on travis. Running `jpm test` passes for me.

lidel:
Thank you! I'll do my best to review and merge it this weekend.  

For now I did a quick check and tests pass for me too. 

It is probably a long shot, but maybe nodejs  at Travis is too old?
Try updating `.travis.yml` in a commit to this branch (4.1 seems to be the [max](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Provided-Node.js-Versions)):

```diff
node_js:
-    - "0.12"
+    - "4.1"
```

and let's see if it helps.

the8472:
added that.

Another thing of note, to make jpm happy locally I had to change the `.jpmignore` file. Does it work as it is in the git tree for you?

test/test-canonical-fs-uris.js
```diff
@@ -0,0 +1,79 @@
...
+// https://github.com/lidel/ipfs-firefox-addon/issues/3
+exports['test subresource loading'] = function (assert, done) {
+  require('sdk/simple-prefs').prefs.fsUris = true
+  gw.redirectEnabled = false
+
+  tabs.open({
+    url: testpage,
+    onReady: (tab) => {
+
+      // first load somehow doesn't have protocol handlers registered. so load resource:// first, then redirect to fs:/ page
```
lidel: Yeah, this is the same bug as in https://github.com/lidel/ipfs-firefox-addon/issues/33#issuecomment-166139377 
We probably need to do this workaround for now, hopefully it will be fixed in WebExtensions.

lidel:
- `jpmignore` works fine for me -- I've built `.xpi` and contents included only whitelisted files


- Travis still fails, so it is not nodejs. Perhaps Travis is slower than our machines, and asynchronous callbacks that are fired after `prefs` change collide with URI tests from `test/test-protocols.js`.. or we have some race condition and under slow runner `prefs.fsUris` is sometimes left as `true`? No idea.

    [This](https://travis-ci.org/lidel/ipfs-firefox-addon/jobs/108628693#L1025) seems to confirm that `fs:` is still present in address bar, while it is expected to be changed into http link.

   Perhaps you could make these tests independent of the default value, and explicitly disable `fsUris` inside of each one (and perhaps add another set for `prefs.fsUris = true`):

 ```diff
 + const { prefs } = require('sdk/simple-prefs')
 
 exports['test newURI(fs:ipfs/<path>)'] = function (assert) {
 +  prefs.fsUris = false
   var newURI = fsHandler.newURI('fs:ipfs/' + ipfsPath, 'utf8', null)
   assert.equal(newURI.spec, pubGwUri.spec + 'ipfs/' + ipfsPath, 'newURI spec')
 }
 ```

  Sadly I don't have any better idea (feel free to add as many commits as you need).

the8472:
> running tests on Firefox 44.0.1/Gecko 44.0.1

That seems to be latest stable. Is there no nightly?

lidel:
No nightly. Currently tests are run against  latest `stable`,  `beta`, and every `.0` release since `43.0`:
https://travis-ci.org/lidel/ipfs-firefox-addon/builds/108628692

the8472:
Well, tests pass for me now. I blame whatever doesn't work on travis

lidel:
I did some manual tests locally under `44`, `45` and `47a1`.

Indeed super weird: automated tests pass for me too, but when I tried to actually use it (enabled `fsUris` on the _Preferences_ screen) it does not seem to work:  i am always redirected to `http://127.0.0.1:8080/(...)` (or `http://ipfs.io/(..)` if redirect was disabled).



Could you  check if `fs:/ipfs/QmYHNYAaYK5hm3ZhZFx5W9H6xydKDGimjdgJMrMSdnctEm` works for you?

the8472:
It displays the fs:/ uri for me in nightly and loads the cat. Works both with local and ipfs.io gateway.

I have had some issues with jpm/npm modules though, maybe you're experiencing the same? Clearing it the generated folders and doing an npm update solved them for me.

lidel:
Removed `node_modules` and installed everything again via `npm install`. 

My versions:
```
node 5.6.0
npm 3.6.0
jpm 1.0.5
```

I run local version via `node_modules/jpm/bin/jpm -b $FIREFOX_BIN run` 
Still have the same problem: `fs:` is replaced with `http:` when redirect is enabled.

Are you running clear profile without any auto-reloading addons?

the8472:
node v5.5.0
npm 3.7.1
jpm 1.0.5

I have a dev profile which I use together with `jpm post` + extension auto installer.

lidel:
Just to be sure it is something on my end, 
please try without dev profile and other addons, just dry run with `jpm -b $FIREFOX_BIN run`.

the8472:
Ok, yeah, now I'm seeing it too *sigh*

Will fix.

the8472:
should work now.

lidel:
Merged – thank you! 

Let's follow this in related issues:

- https://github.com/lidel/ipfs-firefox-addon/issues/48
- https://github.com/lidel/ipfs-firefox-addon/issues/71
- https://github.com/lidel/ipfs-firefox-addon/issues/72

lib/request-fix.js
```diff
@@ -0,0 +1,49 @@
...
+const events = require('sdk/system/events')
+const gw = require('./gateways.js')
+
+/*
+const catman = Cc['@mozilla.org/categorymanager;1'].getService(CI.nsICategoryManager)
+// category ("net-channel-event-sinks")
+catman.addCategoryEntry(in string aCategory, in string aEntry, in string aValue, in boolean aPersist, in boolean aReplace)
+*/
+
+function fixupRequest (e) {
```
lidel: @the8472  I am going over the backlog and trying to have a better understanding of this part of code.  I see this is related to [protocols.js#L121-L131](https://github.com/the8472/ipfs-firefox-addon/blob/0b085707442b45333e8adef73aa1c3728d0622ba/lib/protocols.js#L121-L131) but can't infer full picture from code alone.

Could you elaborate what is the purpose of this fixup?

the8472: I'm not entirely sure if it's still necessary.

There were some cases with redirects coming from the server which needed the LOAD_REPLACE flag to be removed. If it is not removed the new channel's URI overrides originalURI which leads to a http URI in the address bar.

lidel: While implementing #72 I did some cleanup in https://github.com/lidel/ipfs-firefox-addon/commit/c75a05d82c3896d9eba8c1627895bef800978697.

I did not notice any regressions, but would appreciate if you took a second look, perhaps I missed something.

Issue #218, opened 2017-03-19
Add support for chromium-based browsers


AlexBeLi:
Because ipfs-chrome-extension and ipfs-chrome-station don't support ipfs>=0.4.5.

lidel:
This may require some additional work due to minor incompatibilities between WebExtensions and Chrome APIs:
- https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Chrome_incompatibilities
- There is also a polyfill for standarized APIs (which is a subset of APIs provided by Firefox): https://github.com/mozilla/webextension-polyfill
- Chrome does not support `async` / `await` yet (Firefox does and I plan to use it extensively)
  In general, chrome target would probably need some kind of transpilling


Firefox is my daily driver so this has a low priority right now.

If someone wants to invest time to add `npm run build-chrome` sooner, a PR would be much welcome 🙃

mateon1:
I just fixed ipfs-chrome-station, but it might take a while for xicombd/ipfs-chrome-station#15 to get merged, and the extension updated.
For now, you can use https://github.com/mateon1/ipfs-chrome-station/tree/fix/ipfs-0.4.5
To do so, clone the repo and switch to branch `fix/ipfs-0.4.5`. Run `npm install` and `npm run build`. You can then load the `./build` folder in Chrome as an unpacked extension.

diasdavid:
Following up https://github.com/ipfs/in-web-browsers/issues/54#issuecomment-300234021, we want to have this WebExtension serving both.

Is `Only one of browser_action, page_action can be specified in chrome (we use both)` the only thing missing?

lidel:
I remember there were multiple errors about unsupported manifest keys, eg. [recently introduced](https://github.com/lidel/ipfs-firefox-addon/issues/164) `protocol_handlers`.  I've added them to first post. There is probably more.

This "write once, run everywhere" story provided by Mozilla does not pass reality check.
As soon as we use non-canonical API provided by Firefox, we are not Chrome-compatible. 
And as you can see there are discrepancies  even within APIs copied from Chrome.

I suspect we will end up with chunks of conditional code and dedicated manifest and build pipeline for each browser.

That being said, we can simplify stuff by merging "pageAction" back into "browserActions". 
One thing less to worry about.

lidel:
FYI I'll do my best in diagnosing / mitigating compatibility issues in `chromium-compat` branch this week. 
Stay tuned.  👌

lidel:
[...]

PR #250, opened 2017-05-21
Initial stab at Chromium compatibility

lidel:
This PR provides initial support for Chromium 58+ browsers (https://github.com/ipfs/in-web-browsers/issues/54, #218).

### Running in Chromium

First, build it manually:

 ```bash
npm install
npm run build
 ```
Then open up `chrome://extensions` in Chromium-based browser, enable "Developer mode", click "Load unpacked extension..." and point it at `add-on/manifest.json`.

It [should load just fine](https://blog.codinghorror.com/content/images/uploads/2007/03/6a0120a85dcdae970b0128776ff992970c-pi.png), with some minor warnings (see _known issues_ below):

>  2017-05-13-211205_897x506_scrot

### Brief list of issues fixed in this PR

- Chromium does not have `browser` object and requires a polyfill for standardized APIs (which is a subset of APIs provided by Firefox): https://github.com/mozilla/webextension-polyfill 
- Only one of `browser_action`, `page_action` can be specified in Chromium 
- Chromium has some additional restrictions regarding locale files which needs to be addressed (fixed by removing `-` from key names)
- Chromium complains about some i18 placeholders  present in local files 
- Chromium does not support "alpha/beta" suffix in `version` in manifest 
- Chromium does not accept SVG as browserAction icon, PNG needs to be provided instead (https://stackoverflow.com/q/38457208, https://bugs.chromium.org/p/chromium/issues/detail?id=647182, https://bugs.chromium.org/p/chromium/issues/detail?id=29683) 
- Chromium sends `Origin: null` with XHR requests for `/api/v0/swarm/peers` which causes `403 - Forbidden` error
- Chromium does not ship with native CSS classes that are shipped with Firefox 
- Right-click "Upload to IPFS" does not work due to [Origin mismatch](https://cloud.githubusercontent.com/assets/157609/26028017/7e694404-3818-11e7-90c8-3eee78b0bbb8.png)

### Known Issues

Remaining Issues that are not covered by this PR:
- Chromium does not accept Alarm periods less than minimum of 1 minutes. This means "ipfs-api-status-update" will fire approximately every 1 minutes (instead of every 3 seconds). We can leave it that way, or rewrite API polling logic to use something else than Alarms API.
- No support for `web+foo` protocols. Chromium does not support `protocol_handlers` in manifest, this needs to be detected and manual handler needs to be installed somehow.


lidel:
Are we good to merge this? If so, hit me up with review approval.
I'd like to release it into Firefox's Development Channel as `v2.0.0alpha5`

whyrusleeping:
loading it into chrome i get:
```
Failed to load extension from: ~/gopkg/src/github.com/ipfs/ipfs-companion/add-on
Invalid placeholder errorMsg for key notify_inlineErrorMsg
```
Followed by the entire manifest.json in a little scrollybox

whyrusleeping:
chromium version: 58.0.3029.110

lidel:
@whyrusleeping  Are you sure you loaded code from `chromium-compat` branch? 
Mentioned bug was fixed in https://github.com/ipfs/ipfs-companion/pull/250/commits/6f007412093033837fddc1648bca086ee9f5c65d

whyrusleeping:
@lidel youre right, i missed the checkout a branch bit. Works! :+1: :+1:

add-on/src/lib/common.js
```diff
@@ -383,17 +414,60 @@ function updateBrowserActionBadge () {
...
+  let canvas = document.createElement('canvas')
+  let context = canvas.getContext('2d')
+  context.clearRect(0, 0, size, size)
+  context.drawImage(icon, 0, 0, size, size)
+  return context.getImageData(0, 0, size, size)
+}
+*/
+
 // OPTIONS
 // ===================================================================
```
diasdavid: Wouldn't it make more sense to modularize these sections?

diasdavid:
Ah, only see that purple merged label after. awesome!

add-on/src/lib/common.js
lidel: Yes, at some point. 
I've added those "section" comments, as `common.js` grew organically.  

I was not sure what is the best course of action for "webextension modularization", as Firefox WX ecosystem was still brand new when I started the rewrite from legacy SDK.

In Chrome world people use [ES6 modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import), but feature is not implemented in any browsers natively at this time and requires you to transpile everything to a single `.js` file ([example](https://github.com/samuelsimoes/chrome-extension-webpack-boilerplate)).

I avoided it and right now add-on code runs "as-is" (no transpilling), but I am not sure if this approach can be maintained in the long run, especially now that we support multiple browser vendors.

If you feel it is the time, create a new Issue and lcopy/link this comment there.
Alternatively, we leave it be a monolith until ES6 modules are natively supported by both Firefox and Chrome.

PR #261, opened 2017-07-01
chore(package): bump versions and ignore package-lock.json

lidel:
According to [npm help package-lock.json](https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md):

> `package-lock.json` is automatically generated for any operations where `npm` modifies either the `node_modules` tree, or `package.json`. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.
> 
> This file is intended to be committed into source repositories, and serves various purposes:
> 
>   * Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> 
>   * Provide a facility for users to "time-travel" to previous states of `node_modules` without having to commit the directory itself.
> 
>   * To facilitate greater visibility of tree changes through readable source control diffs.
> 
>   * And optimize the installation process by allowing `npm` to skip repeated metadata resolutions for previously-installed packages.

My understanding is that including it in repository should finally give us _deterministic builds_, but please someone review this before we merge it.

PS. I know that `package-lock.json` may produce errors if `node_modules` was already present until fix for https://github.com/npm/npm/issues/16839  is released, so if you experience any issues remove `node_modules` before running `npm install` for this branch.


diasdavid:
I've experienced many errors with the new npm5 which made me revert the usage of it on js-ipfs and js-ipfs-api. I do hope to use it in the future and feel free to use it for this repo, just bare in mind that there will be issues of missing npm installs and lack of proper updates until all npm5 issues are fixed. These issues will be certainly frustrating for other contributors that have less introspection of what is going on.

lidel:
I played with it a bit, and indeed, currently it is very buggy.
I merged version bumps, but I've added `package-lock.json` to  `.gitignore`  so that it is not committed by mistake.

PR #262, opened 2017-07-06
Add zh_CN locale, and an tiny fixes for en

yfdyh000:


lidel:
Thank you for your contribution!  👍 

`zh_CN` locale will be included in `v2.0.0alpha7` release

PR #277, opened 2017-09-04
Improved Automatic Mode and Options UX

### LINKED ISSUE #274: Add quick toggle for gateway redirect/all functionality
Do something similar to IPFS Station (chrome extension), and include a toggle for gateway redirect in the main interface.
Possibly add a way to temporarily toggle all functionality of the extension (globally and/or per site), similar to how ublock does it.

I find that I spend a lot of time going to the extensions page and manually disabling and then re-enabling the extension when I want to cache a hash on a public gateway.

lidel:
- Closes #274: makes it possible to manually enable/disable redirect while Automatic Mode is enabled.
Automatic Mode will toggle redirect only when API state changes between online/offline.
- Improves UX of Options (Preferences) screen by addressing concerns raised in https://github.com/ipfs/ipfs-companion/issues/274#issuecomment-326869463
- dependency bump


lidel:
I've added a quick description for every option on the _Preferences_ screen and made it look kinda decent under both Chrome and Firefox.

Please let me know if something sounds odd or could be replaced with better label/description.
Reward is one virtual cake.

> 2017-09-07-204307_983x1097_scrot

add-on/src/options/options.html
```diff
@@ -1,56 +1,167 @@
...
-
-        <p><label>Automatic Mode <small>(Redirect only when IPFS API is Online and node has peers)</small> <input type="checkbox" id="automaticMode" /></label></p>
-
+      <fieldset>
+        <legend>Gateways</legend>
+        <div>
+          <label for="publicGateways">
+            <dl>
+              <dt>Public Gateways</dt>
+              <dd>list of hostnames known for exposing IPFS resources under <code>/ipfs/</code> path</dd>
```
Vaelatern: Recommend rewording:
`<dd>List of hostnames known to expose IPFS under <code>/ipfs/</code></dd>`

lidel: 🍰

add-on/src/options/options.html
```diff
@@ -1,56 +1,167 @@
...
+            </dl>
+          </label>
+          <textarea id="publicGateways" rows="4" required placeholder="List of hostnames separated with spaces or newlines"></textarea>
+        </div>
+        <div>
+          <label for="customGatewayUrl">
+            <dl>
+              <dt>Custom Gateway</dt>
+              <dd>URL address of preferred HTTP2IPFS&nbsp;Gateway</dd>
```
lidel: > dunno if 'URL address' is correct. it's either URL or web address, I think (thx AndyS2!)

add-on/src/options/options.html
```diff
@@ -1,56 +1,167 @@
...
+              <dd>hint: this is where <code>/webui</code> lives</dd>
+            </dl>
+          </label>
+          <input type="url" id="ipfsApiUrl" inputmode="url" required pattern="https?://.+"/>
+        </div>
+        <div>
+          <label for="ipfsApiPollMs">
+            <dl>
+              <dt>Status Poll Interval</dt>
+              <dd>how often peer count is refreshed (in miliseconds)</dd>
```
lidel: from AndyS2:
> Time interval after which peer count is refreshed (in milliseconds)

lidel: from AndyS2:
> Time interval after which peer count is refreshed (in milliseconds)

add-on/src/options/options.html
lidel: from AndyS2:
> dunno if 'URL address' is correct. it's either URL or web address

add-on/src/options/options.html
lidel: from AndyS2:
> starting everything small looks a bit weird. How about 'A list of hostnames ...' and changing everything like that?

add-on/src/popup/quick-upload.html
```diff
@@ -1,5 +1,4 @@
-<html>
-
+<!DOCTYPE html>
```
mateon1: Adding a doctype is good, but now the `<html>` tag was removed! (And github doesn't escape < html > in review comments... Oops)

add-on/src/options/options.html
mateon1: I believe "How often peer count is refreshed" is fine, use simpler words whenever it makes sense.

add-on/src/options/options.html
mateon1: Agreed with Andy, it should probably be "URL"

add-on/src/options/options.html
```diff
@@ -1,56 +1,167 @@
...
+              <dd>URL address of preferred HTTP2IPFS&nbsp;Gateway</dd>
+            </dl>
+          </label>
+          <input type="url" id="customGatewayUrl" inputmode="url" required pattern="https?://.+"/>
+        </div>
+        <div>
+          <label for="useCustomGateway">
+            <dl>
+              <dt>Use Custom Gateway</dt>
+              <dd>replace requests to <em>Public</em> gateway with <em>Custom</em> one</dd>
```
mateon1: "Redirect requests made to *Public* gateways to the *Custom* gateway"

add-on/src/popup/browser-action.html
```diff
@@ -1,5 +1,4 @@
-<html>
-
+<!DOCTYPE html>
```
mateon1: Same in this file, the <html> tag was removed

add-on/src/popup/quick-upload.html
lidel: I think its plain only in review comments. 
Syntax is highlighted correctly when [full file is displayed](https://github.com/ipfs/ipfs-companion/blob/5bee3b0f8df8fe769b537496a0cf14c0ed7e895d/add-on/src/options/options.html).

ps. 🍰

lidel:
Big thanks and 🍰 for everyone!

PR #283, opened 2017-09-17
feat: Poor Man's Protocol Handlers

lidel:
This is an implementation of ridiculous idea from https://github.com/ipfs/ipfs-companion/issues/164#issuecomment-328374052
I am really sorry, but its the only way to get `dweb://` working across all browsers  🙄  

----

As usual, I am not sure how to label checkbox responsible for toggling this feature in _Preferences_.
Current label looks like this:

> 2017-09-17-024645_537x81_scrot

Any ideas how to improve it before I merge this PR?

### DIFF
--- a/add-on/manifest.json
+++ b/add-on/manifest.json
@@ -2,7 +2,7 @@
     "manifest_version": 2,
     "name": "IPFS Companion",
     "short_name": "IPFS Companion",
-    "version" : "2.0.9",
+    "version" : "2.0.10",
 
     "description": "Browser extension that simplifies access to IPFS resources",
     "homepage_url": "https://github.com/ipfs/ipfs-companion",
--- a/add-on/src/lib/common.js
+++ b/add-on/src/lib/common.js
@@ -37,6 +37,7 @@ async function initStates (options) {
   state.automaticMode = options.automaticMode
   state.linkify = options.linkify
   state.dnslink = options.dnslink
+  state.catchUnhandledProtocols = options.catchUnhandledProtocols
   state.dnslinkCache = /* global LRUMap */ new LRUMap(1000)
 }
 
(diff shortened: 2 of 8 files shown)


lgierth:
Awesome :):) Just from a quick glance it might be better to remove `dweb://` here since it's actually `dweb:` and a URI scheme, not a URL scheme. Or can you handle URIs here too? Then `dweb:/ipfs/$cid` and `dweb:/ipns/$peerid` would be correct examples of these.

lidel:
Good question @lgierth !
Right now it supports both single and double slash ( `dweb:/ipfs/$cid` and `dweb://ipfs/$cid` )
Should it also support versions without  leading slash (so that `ipfs:$cid` works too) ?

lgierth:
Right, you get the whole string so you can match whatever you want :) It should only be `ipfs://` `ipns://` and `dweb:`. I've promised to write exactly this down for a while (sorry @diasdavid!)

> Should it also support versions without leading slash (so that ipfs:$cid works too) ?

This would not be a URL anymore, we'd then have a URL scheme named `ipfs` *and* a URI scheme named `ipfs`

lgierth:
> It should only be ipfs:// ipns:// and dweb:

And maybe later also `ipld://` if we decide for an `/ipld` namespace (I think that's still up for discussion?)

lgierth:
For reference https://github.com/ipfs/specs/pull/152#issuecomment-284628862

lidel:
Thanks, I've updated label to: 

> 2017-09-17-163656_519x100_scrot


That being said, extension is more liberal at the moment and both `:/` and `://`  work fine for all schemes (requests are normalized to HTTP URL and redirected to public/local gateway). 

Is that okay, or should we explicitly only support <code>ipfs://$cid</code>, <code>ipns://$peerid</code> and <code>dweb:/ipfs/$cid</code>?

lgierth:
> Is that okay, or should we explicitly only support ipfs://$cid, ipns://$peerid and dweb:/ipfs/$cid?

Better only support what's explicitly supported. Otherwise someone will start using it and we eventually 1) break people's stuff when we remove that implicit support, or 2) have to actually continue supporting it.

> Should fs: be removed?

Yeah can drop fs:

PR #308, opened 2017-11-15
Adds firefox development instructions

alanshaw:
This just adds instructions for installing an "unpacked" extension in Firefox (wisdom gleaned from [this article](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Temporary_Installation_in_Firefox)), and makes it obvious that this is the way you'd also do development.

Can anyone help with these questions?:

1. ~~Is `npm run build` necessary in this context?~~ YES - for copy npm deps into lib
2. I've just read the `CONTRIBUTING.md` and it says use `npm start` to get in installed on firefox - is that still the recommended way?
3. Is it usual for dev instructions be in `CONTRIBUTING.md`? I didn't immediately think to look there...should they be in the `README.md` or at least linked to, from the `README.md`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,20 +89,26 @@ For historical background on the rewrite see [Issue #20: Move to WebExtensions](
 
 Install the latest signed release from [Chrome Web Store](https://chrome.google.com/webstore/detail/ipfs-companion/nibjojkomfdiaoajekhjakgkdhaomnch).
 
-### Other Browsers Supporting WebExtensions API
+### Development or other Browsers Supporting WebExtensions API
 
 Try manual installation:
 
 1. Download Sources
 2. Build it:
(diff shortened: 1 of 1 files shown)


lidel:
Ad 2) You can use it both ways.  `npm start` is just a shorthand for `npm run build` + `npm run firefox` (which runs `web-ext run -s add-on/ --browser-console` behind the scenes).   Installation via `web-ext` may be better for development because  it automatically reloads extension when you make source changes to it or hit "R" in the console.

Ad3) I agree, there should be some information in `README`, with more detailed info in `CONTRIBUTING` file
Thank you for addressing this!  I moved "Contributing" section to be right after "manual install" steps you provided.

PR #313, opened 2017-11-17
Adds docker-build command

olizilla:
_Provides a "quick" way to compare the build output on "another" os._

I'm trying to find a minimal way to use docker as a builder; to have a single command that installs the deps and runs the build in a container, and then ejects the final build artefact.

This PR adds an npm run script called `docker-build` 

```json
"docker-build": "docker run -it --rm --name ipfs-companion-build -v \"$PWD\":/src -w /src node:8.9.1 /bin/bash -c \"npm install && npm run build\""
```

that'll spin up a node:8.9.1 debian jessie based docker container. It runs `npm install && npm run build` in the container. The project dir on the host is mounted as a volume into the container, so the zip file is written to the build dir on the host machine, just as if the build had been run locally.

There are some trade offs. It's simple to run, but the isolation is lower than you'd normally expect from a docker build. This means that your local node_modules dir will contain linux specific modules after running it, so you need to run a local npm rebuild to flip them back to your host os flavour. By being more selective with the mount config, we could probably avoid that, but i wanted to share this as is, to see if it's a reasonable idea.

The more _dockery_ way would be to add a `Dockerfile` to the project. The process to get a bundle would then be

```sh
# create the image
docker build -t ipfs-companion .

# spin up the image `ipfs-companion` as a container process called `ipfs-companion-builder`
docker create --name ipfs-companion-builder ipfs-companion
```
_meanwhile... in another shell_

```sh
# Copy the zip from the container to your local machine.
docker cp ipfs-companion-builder:/src/build/ipfs_companion-2.0.15.zip ./build
```

Which seems way more cumbersome and manual...


olizilla:
As always, writing up the PR notes has helped me think this through.

The `npm run docker-build` command is cool, but the pro solution to this is to have a Dockerfile that is used to build and image during CI for a release. 

On a successful build the CI process would be configured to publish the image to the public docker registry along with the steps to run the container, extract the zip and and publish it to the various browser add-on registries. That'd be rad.

lidel:
Really cool!

I think the issue of polluting host's `node_modules` can be solved by overriding it inside of a container with another (nested) volume.

Quick PoC:

```bash
> mkdir -p node_modules_docker
> docker run -it --rm --name ipfs-companion-build -v "$PWD":/src -v "$PWD"/node_modules_docker:/src/node_modules -w /src node:8.9.1 /bin/bash -c "npm install && npm run build"
> du -sh node_modules*
218M    node_modules
218M    node_modules_docker
```

We don't really care about `node_modules_docker`, it can be removed after the build.

Just for fun, another PoC, where `node_modules`  does not leave docker and is kept in ephemeral 300MB RAM-disk:

```bash
> docker volume create --driver local --opt type=tmpfs --opt device=tmpfs --opt o=size=300m,uid=1000 node_modules_tmpfs
> docker run -it --rm --name ipfs-companion-build -v "$PWD":/src -v node_modules_tmpfs:/src/node_modules -w /src node:8.9.1 /bin/bash -c "npm install && npm run build"
> docker volume rm node_modules_tmpfs 
```

lidel:
@olizilla Now that #311 is merged, I feel `npm run docker-build` may be enough for Mozilla. 
Perhaps we could merge this and add a `Dockerfile` in separate PR?

olizilla:
i'll flip this pr to use the yarn.lock instead

PR #314, opened 2017-11-17
Choo

alanshaw:
I'm just going to park this here...

The UI interactions are currently old school manual DOM mutations. The complexity of the browser actions popup has got to a point where managing it is getting tricky. I thought it might help to introduce a framework to alleviate this problem so this PR adds [choo](https://github.com/choojs/choo), "A 4kb framework for creating sturdy frontend applications" 🚂 🚋 🚋 😛 

Choo is kinda like React in that the UI is driven by state and is kinda like redux in that a route has a "store" where state is initialised and mutated. Components (just functions) deal with what should be shown/hidden based on the state that is provided to them. So instead of manipulating individual DOM elements when things happen, we simply set some state and let the components work out what they should look like.

I'm totally open for this not being accepted or switched up for React or something. I thought that choo would have the smallest impact on bundle size and also means we don't have to do any transpiling (like we would have to do for jsx) so we're still using vanilla JS - hooray.

N.B. if you're twitchy about the lack of HTML syntax highlighting and are using Atom then [do this](https://gist.github.com/alanshaw/a474b1b67e7dfc7c7483059b53af925b)
N.B. this branch is based on the current browserify branch - I'll rebase when/if it gets merged 😄 
N.B. it's worth looking at [just the choo changes](https://github.com/tableflip/ipfs-companion/compare/7d2a04a2151c8e58f2d1c33d986652f55db524e1...tableflip:62c823f8f9cc39201c60eeaccfe48ae598f63d95)
N.B. tests need to get fixed


alanshaw:
@lidel thanks for the kind comments!

> is it something we can easily fix on our side, or is it cost of using Choo?

It's not a limitation of Choo or caused by it, I actually added the delay as I noticed the popup wasn't rendering correctly when it received an update as it was animating in. This was prior to adding choo and is probably out of scope for this PR so I'll remove it.

lidel:
@alanshaw Bam! I merged #311 -- feel free to rebase this PR 👌

alanshaw:
@lidel I've rebased, but I'm getting this error when adding the extension to chrome:

screen shot 2017-11-21 at 09 07 32

lidel:
@alanshaw that's my fault, I committed [Firefox-specific version suffix in `manifest.json`](https://github.com/tableflip/ipfs-companion/blob/d0642e1d383a27dccda47ef8b0ad65774cea9d3c/add-on/manifest.json#L5), my apologies.

Please remove "rc1", it will fix the error under Chrome.

lidel:
Choo is now in `master`. Thank you @alanshaw ! 

I made a test release with it: [🚂🚋🚋🚋v2.1.0rc2🚋🚋](https://github.com/ipfs/ipfs-companion/releases/tag/v2.1.0rc2)

Issue #312, opened 2017-11-16
Muon-based Brave Browser Integrations 


olizilla:
>  - Brave [switched to Chromium in 2018](https://web.archive.org/web/20181017002230/https://brave.com/development-plans-for-upcoming-release/) so I collapsed outdated notes that were related to Muon-based Brave build.  
> - IPFS Companion works perfectly with Chromium-based Brave, and this PR updates our docs to reflect that (https://github.com/ipfs-shipyard/ipfs-companion/pull/609)
> - We now work toward bundling go-ipfs with Brave, with the goal of having something similar to how [Tor Tabs](https://brave.com/tor-tabs-beta) work. 
> - **Follow** → https://github.com/brave/brave-browser/issues/819
>
> – @lidel, _Q4 2018_

---- 

<details>
  <summary>Click to expand historical notes about Muon-based Brave (early 2018)</summary>

`ipfs-companion` runs in brave! 

The following ipfs address styles resolve, either to a local gateway, or the public one.

- ipfs.io/ipfs/QmZyRCtk8MGpuWXn2d1QyTfLWKMTHrT2aHs2vDkfry5zKT
- ipfs://QmZyRCtk8MGpuWXn2d1QyTfLWKMTHrT2aHs2vDkfry5zKT
- dweb:/ipfs/QmZyRCtk8MGpuWXn2d1QyTfLWKMTHrT2aHs2vDkfry5zKT

![screenshot 2017-11-16 10 14 43](https://user-images.githubusercontent.com/58871/32887068-9596dfd0-caba-11e7-954c-494ef255e52f.png)

In the case of `ipfs://` style, brave prompts the user if it's ok to "open in an external application. The url shown is the gateway url rather than what's been entered in the url bar, which could cause confusion. 

![screenshot 2017-11-16 10 15 03](https://user-images.githubusercontent.com/58871/32887083-9eaf5926-caba-11e7-9490-2ea026727c41.png)

A nice feature of using the `ipfs://` form is that it is preserved in the url bar. The other forms are translated to a gateway url.

NURI style isn't supported (just for completeness)
/ipfs/QmZyRCtk8MGpuWXn2d1QyTfLWKMTHrT2aHs2vDkfry5zKT is redirected to
file:///ipfs/QmZyRCtk8MGpuWXn2d1QyTfLWKMTHrT2aHs2vDkfry5zKT

Brave supports a subset of the the chrome-extensions api, so there are few issues to figure out.

[...]

PR #315, opened 2017-11-17
Fallback to opening options page manually in brave

olizilla:
In brave the `runtime.openOptionsPage()` call is failing, so the user can't update their prefs.

See: https://github.com/brave/browser-laptop/issues/7812

This PR adds a fallback to opening the options page as a browser tab if the `runtime.openOptionsPage()` throws.

screenshot 2017-11-17 16 13 59

Also, we add try/catch around calls the following _currently_ unsupported apis
- `contextMenus.create`
- `contextMenus.update`

In the meantime I'll [follow up with brave](https://github.com/brave/browser-laptop/issues/9556#issuecomment-345229249) to see what we need to do to make them available.

WIP on #312

### DIFF
--- a/add-on/src/lib/ipfs-companion.js
+++ b/add-on/src/lib/ipfs-companion.js
@@ -206,7 +206,7 @@ try {
     onclick: copyAddressAtPublicGw
   })
 } catch (err) {
-  console.error('Error creating contextMenus', err)
+  console.log('[ipfs-companion] Error creating contextMenus', err)
 }
 
 function inFirefox () {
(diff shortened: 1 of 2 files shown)


lidel:
LGTM, but lets keep it open until #311 is merged.

lidel:
@olizilla  now that #311 is merged, feel free to rebase this PR against master  so that we can merge it too.

olizilla:
@lidel all fixed up! We still don't get to see the browserAction icon in brave yet which is annoying but this PR gets the basics working in brave.

PR #317, opened 2017-11-20
Remove private fields from bundled package.json

olizilla:
Fixes the issue of private, build machine specific fields from package.json appearing in the bundle, by adding https://github.com/fanatid/browserify-package-json as a global browserify transform 

After this change the diff is clean between a local build via `npx yarn@1.2.1 && npx yarn@1.2.1 build` and a `docker-build` run.

See: https://github.com/ipfs/ipfs-companion/pull/311#issuecomment-345771694


lidel:
Perfect! Thanks!

PR #319, opened 2017-11-28
adds popup header and adds browserify-css and tachyons

alanshaw:
This is the popup UI component, as per the [mockups](https://github.com/ipfs/ipfs-companion/pull/310#issuecomment-347244281). It's possible for this PR to be merged and released as is.

This adds a header to the popup displaying the IPFS icon (online/offline):

screen shot 2017-11-28 at 13 26 55
screen shot 2017-11-28 at 13 27 11

This also add the HTML code and CSS for the proposed embedded/external toggle but the component is currently `display: none`. It'll become visible when JS IPFS is added to the project.

screen shot 2017-11-28 at 13 28 13

I've added `browserify-css` to allow us to require CSS files and added `tachyons` to give us consistent spacing, font sizes and ways of composing styles for components without bloating CSS (and is also consistent with other IPFS sites https://peerpad.net https://ipld.io). I've actually been able to remove _all_ the styles in the browser action popup as they can all be expressed using tachyons atoms.


add-on/src/popup/browser-action/mdc.switch.css
```diff
@@ -0,0 +1,172 @@
+/*!
```
olizilla: No module available for `mdc.switch.css` ?

alanshaw: `@material/switch` - but it only allows you to override the "on" state:

> It is not currently possible to customize the color of a disabled or off (unchecked) switch
https://www.npmjs.com/package/@material/switch#sass-mixins

Our "off" state is the same as our "on" state (visually), we're not so much on/off as left/right so I needed to edit the CSS a little to achieve this.

alanshaw:
Nice, I like it!

lidel:
If anyone wants to smoke-test refreshed GUI under Firefox, there is a new beta in Development channel: 
https://addons.mozilla.org/en-US/firefox/addon/ipfs-companion/versions/beta

In my case it works well, even with `layout.css.devPixelsPerPx` bigger than `1.0`  (HiDPI screens) 👍

richardschneider:
Fails to load in Firefox 55.0.3 (32-bit)
companion

lidel:
@richardschneider I am afraid we switched min. version to Firefox 57, 
older versions lacked some APIs on Android (https://github.com/ipfs/ipfs-companion/issues/109#issuecomment-336747077). 

I see it is possible to set different range for desktop and mobile, 
so I've just manually switched min. version for desktop to "55".  

Try again 🙂

richardschneider:
Fast work, thanks it now load and works (in a smoke test).

However, WebUI seems broken, no data is displayed.

webui

lidel:
The issue  you are experiencing is not related to ipfs-companion, I think.

I am using `go-ipfs/0.4.13/cc01b7f` and WebUI loads fine.
Try to update to 0.4.13 if you are not using it already.

If the problem persists, report it at https://github.com/ipfs/webui/issues

PR #320, opened 2017-11-28
Init ipfs-api or js-ipfs

olizilla:
Provide a mechanism for swapping between embedded and external ipfs nodes.

Adds `ipfsNodeType` as an option which is expected to be either `embedded` or `external` for use in the proposed UI in #319 

To allow js-ipfs and ipfs-api to be used interchangeably, I've tweaked add-on code to use global `Buffer` impl povided by browserify, and use `ipfs.files.add` instead of `ipfs.add`. Intialisation is pulled into there own modules so we've got the flexibility to tweak the returned objects in the future if we find we need to bridge temporary api divergence, or customise anything.

Fixes a bundling issue by adding `prundupify` as a browserify plugin... due to https://github.com/browserify/factor-bundle/issues/51 (factor-bundle and regular browserify de-duping don't play well together, which became an issue when bundling js-ipfs)

This is bite-size chunk of a bigger change, so I figured, best to get your thoughts on it early. I'm looking at pulling the external api polling from `ipfs-companion.js` to the `external.js` wrapper, so it'll become more similar to the `embeded.js` wrapper, as both will have some clean up to do in the destroy function.


package.json
```diff
@@ -68,9 +68,11 @@
   },
   "dependencies": {
     "choo": "6.6.0",
+    "ipfs": "^0.26.0",
```
lidel: Let's try to keep all versions locked, so that we don't have big divergence between `npm` and `yarn`, at least on the direct dependency level.

diasdavid: Yeah, for IPFS land, always use `~` at least. A minor version might have a breaking change (0.27 will)

olizilla:
Toggling ipfs node type works in Brave!

js-ipfs-companion

diasdavid:
🎆🎇 THAT IS SO RAD!!!!! 🎆🎇
👏🏽👏🏽👏🏽👏🏽

Is the ipfs-companion also capturing `ipfs://`?

olizilla:
@diasdavid 
soon

olizilla:
@diasdavid was there a trick to getting the `chrome.protocol` apis to be available in https://github.com/ipfs-shipyard/browser-laptop/pull/1 ?  I've tried adding it as a requested permission, but I'm now seeing `chrome.protocol` as undefined, and trying to track down why.

diasdavid:
@olizilla it might have been to how the extension was being loaded internally. Could you check with the Brave team?

olizilla:
@alanshaw solved it!

> Needed `manifestLocation` set to `component`. I think https://github.com/brave/muon/blob/1fd29625480ea3a7b66bd71da11f7bcc49daf420/brave/common/extensions/api/_api_features.json#L35-L42 restricts the protocol api to this.

For future explorers, you have to load the plugin programatically in brave and pass in an additional arg to get access to privileged apis like `chrome.protocol`

In [brave/browser-laptop](https://github.com/brave/browser-laptop) you link the add-on into your `app/extensions` dir add the following to [`app/extensions.js`](https://github.com/brave/browser-laptop/blob/81f7991ed8b2cfd4a411fee2b638cfd60f090db0/app/extensions.js#L500-L502)

```js
  // Enable IPFS
  extensionInfo.setState('ipfs', extensionStates.REGISTERED)
  loadExtension('ipfs', getExtensionsPath('ipfs'), undefined, 'component')
```
 it's the `component` arg that gives it super powers.

olizilla:
TODOs plucked from https://github.com/ipfs/ipfs-companion/pull/310#issuecomment-347342404
- [ ] _it should have analogous, well-documented checkbox in the Preferences_
- [ ] short description of Embedded option in a tool-tip when user hovers cursor on it.
- [ ] explicitly list "API" address (under "HTTP GATEWAY").

olizilla:
Ok, i've made a smaller change around

> refactor offline check

Currently a peerCount of -1 means we couldn't connect to the external ipfs proc.
This updates the automatic gateway switching logic to use that signalling value
to decide whether to switch to the public gateway. Zero peers means we're
initialising, rather than we're offline.

This improves the ux when switching between embedded and external nodes.

A more extensive change can be done later, wrapping the ipfs-api client with a
cache, and emitting events for when ipfs state changes.

I'm planning on adding a feature flag of some sort, so we can merge this PR without exposing the js-ipfs option to existing users just yet. I'm focused on getting this PR mergable, before tackling any more changes.

lidel:
@olizilla thank you!   The "0 == initializing rather than offline" change sounds good.

As soon we have the flag I'll do my best to find time to smoke-test and merge it.

add-on/src/lib/ipfs-client/index.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
-const embedded = require('./embedded')
 const external = require('./external')
+const embedded = require('./embedded')
```
diasdavid: Some feedback from the last IPFS all hands was that `embedded` might confuse devs to think that know there is a `window.ipfs` in every page. Is that one of the goals too? It would be a good feature to have (optionally)

olizilla:
I've added a minimal feature flag for the embedded ipfs work. Marking `ipfs: false` in the browsers section of the package.json, causes browserify to stub out `require('ipfs')`, returning an empty object instead of js-ipfs. I've added an `isJsIpfsEnabled` test that checks for that, and used it to hide the fieldset in the options page, and the toggle switch in the browser action menu. It will also cause js-ipfs to be excluded from the bundle, which is nice.

Ideally this would be controlled by an env var like "JS-IPFS-ENABLED=true". I'll make that work in a future PR, if it's needed. To make that work we'd need to extract the browserify build process to a js file and use the browserify api, so we can then parameterise the excludes option based on env...

alanshaw:
👓

add-on/src/lib/ipfs-client/index.js
olizilla: I think that would be simple to add. Is the idea that uses would write scripts against `window.ipfs`  but not bundle it with their script? It sounds a little bit like those browser extensions that add jquery to every page, so you can noodle around in the dev console. 

Do people have a use-case in mind? T feel like we should be directing devs to the js-ipfs examples, to show them how to bundle it.

diasdavid: The difference between adding jquery and ipfs is that on one you would be just packing more code, the second you are opening more ports and doing more crypto which are both expensive operations.

The Metamask team has been injecting Web3 to every page to let devs use Ethereum directly if WebPages are there.

I feel this should be one of the options for the extension and until we have concrete use cases, just refer to an example on how to use it to seed it to developers.

VictorBjelkholm: Lots of benefits for embedding an already running IPFS node in webpages, too many to list here, but for starter, it makes startup times a lot better as nothing has to initialized, just start using it directly (node in background is already running so!). Performance of loading pages and loading content will be faster as well. Things can be shared via MFS between applications, as long as they agree on a standard structure. ID can follow between applications as well. Just a few reasons.

olizilla: Very cool! I like the idea of making ipfs available to webapps and seeing what people do with it.

lidel: See https://github.com/ipfs/ipfs-companion/issues/330, I created it to track `window.ipfs`  effort :)

olizilla: This! https://github.com/VictorBjelkholm/js-ipfs-in-the-browser

add-on/_locales/en/messages.json
```diff
@@ -7,24 +7,40 @@
...
+    "message": "Peers",
     "description": "A label in Node status section of Browser Action pop-up"
   },
+  "panel_statusSwarmPeersTitle": {
+    "message": "The number of other ipfs nodes you can connect to",
+    "description": "A label tool-tip in Node status section of Browser Action pop-up"
+  },
   "panel_quickUpload": {
-    "message": "Quick Upload",
+    "message": "Share files via IPFS",
```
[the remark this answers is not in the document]
lidel: @alanshaw Yes, this is what should be done, but I think this step  can be skipped  now that we [use Crowdin](https://crowdin.com/project/ipfs-companion).

Crowdin will change them back to english after I upload new version of source locale (`en`), at least that is what happened when I've added new keys (other locales automatically got english stubs).

alanshaw:
@lidel I think this PR is ready to merge if you'd like to take a final look? If we can get this in I'll rebase #323 and it'll also be ready for review & merge.

lidel:
Firefox users can try [v2.1.1beta5](https://addons.mozilla.org/en-US/firefox/addon/ipfs-companion/versions/2.1.1beta5), which includes these changes. 👍

diasdavid:
Woooooot! :D

diasdavid:
How can I switch to a js-ipfs node? 

image

I don't see an option on preferences.

lidel:
@diasdavid It is Brave-only for now 🙂  (disabled in default build).
To enable it, you need to mark ipfs: true in the browser section of the `package.json` and rebuild via `npm run yarn-build` 
(Thanks to this build flag Firefox/Chrome versions do not need to ship with unused codebase of `js-ipfs`)

alanshaw:
@diasdavid @lidel you have to remove `"ipfs": false` from the `package.json`, setting it to `true` won't work! We'll add a note to the README.

PR #321, opened 2017-11-29
Refactor ipfs-companion so it no longer sets properties on the window

alanshaw:
This PR refactors the `ipfs-companion` module so it no longer sets properties on the window object.

I'm viewing the `ipfs-companion` module as something I'd like to create an instance of, or multiple instances of, and also teardown - with the end goal of making it easier to reason about, maintain and test. So this PR stops it from setting properties on the window object and instead `init` returns an API object which is the method one should communicate with the module.

`background.js`, after creating the new instance, sets it to `window.ipfsCompanion` so that the popup, quick upload and options pages can make use of it's API.

* Pulls out the notifier logic into it's own module
* Pulls out the copying logic into it's own module
* Uses `port.postMessage` to signal to the companion to perform a copy (instead of calling the functions on the background page)
* Refactors some of the async `FileReader` logic into async/await style


alanshaw:
@lidel ok all fixed up - I moved the context menu initialisation and update stuff into the `context-menus` module and that has fixed it.

olizilla:
This is quickly going to get difficult to merge with #320 and friends, so I'm gonna take a look at it now and see if it's good to go  🚀✨ 👀

olizilla:
@lidel do you have a minute to merge this? Merging is blocked on re-review. In retesting this I can see the error message has been resolved.

lidel:
@olizilla my apologies, I run out of time to review this yesterday.

Old error is gone, but now I get this one on every API poll (_while external node is down_):

```
ReferenceError: can't access lexical declaration `offlinePeerCount' before initialization  background.js:360:
```

Are you able to reproduce? If not, it may be the machine I am on, I'll retest on another one in 5h.

olizilla:
@lidel yes, good spot, I'm seeing that too. @alanshaw can you take a look when you get a moment?

olizilla:
Actually, I see the problem... pesky hoisting / lexical scoping. Am fix.

lidel:
Thanks! 👌

PR #323, opened 2017-12-01
Adds self served directory listings

alanshaw:
\o/ Directory viewer in Brave for handled `ipfs://` requests!

screen shot 2017-12-04 at 09 34 14

screen shot 2017-12-04 at 09 34 22


add-on/_locales/en/messages.json
```diff
@@ -36,7 +36,7 @@
     "description": "A label in Node status section of Browser Action pop-up"
   },
   "panel_quickUpload": {
-    "message": "Quick Upload",
+    "message": "Share files via IPFS",
```
lidel: I must say, it is a much better name for this feature 👍

package.json
```diff
@@ -69,10 +69,16 @@
   },
   "dependencies": {
     "choo": "6.6.0",
+    "doc-sniff": "^1.0.1",
```
lidel: I've found two ways to avoid the chore of removing `^` from `package.json`
- while doing `npm install` of new dependency, one can do it with `--save --save-exact`
- to make it default behaviour: `npm config set save-exact true`

olizilla:
@lidel when you get a moment to look at this pr, this compare view to just check the changes without all of #320 in too, is helpful:

https://github.com/tableflip/ipfs-companion/compare/84b2ada880b2cc8296c6752bbd816f7ae3aa0126...tableflip:feat/dir-view#diff-ca7aa692aef78b2031d7a5da30e03248R10

add-on/src/lib/ipfs-protocol.js
```diff
@@ -1,31 +1,57 @@
...
-      resolve({mimeType, data: res.toString('utf8')})
-    })
-  })
+async function getDataAndGuessMimeType (ipfs, path) {
+  let data
+
+  try {
+    data = await ipfs.files.cat(path)
+  } catch (err) {
+    if (err.message.toLowerCase() === 'this dag node is a directory') {
```
lidel: Ouch, is this the only way we can detect directories?

alanshaw: I probably should have explained this a little...here goes:

It's similar to how the js-ipfs http gateway works. It tries to "resolve" a path (e.g. /ipfs/$hash/path/to/a/file) by using `ipfs.dag.get` to walk the path and eventually find the hash of the file at the end. It uses `Unixfs.unmarshal` on the end file to detect a directory and will [raise an error if so](https://github.com/ipfs/js-ipfs/blob/master/src/http/gateway/resolver.js#L94-L100) "this dag node is a directory". [If this error occurs](https://github.com/ipfs/js-ipfs/blob/master/src/http/gateway/resources/gateway.js#L39) then it'll try to get the links using `ipfs.object.get` and render a directory view.

Instead of walking the path using `ipfs.dag.get` (which isn't available in js-ipfs-api) we use `ipfs.files.cat` which has the advantage of being able to pass a path to it (so the walking is done for us) and is what we'd call to get file content anyway. It also returns the same error - "this dag node is a directory" for directories. We use `ipfs.ls` instead of `ipfs.object.get` because:

1. `ipfs.files.cat` resolved the hash but didn't return it to us so we can't use `ipfs.object.get` anyway since it needs a hash
2. We can pass a path to `ipfs.ls` and it'll give us the links

Is there a better way? @diasdavid

lidel:
@alanshaw  If you resolve conflict with master, I think we can merge this before it gets cold. 👍

diasdavid:
🌟

PR #331, opened 2017-12-12
Refactors options page to use choo

alanshaw:
\o/ nothing changes visually, but options page now uses choo!...bringing it in line with the other pages

* Modules for each form/section
* Simplifies the code
* Removes `lib/data-i18n.js` as it's no longer used anymore


lidel:
@alanshaw  I see some files related to dir listing. 
This should be merged after #323, right?

add-on/src/options/forms/api-form.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+/* eslint-env browser, webextensions */
+
+const browser = require('webextension-polyfill')
+const html = require('choo/html')
+
+function apiForm ({ ipfsApiUrl, ipfsApiPollMs, automaticMode, onOptionChange }) {
+  const onIpfsApiUrlChange = onOptionChange('ipfsApiUrl')
```
lidel: This should be normalized just like `customGatewayUrl` is

alanshaw:
@lidel closing as I've opened #337 which has only the options page changes in and the change requested above (it doesn't depend on #323).

I'm not Android enabled so unfortunately I cannot verify, sorry :(

lidel:
@alanshaw No problem, I'll test  #337 on Android and let you know if any issues.

Issue #330, opened 2017-12-11
Expose IPFS API as window.ipfs


lidel:
There is an area of interest that can be summarized as.. 

> ..exposing a subset of IPFS APIs as `window.ipfs` on every webpage ✨ 

**tl;dr** If we have it, websites could detect that `window.ipfs` already exists and use it instead of spawning own `js-ipfs` node (saves resources, battery etc).

We had discussions in https://github.com/ipfs/js-ipfs-api/issues/387 and https://github.com/lidel/ipfs-firefox-addon/issues/37, https://github.com/ipfs/in-web-browsers/issues/9, https://github.com/ipfs/in-web-browsers/issues/35  but the consensus at the time was that it is not safe without some kind of access-control (eg. via proposed API Tokens: https://github.com/ipfs/go-ipfs/issues/1532).

@VictorBjelkholm  suggested a simple solution to the lack of access controls in the API itself: 
On the first use of `window.ipfs` user could be prompted with something like _"Do you want to allow scripts at `<website>` to access API of your IPFS node?"_. The choice would be remembered eg. for the page til the end of browsing session.  There could also be an additional _"remember for this site"_ checkbox to make UX better for users who wants to make permission persist between browser restarts. 

## Tasks

- [x] Create a [content script](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Anatomy_of_a_WebExtension) that creates a proxy object under `window.ipfs` of the page it is injected into
    - [x] This object should expose  a _subset_ of IPFS APIs
      > `<victorbjelkholm>`  there is of course a lot more once there is a full ipfs node in the browser, but if we only have the simplest and most useful functions (add, cat, pubsub subscribe/publish, dag get/put), it suddenly becomes a lot simpler and useful quickly
    - [x] Operations performed on `window.ipfs` should be proxied to the real IPFS API object present in webextension's [Background page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Anatomy_of_a_WebExtension) (requires robust (de)serialization, we don't want to mangle upload data etc)
   - [x] First use of `window.ipfs` should trigger a dialog where user confirms that access to IPFS API is granted to requesting site (and decides if permission is for this session only or permanent)
- [x] Add a checkbox under "Preferences/Experiments" to control if content script  is injected to every visited page, make it disabled  by default initially
- [x] Write documentation, namely provide code snippet that enables apps to use the IPFS Companion IPFS node if it is available

### Potential problems

[...]

PR #333, opened 2017-12-12
window.ipfs

alanshaw:
This is basically the good work of @VictorBjelkholm done in https://github.com/ipfs/ipfs-companion/commit/6b0ac8b6c0dbb45928a6e2ecaffed5d5e26f0ab7 but cleaned up a little.

refs #330

This is a first pass effort at getting this to work.

Demo: https://youtu.be/t1ldUp_mjDk

### What's in this PR?

* `window.ipfs` property in your web pages!
    * also adds window.Buffer if not there
    * also adds window.Ipfs
* Proxies to running js-ipfs or go-ipfs node (via js-ipfs-api) in the extension
* Adds option to the preferences to enable/disable adding ipfs to the window

screen shot 2018-01-12 at 16 02 59

* Access controls on certain [ipfs functions](https://github.com/tableflip/ipfs-companion/blob/9b29486e9a202b431a695b08f8d6a45f0b55df4f/add-on/src/lib/ipfs-proxy/index.js#L9-L31) prompt user for a allow/deny decision on first use
* Adds a page for managing permissions

screen shot 2018-01-12 at 16 02 38

* Adds tests for access control and permissions page
* Adds `nyc` for code coverage

### What's still todo (in a future PR)?

* Add a custom confirm dialog. Firefox does not support `window.confirm` from the background process so it fails - **all access to `window.ipfs` on Firefox will be automatically denied**, however as a temporary workaround, Firefox users can use the manage permissions page to enable permissions after they have been auto denied
    * Another reason this is needed is that we want to be able to grant access to all (or a subset of) the API for a given origin, `window.confirm` can't cover these UX requirements 
* Tighten access restrictions to origin/ipfs/hash - currently only restricted to origin
* Level up [ipfs-postmsg-proxy](https://github.com/tableflip/ipfs-postmsg-proxy) to support streaming properly and then also the progress options for files add. More info [here](https://github.com/tableflip/ipfs-postmsg-proxy#caveats)

### How does it work?

https://www.youtube.com/watch?v=aXLERp9-Mis


add-on/manifest.json
```diff
@@ -38,6 +38,22 @@
...
     },
 
+    "content_scripts": [
+      {
+        "matches": [
+          "file://*/*",
+          "http://*/*",
+          "https://*/*"
+        ],
+        "match_about_blank": true,
```
VictorBjelkholm: Don't think it's necessary to inject on blank pages no?

add-on/manifest.json
```diff
@@ -38,6 +38,22 @@
         "page": "dist/background/background.html"
     },
 
+    "content_scripts": [
+      {
+        "matches": [
+          "file://*/*",
+          "http://*/*",
```
lidel: A single entry with `<all_urls>` would be better here:

> A match pattern is essentially a URL that begins with a permitted scheme (http, https, file, or ftp, and that can contain '*' characters. The special pattern `<all_urls>` matches any URL that starts with a permitted scheme. ([src](https://developer.chrome.com/extensions/match_patterns))

alanshaw: We'll eventually inject programmatically due to the need to turn this feature off, but yes that would be better.

alanshaw:
Extracting this proxy idea for `window.ipfs` into a module, using `interface-ipfs-core` tests to establish whether it works or not: https://github.com/tableflip/ipfs-postmsg-proxy

add-on/src/contentScripts/ipfs-proxy/content.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+const browser = require('webextension-polyfill')
+const injectScript = require('./inject-script')
+
+function init () {
+  const port = browser.runtime.connect({ name: 'ipfs-proxy' })
```
VictorBjelkholm: If other extensions know this name, is it possible they can connect to it? That would be bad, as they could bypass the permissions you've setup, if I understand correctly.

alanshaw: It would! I'll check it out.

alanshaw: ...although thinking about it, the access control is all done in the extension background process, so it doesn't matter who's connecting and asking for stuff, it still has to be authorised by the user.

I'll double check but I think it might be ok.

alanshaw: Ok so, the [background process is listening to `runtime.onConnect`](https://github.com/tableflip/ipfs-companion/blob/9b0009dd945c7d560a335c534d8698fd72315550/add-on/src/lib/ipfs-proxy/index.js#L67):

> Fired when a connection is made with either an extension process or a content script.
> https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/onConnect

It would need to listen to [`runtime.onConnectExternal`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/onConnectExternal) to receive connections from other extensions.

Aside from that, even if connections from other extensions _were_ allowed, the authorization is done in the background process, so the user would still need to have approved the function to be called before the proxy server let it through to the IPFS instance.

add-on/src/lib/ipfs-companion.js
```diff
@@ -43,9 +46,14 @@ module.exports = async function init () {
       onCopyAddressAtPublicGw: () => copier.copyAddressAtPublicGw()
     })
     modifyRequest = createRequestModifier(getState, dnsLink, ipfsPathValidator)
+    ipfsProxy = createIpfsProxy(() => ipfs, getState)
     registerListeners()
     await setApiStatusUpdateInterval(options.ipfsApiPollMs)
-    await storeMissingOptions(options, optionDefaults, browser.storage.local)
+    await storeMissingOptions(
+      await browser.storage.local.get(),
```
olizilla: Can you explain this change? The code looks fine, just curious.

alanshaw: Yes! Previously `storeMissingOptions` wasn't doing anything.

**On first run**, `options` will be the same as `optionDefaults`. Due to the line earlier:

```js
const options = await browser.storage.local.get(optionDefaults)
```

Regarding the parameter to `get`:

> A key (string) or keys (an array of strings or an object specifying default values) to identify the item(s) to be retrieved from storage
> https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/storage/StorageArea/get#Parameters

...so, it finds `{}` in storage and then assigns `optionDefaults` to it.

So `storeMissingOptions` is given two identical objects and does nothing.

Similarly for subsequent runs, options will be assigned any missing defaults so there's never anything to store!

Instead, we get a fresh copy of the current options, without defaults applied.

add-on/src/lib/ipfs-proxy/access-control.js
```diff
@@ -0,0 +1,73 @@
...
+    return { origin, permission, allow: acl[origin][permission] }
+  }
+
+  async setAccess (origin, permission, allow) {
+    return this._writeQ.add(async () => {
+      const access = { origin, permission, allow }
+      const acl = await this.getAcl()
+
+      acl[origin] = acl[origin] || {}
+      acl[origin][permission] = allow
```
olizilla: Guard against the **_forbidden origins_** like `__proto__`?  👻

olizilla: It's less pretty, but might be safer to structure the acl with data that comes from the outside as values rather than keys...

add-on/src/lib/ipfs-proxy/index.js
```diff
@@ -0,0 +1,137 @@
...
+  'object.patch.appendData',
+  'object.patch.setData',
+  'pin.add',
+  'pin.rm',
+  'pubsub.publish',
+  'swarm.connect',
+  'swarm.disconnect'
+]
+
+function createIpfsProxy (getIpfs, getState) {
```
olizilla: This function is doing some intricate plumbing. It's very cool. Please add some docs to pin down some of the helpful background context that is in your head so future us can pick it up.

Things like "this is when a tab is created, but also _caveat_, _caveat_" and "This end is talking from proxy to ipfs node, this end is content script to proxy.", those sort of things.

package.json
```diff
@@ -37,7 +40,10 @@
   "private": true,
   "preferGlobal": false,
   "devDependencies": {
+    "babel-preset-env": "^1.6.1",
```
olizilla: Pin those deps.

package.json
```diff
@@ -18,13 +15,19 @@
...
     "build:copy:wx-polyfill-lib": "shx cp node_modules/webextension-polyfill/dist/browser-polyfill.min.js add-on/dist/contentScripts/browser-polyfill.min.js",
-    "build:js": "browserify -p prundupify -t browserify-css -t [ browserify-package-json --global ] add-on/src/background/background.js add-on/src/options/options.js add-on/src/popup/browser-action/index.js add-on/src/popup/quick-upload.js -p [ factor-bundle -o add-on/dist/background/background.js -o add-on/dist/options/options.js -o add-on/dist/popup/browser-action/browser-action.js -o add-on/dist/popup/quick-upload.js ] -o add-on/dist/ipfs-companion-common.js",
+    "build:js": "browserify -p prundupify -t browserify-css -t [ browserify-package-json --global ] add-on/src/background/background.js add-on/src/options/options.js add-on/src/popup/browser-action/index.js add-on/src/popup/quick-upload.js add-on/src/pages/proxy-acl/index.js -p [ factor-bundle -o add-on/dist/background/background.js -o add-on/dist/options/options.js -o add-on/dist/popup/browser-action/browser-action.js -o add-on/dist/popup/quick-upload.js -o add-on/dist/pages/proxy-acl/proxy-acl.js ] -o add-on/dist/ipfs-companion-common.js",
```
olizilla: Note to self... capture the browserify config in a file so we can re-use it for watchify.

test/functional/lib/ipfs-proxy/access-control.test.js
```diff
@@ -0,0 +1,134 @@
...
+describe('lib/ipfs-proxy/access-control', () => {
+  it('should maintain ACL', async () => {
+    const accessControl = new AccessControl(new Storage())
+    let acl = await accessControl.getAcl()
+
+    // Ensure ACL starts out empty
+    expect(acl).to.deep.equal({})
+
+    const sets = [
+      ['http://ipfs.io', 'ipfs.files.add', true],
```
olizilla: Worth ensuring that urls are normalised correctly...

are

- `http://ipfs.io
- `http://ipfs.io/`
- `http://ipfs.io/ipfs/QmHash`
- `https://ipfs.io/`

equivalent for permissions purposes? are

- `https://www.ipfs.io`
- `https://ipfs.io`

different for permissions purposes?

olizilla:
`npm run build` fails with:

```
Error: Cannot find module 'babelify' from '/Users/oli/Code/ipfs/ipfs-companion/node_modules/ethereumjs-block'
```

More context:

```
> ipfs-firefox-addon@ build:js /Users/oli/Code/ipfs/ipfs-companion
> browserify -p prundupify -t browserify-css -t [ browserify-package-json --global ] add-on/src/background/background.js add-on/src/options/options.js add-on/src/popup/browser-action/index.js add-on/src/popup/quick-upload.js -p [ factor-bundle -o add-on/dist/background/background.js -o add-on/dist/options/options.js -o add-on/dist/popup/browser-action/browser-action.js -o add-on/dist/popup/quick-upload.js ] -o add-on/dist/ipfs-companion-common.js

Error: Cannot find module 'babelify' from '/Users/oli/Code/ipfs/ipfs-companion/node_modules/ethereumjs-block'
    at /Users/oli/Code/ipfs/ipfs-companion/node_modules/resolve/lib/async.js:48:31
    at processDirs (/Users/oli/Code/ipfs/ipfs-companion/node_modules/resolve/lib/async.js:182:39)
```

This was after a `rm -rf node_modules && npm i`

olizilla:
```
npx yarn@1.3.2
npx yarn@1.3.2 build
```
works fine though.

add-on/src/lib/ipfs-companion.js
lidel: Nice catch!  I think I've made this bug while porting from legacy SDK to WebExtension APIs. 
Due to the "fallback to defaults" it worked anyway, so we never noticed :):)

olizilla:
No `window.ipds` when run in chrome atm.

add-on/src/lib/ipfs-companion.js
alanshaw: ...probably should be in a separate PR - apologies

test/functional/lib/ipfs-proxy/access-control.test.js
alanshaw: The origin for permissions purposes is [extracted out of the URL that is reported by the port](https://github.com/tableflip/ipfs-companion/blob/af594639a81a87e54f5de1294133f11d29750c9b/add-on/src/lib/ipfs-proxy/index.js#L41) so this information doesn't come from "outside", and nor does the permission name (comes from [`ACL_FUNCTIONS`](https://github.com/tableflip/ipfs-companion/blob/af594639a81a87e54f5de1294133f11d29750c9b/add-on/src/lib/ipfs-proxy/index.js#L48-L51)) or allow/deny decision (comes from [`window.confirm`](https://github.com/tableflip/ipfs-companion/blob/af594639a81a87e54f5de1294133f11d29750c9b/add-on/src/lib/ipfs-proxy/index.js#L130)).

So the origin _is_ normalised already by virtue of parsing the URL and extracting the origin. For permissions we're using the same-origin policy, so URLs for a different origin need different approval.

> An origin is defined as a combination of URI scheme, host name, and port number
> https://en.wikipedia.org/wiki/Same-origin_policy

* `http://ipfs.io` and `http://ipfs.io/` are normalised to `http://ipfs.io`
* `http://ipfs.io/ipfs/QmHash` not supported as a separate "origin" yet (mentioned in "What's still todo (in a future PR)?" in the description) so origin would be `http://ipfs.io`
* `https://ipfs.io` **is a different origin** for permissions (because scheme differs)
* `https://www.ipfs.io` and `https://ipfs.io`, **also a different origin** for permissions (because host differs)

add-on/src/lib/ipfs-proxy/access-control.js
alanshaw: I think a `Map` instead of an plain old object would be safer right?

alanshaw:
@olizilla review feedback done, is good now?

add-on/src/lib/ipfs-proxy/index.js
```diff
@@ -5,6 +5,8 @@ const browser = require('webextension-polyfill')
 const { createProxyServer, closeProxyServer } = require('ipfs-postmsg-proxy')
 const AccessControl = require('./access-control')
 
+// These are the functions that require an allow/deny decision by the user.
+// All other exposed IPFS functions are available to call without authorization.
```
olizilla: Perhaps one for a next PR, but ongoing maintenance would be simpler if we flip this to a whitelist of methods that can be called without access control... as the ipfs api changes we can jump to the latest release and then figure out later if a new api should be whitelisted.

test/functional/lib/ipfs-proxy/access-control.test.js
olizilla: ah sure, yes, I was trying to say that there is an implicit dependency for this to function correctly on the url having been normalised before it gets passed on to the ACL, but there is no test to assert it. 

With the new changes the ACL will error out if the user tries to set a value that isn't a valid origin as normalised by `URL`, so i guess that covers it.

olizilla:
It's working in Chrome again now! Is very nice.

test/functional/lib/ipfs-proxy/access-control.test.js
alanshaw: ✨ 🚀 ✨

add-on/src/lib/ipfs-proxy/access-control.js
```diff
@@ -0,0 +1,104 @@
...
+/* eslint-env browser */
+
+const EventEmitter = require('events')
+const PQueue = require('p-queue')
+
+class AccessControl extends EventEmitter {
+  constructor (storage, key = 'ipfsProxyAcl') {
+    super()
+    this._storage = storage
+    this._key = key
```
lidel: What if we store ACL keys with a prefix?
eg.  set `this._key = 'ipfsProxyAcl.' + key` (or add/remove a prefix in `getAcl` and `_setAcl`)
It removes risk of overriding storage keys + makes it easier to browse/sort local storage keys when debugging the extension.

add-on/src/lib/ipfs-proxy/index.js
```diff
@@ -0,0 +1,153 @@
...
+}
+
+async function requestAccess (origin, permission) {
+  const msg = `Allow ${origin} to access ipfs.${permission}?`
+
+  // TODO: add checkbox to allow all for this origin
+  let allow
+
+  try {
+    allow = window.confirm(msg)
```
lidel: This is just FYI, but just like you said this won't work on Firefox. Modals are not allowed from background page. 
Confirmed with `ipfs.add` – it results in no prompt, access is automatically denied:

> 2018-01-24-231332_401x57_scrot
>  2018-01-24-231415_281x122_scrot

Browser Console confirms prompt was not allowed:

>   2018-01-24-231606_835x137_scrot

There are related tickets at bugzilla **tl;dr** this is by design and they suggest to display prompt via content script:

> We don't intend to support alert() in background pages.
**If you want to alert data based on a context menu item, you can do so from a content script.**
> – https://bugzilla.mozilla.org/show_bug.cgi?id=1203394#c41

> I'm telling you that it won't be accessible from background pages. Alerts are modal dialogs, and they're meant to be tied to a specific, user-visible window. If you want to trigger an alert from a context menu operation, it needs to be triggered from the appropriate content window.
> – https://bugzilla.mozilla.org/show_bug.cgi?id=1203394#c43

add-on/_locales/en/messages.json
```diff
@@ -265,6 +265,18 @@
     "message": "DNSLINK Support",
     "description": "An option title on the Preferences screen (option_dnslink_title)"
   },
+  "option_ipfsProxy_title": {
+    "message": "window.ipfs",
+    "description": "An option title for enabling/disabling the IPFS proxy"
```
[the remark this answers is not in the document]
alanshaw: Sure 👍

add-on/src/contentScripts/ipfs-proxy/page.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+const Ipfs = require('ipfs')
+const { createProxyClient } = require('ipfs-postmsg-proxy')
+const _Buffer = Buffer
+
+window.Buffer = window.Buffer || _Buffer
+window.Ipfs = window.Ipfs || Ipfs
```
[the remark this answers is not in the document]
alanshaw: Maybe you want an IPFS node configured with particular swarm peers so using the one provided by the extension is not possible? ...and you don't want to download the IPFS lib because you're on a poor connection?

Maybe...for experiments?

Are they good enough?

add-on/src/lib/ipfs-proxy/index.js
```diff
@@ -0,0 +1,153 @@
+'use strict'
+/* eslint-env browser */
+
+const browser = require('webextension-polyfill')
+const { createProxyServer, closeProxyServer } = require('ipfs-postmsg-proxy')
+const AccessControl = require('./access-control')
+
+// These are the functions that require an allow/deny decision by the user.
+// All other exposed IPFS functions are available to call without authorization.
```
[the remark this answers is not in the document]
alanshaw: I agree and I'm happy to switch, but can this be done in a separate PR so we can get this merged?

Just FYI, bumping js-ipfs alone wouldn't expose new functions, you'd have to also add the new functions to ipfs-postmsg-proxy for them to become available.

package.json
```diff
@@ -48,8 +54,11 @@
     "fs-promise": "2.0.3",
     "geckodriver": "1.10.0",
     "husky": "0.14.3",
+    "ignore-styles": "^5.0.1",
```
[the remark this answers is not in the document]
alanshaw: Yep, no problem

add-on/src/lib/ipfs-proxy/index.js
alanshaw: Yep, as I also mentioned in the description, the workaround for Firefox right now will be to go to the "manage permissions" page and change the "deny" to "allow".

I will send a separate PR for a custom confirm dialog with options to allow/deny all.

add-on/src/lib/ipfs-proxy/access-control.js
alanshaw: So, just to confirm our keys would look like `'ipfsProxyAcl.http://google.co.uk` for example?

Currently things look sort of like this:

```js
{
  ipfsProxyAcl: {
    'http://google.co.uk': {
      'files.add': true
    },
    'https://ipfs.io': {
      'files.add': false,
      'object.new': false
    }
  }
}
```

...and we'd be altering it to be something like:

```js
{
  'ipfsProxyAcl.http://google.co.uk': {
    'files.add': true
  },
  'ipfsProxyAcl.https://ipfs.io': {
    'files.add': false,
    'object.new': false
  }
}
```

I like this and I think it makes things simpler, the only issue I can think of is getting the entire ACL for the manage permissions page might be tricky - we'd have to read everything in storage into memory and filter just the keys we need.

I think that's a better trade of though since atm we're reading the entire ACL into memory for every `getAccess` and `setAccess`, which will be far more commonly called than `getAcl`.

add-on/src/contentScripts/ipfs-proxy/page.js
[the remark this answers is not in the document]
lidel: I'd say yes, it would make developer's life easier.  👍 
But perhaps it should be renamed  to `window.ipfsApiClient` or something like that?

@diasdavid, would appreciate your thoughts on this from `js-ipfs-api` perspective

add-on/src/lib/ipfs-proxy/index.js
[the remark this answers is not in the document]
lidel: I think it can be a separate PR.
As long as the checkbox enabling window.ipfs is disabled by default, we can merge it. 👍

add-on/src/lib/ipfs-proxy/access-control.js
lidel: Sounds good. 
I think its the only thing that blocks merging this  PR, other things can be addressed in separate PRs

add-on/src/lib/ipfs-proxy/index.js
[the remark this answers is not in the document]
alanshaw: Yep, window.ipfs checkbox is disabled by default

alanshaw:
@lidel is it good to go now?

lidel:
Yes! 💯 Fantastic  work @alanshaw! 🚀  ✨

add-on/src/lib/ipfs-proxy/index.js
lidel: I've noticed and found how Greasemonkey displays confirmation dialog (under Firefox):

https://github.com/greasemonkey/greasemonkey/blob/9ce5dabc74496fa6a631d991302ad4fd80730f1d/src/bg/user-script-detect.js#L76-L99

What is also interesting is Android detection via `chrome.runtime.getPlatformInfo` :)

PR #335, opened 2017-12-13
Nice "Share files via IPFS" with file drop.

olizilla:
Add a nice upload page layout and add repo and peer stats.

screenshot 2017-12-12 18 02 48

TODO:

- [x] Wire up repo stats
- [x] Wire up file drop


lgierth:
Looks pretty good! -- the peers number is misleading though, we share this data with *many* more nodes, it's just that they're not currently connected to us.

About the `repo/stats` command, it can be quite slow since it has to scan the whole blockstore.

olizilla:
@lgierth thanks! I think having a couple of stats helps keep it interesting. 

I looked at what station shows https://github.com/ipfs-shipyard/station#a-menubar-ipfs-application-to-get-you-on-the-distributed-web

I've not got much in my repo, so it's probably not a good test of the perf on stats requests. Would it be reasonable to campaign for stats to be cached inside ipfs core, so that we can visualise them in UIs without worrying about it slowing down with repo size / usage?

We currently show the same peer count on the badge on the browser add on button. Have you got any thoughts on what could make it less misleading. I'm ok with showing the user how many nodes they can see right now, even if it's not the whole story... my intention was for it to be a simple way to make new users feel the p2p joy.

add-on/src/popup/quick-upload.js
```diff
@@ -47,23 +71,37 @@ function quickUploadStore (state, emitter) {
...
+      <div class="mw8 center pa3 white">
+        <header class="flex items-center">
+          ${logo({
+            size: 80,
+            path: '../../icons',
+            heartbeat: false
+          })}
+          <div class="pl3">
+            <h1 class="f2 fw5 ma0">${browser.i18n.getMessage('panel_quickUpload')}</h1>
+            <p class="f3 fw2 lh-copy ma0 light-gray">Currently sharing <strong class="fw6">${filesCount}</strong> files with <strong strong class="fw6">${peerCount}</strong> IPFS peers</p>
```
lidel: What if we reword it a bit to indicate that we talk about 'active' connections? I am not a native speaker but maybe something like:

> Currently sharing 3389 files (and|while) connected to 1038 IPFS peers.

add-on/src/popup/quick-upload.js
```diff
@@ -47,23 +71,37 @@ function quickUploadStore (state, emitter) {
...
+          </div>
+        </header>
+
+        <label for="quickUploadInput" class='db relative mv5 hover-inner-shadow' style="border:solid 2px #6ACAD1">
+          <input class="db absolute pointer w-100 h-100 top-0 o-0" type="file" id="quickUploadInput" onchange=${onFileInputChange} style=""/>
+          <div class='dt dim' style='padding-left: 100px; height: 300px'>
+            <div class='dtc v-mid'>
+              <span class="f3 link dim br1 ph4 pv3 dib white" style="background: #6ACAD1">
+                Pick a file
```
lidel: I feel adding i18 keys (locales) should be a part of this PR – otherwise we may miss it and make translators super sad 😿

olizilla: Good point. Sorry, that was lazy of me!

olizilla: _thinking out loud_ I think I generally hold off on i18n until I've got some general approval on a new design.

lidel: Now that you mentioned, I got your point: it may be counter-productive to translate too early (eg. we may change/remove labels and translation work is lost). Especially now that we have multiple people working on multiple things. 

Perhaps we could make the process more efficient by agreeing that _it is okay to commit english inline in places that can change before stable release_, and instead add "localization" step before final non-beta release?

olizilla:
Text is extracted to `messge.json`. I've not duplicated it to the others yet. I'm in favour of leaving that till we're ready to cut a release.

It was useful to pull the english text into the `message.json` though, as it makes inline styling harder, so it's good to get that reality check as soon as possible.

lidel:
Android version looks okay:
>

PR #337, opened 2017-12-14
Refactors options page to use choo

alanshaw:
Cherry picked the relevant commit out of #331 since this change does not depend on the directory viewer branch.

---

\o/ nothing changes visually, but options page now uses choo!...bringing it in line with the other pages

* Modules for each form/section
* Simplifies the code
* Removes `lib/data-i18n.js` as it's no longer used anymore


alanshaw:
@lidel this should be ready to merge now 💃

PR #339, opened 2017-12-16
[WIP] Native feel of Browser Action popups

lidel:
The goal of this PR is to make the little triangle at the top of the popup feel "native" (where possible).
I am not sure if we can make it work across all browsers, so this is kind of an "exploratory PR".

Rationale:  
- Better UX with dark themes
- Native look-and-feel increases  trust <sup>(tm)</sup>
- Future features (eg. https://github.com/ipfs/ipfs-companion/issues/330)  will require us to display permission dialogs.  Lost likely via browser action popup. We want it to look real, not like a phishing attack. (well, we could just use white background, but still..)

## Remaining work

- [x] Firefox
- [ ] Chrome (if possible)
- [ ] Brave (if possible)

### DIFF
--- a/add-on/src/popup/browser-action/header.js
+++ b/add-on/src/popup/browser-action/header.js
@@ -8,7 +8,7 @@ const isJsIpfsEnabled = require('../../lib/is-js-ipfs-enabled')()
 
 module.exports = function header ({ ipfsNodeType, onToggleNodeType, isIpfsOnline }) {
   return html`
-    <div class="pv3 br2 br--top ba bw1 b--white" style="background-image: url('../../../images/stars.png'), linear-gradient(to bottom, #041727 0%,#043b55 100%); background-size: auto 100%">
+    <div class="pv3 ba bw1 bt-0 br-0 bl-0" style="background-image: url('../../../images/stars.png'), linear-gradient(to bottom, #041727 0%,#043b55 100%); background-size: auto 100%">
       <div class="tc mb2" title="${isIpfsOnline ? '' : 'offline'}">
         ${logo({
           size: 52,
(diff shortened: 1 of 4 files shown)


lidel:
# Firefox

## Before

> 2017-12-16-144623_1718x1419_scrot

## After

> 2017-12-16-142252_1146x711_scrot

olizilla:
I think we only added the border to try and make it look consistent across browsers, and to partially re-create the firefox specific styling that used to be there. I'm in favour of tailoring the styles so blends in nicely to each browser, if possible. They each, as always, seem to have their own unique quirks.

alanshaw:
I did it like this in Firefox originally (since Firefox is my main browser), but Chrome/Brave have a 1px white border and doesn't allow you to style the arrow like that. It was simplest to keep them all consistent when I originally did the work but if this looks ok in Chrome/Brave and doesn't complicate things too much then I'm ok with it...I mean I'd prefer it if all the browsers looked like what you're proposing!

alanshaw:
Note that in macOS Firefox the borders have radius. I think this is one of the things that made it complicated because you'd have to remove it from Chrome, and FF Linux...

screen shot 2017-12-18 at 10 17 09

lidel:
Thanks for the feedback!  I noticed the issue under Chrome too -- no matter what I do the white arrow is not styled. Unfortunate. We should avoid browser-specific CSS, so let's park this PR for now.

I feel we will either:

- abandon the idea
- move icon with dark background to the bottom: that way all browsers have seamless arrow in white color

alanshaw:
IMHO I think we should leave it as is for now

lidel:
Yeah, I think we can revisit this when style guide lands, but that will be a different PR. Closing this one.

Issue #318, opened 2017-11-21
Popup render issue in Chrome


alanshaw:
If the popup receives an update than changes it's height whilst it is animating in then it goes bad:

<img width="360" alt="screen shot 2017-11-21 at 20 07 39" src="https://user-images.githubusercontent.com/152863/33094335-b7c5762a-cef7-11e7-9638-3913bb39e1dc.png">

lidel:
Hm was not able to replicate so far, seems to be specific to Chrome.

Is this the same issue as https://bugs.chromium.org/p/chromium/issues/detail?id=457887 ? 
Its over 1 year old, and they prioritize things oddly (eg. [it took 4 years to fix `background-size`](https://bugs.chromium.org/p/chromium/issues/detail?id=110195)), so we may want to address this on our side.

Perhaps we could detect running in Chrome and producing some kind of  additional, delayed update after initial load?

lidel:
@alanshaw mind checking if changes merged with #336 helped by any chance?

jnthnvctr:
Hi @lidel, unfortunately I believe this is still persisting (I'm also on macOS accessing using Chrome).
Chrome version I'm on is: 63.0.3239.84

This thread had a few work arounds people had found for the bug:  https://bugs.chromium.org/p/chromium/issues/detail?id=428044

lidel:
Thanks @jnthnvctr! @olizilla was able to replicate it in https://github.com/ipfs-shipyard/ipfs-companion/pull/354#issuecomment-356303354.

**tl;dr** it is indeed specific to Chrome on macOS ([potential explanation](https://bugs.chromium.org/p/chromium/issues/detail?id=457887#c4))

I am looking for a volunteer with a Mac to go over potential fixes in #355 :)

lidel:
Seems to be fixed in gurrent `master` – https://github.com/ipfs-shipyard/ipfs-companion/pull/355#issuecomment-368478592
Feel free to reopen if happens again.

olizilla:
@lidel it's happening again on the beta-channel chrome release

<img width="425" alt="screenshot 2018-04-04 13 12 02" src="https://user-images.githubusercontent.com/58871/38306971-249dcd58-380a-11e8-9e5f-a333799437fd.png">

also, i don't have permissions to re-open bugs on this repo, plz can haz?

lidel:
@olizilla try to reopen now :)

PR #354, opened 2018-01-07
Browser Action popup fixes

### LINKED ISSUE #338: Error while removing IPFS Pin
Removal of a pin triggers an error that lands in Browser Console and is displayed via system notification.
What is interesting is that the _Pin itself is indeed removed_. 
It looks like some kind of race condition or redundant event being yielded.
I did not investigate it deeper, just creating this issue to do it before release.

Below is a log from Firefox (Quick Upload than unPin):

> 2017-12-14-215958_838x431_scrot

I am using go-ipfs v0.4.13/cc01b7f.

lidel:
- closes #338
- as a side-effect this PR may improve situation with #318 
  - I tested in Chrome under Linux and popup never breaks
  - would be great if someone checked under macOS, as we suspect it to be platform-specific
- it should not break anything under Brave, but l'd sleep better if someone confirmed it works there too – would appreciate a quick smoke-test :-)


olizilla:
👀

olizilla:
Brave: ✅

olizilla:
With some ruthless clicking I still able to get chrome to do the jank render

screenshot 2018-01-09 14 34 09

and I'm kinda proud of this one

screenshot 2018-01-09 14 33 44

but otherwise, all was well. Firefox rendered all smooth and nice no matter how much I clicked and pin/unpin works without error in all.

lidel:
Thanks @olizilla, let's keep #318 open then :)

ps. the last one was glorious!

PR #365, opened 2018-01-31
Add Jenkinsfile with docker-based build

### LINKED ISSUE #364: Jenkins Builds
@victorbjelkholm noted that we should have ipfs-companion automatically built on jenkins, so that people can just download pre-built artifacts if they want to test bleeding edge features. 

### Prerequisites

just any Linux and Node will do

### Key Commands 
Install all deps & build the package: `npm run yarn-build` → `./build/ipfs_companion-*.zip`
Run tests: `npm test`

###  Tasks 

- [ ] set up Jenkins (@VictorBjelkholm ?)
- [ ] add Jenkins file (@lidel)

lidel:
This PR aims to add stable Jenkins-based CI setup + reproducible build.

When done, closes #364

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -1,5 +1,7 @@
 node(label: 'linux') {
-  sh 'npm run docker-build'
-  sh 'npm run docker-test'
-  sh 'archiveArtifacts artifacts: 'build/*.zip''
+  checkout scm
+  sh 'docker build -t ipfs-companion .'
+  sh 'docker run -i ipfs-companion yarn test'
+  sh 'docker run -i -v $(pwd)/build:/usr/src/app/build ipfs-companion yarn build'
+  archiveArtifacts artifacts: 'build/*.zip'
 }
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -1,0 +1,5 @@
+node(label: 'linux') {
+  sh 'npm run docker-build'
+  sh 'npm run docker-test'
+  sh 'archiveArtifacts artifacts: 'build/*.zip''
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM node:9.4.0-alpine
+RUN apk add --update make gcc g++ python
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+COPY package.json /usr/src/app/package.json
+COPY yarn.lock /usr/src/app/yarn.lock
+RUN yarn
+COPY . /usr/src/app
--- a/README.md
+++ b/README.md
@@ -112,6 +112,19 @@ Try manual installation:
         1. Enter `about:debugging` in the URL bar
         2. Click "Load Temporary Add-on" and point it at `add-on/manifest.json`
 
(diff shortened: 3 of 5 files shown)


ci/Jenkinsfile
```diff
@@ -0,0 +1,5 @@
+node(label: 'linux') {
+  sh 'npm run docker-build'
```
lidel: Do we have npm on `$PATH` here? If not I need to inline docker commands instead of relying on npm script.

VictorBjelkholm: We could have but makes things more fragile. Neither yarn or npm handles concurrent builds very well, and npm doesn't even run concurrently very well so if we can skip using npm in the Jenkinsfile itself, it would be better as we can run multiple builds on one machine instead of one build per machine as now.

ci/Jenkinsfile
```diff
@@ -0,0 +1,5 @@
...
+  sh 'npm run docker-test'
+  sh 'archiveArtifacts artifacts: 'build/*.zip''
```
VictorBjelkholm: Should remove the `sh` here as `archiveArtifact` is a method just like `sh`. So the result would be:

```groovy
archiveArtifacts artifacts: 'build/*.zip'
```

olizilla:
I think I created this confusion. We're using docker in a very un-normal way right now. It was a quick hack to verify that we had a build process that was producing the same artefact on different systems, just to confirm that what we were sending mozilla was reproducible, but we didn't migrate to a full, docker based release process.

Our `docker-build` npm script should be removed, and a Dockerfile should be added to the repo; @VictorBjelkholm's suggestion here is a good starting point: https://github.com/ipfs-shipyard/ipfs-companion/issues/364#issuecomment-361895348

I think the general idea is the Dockerfile should create an image with the expected build tools inside (node, npm), and then we need to be able to run commands against that container to run the tests and export the artefact. Happy to chat about it if you're online.

lidel:
Ok, I see we need to inline the docker commands into Jenkinsfile, but, do we need `Dockerfile`? 

I think we can use original `node` image + mount sources + run `npm run yarn-build` and everything will be installed in proper versions and package will be produced in `build/`.  Same with tests.

VictorBjelkholm:
@lidel we need to build a docker image with the required dependencies and then install our dependencies so we can use it later for running tests and doing the actual build. We could inline all the commands with the arguments and use the container ID from the builds, but having a Dockerfile greatly simplifies the setup.

VictorBjelkholm:
Made some improvements to this branch, it's in this PR: https://github.com/ipfs-shipyard/ipfs-companion/pull/366/files

lidel:
Superseded by #366

lidel:
@VictorBjelkholm  merged your PR into this one, check now, all looks good :)
I switched to Jenkins being mandatory check for new PRs to `master`, TravisCI will be now optional. 

Question: should I remove TravisCI integration as a part of this PR, or do we want to keep it for now as a backup?

VictorBjelkholm:
@lidel keep it for a while, just to make sure everything is OK. If Travis passes but Jenkins does not, please report the issue to me (unless you can figure it out of course)

PR #366, opened 2018-02-01
Extract into own Dockerfile

VictorBjelkholm:


VictorBjelkholm:
@lidel this works now and is ready to be merged into your `jenkins-ci` branch. We end up with a nice little zip file in our artifacts: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%20Shipyard%2Fipfs-companion/detail/PR-366/5/artifacts

README.md
```diff
@@ -112,6 +112,17 @@ Try manual installation:
         1. Enter `about:debugging` in the URL bar
         2. Click "Load Temporary Add-on" and point it at `add-on/manifest.json`
 
+You can also do the build via Docker. Run the following commands for ending up
+with a built extension inside the `build/` directory.
+
+```sh
+docker build -t ipfs-companion .
+docker run -it ipfs-companion yarn test # Make sure all tests are passing before building
+docker run -it $(pwd)/build:/usr/src/app/build ipfs-companion yarn build
```
olizilla: should this be `docker run -i -v` as per the Jenkinsfile?

VictorBjelkholm: @olizilla yeah, could be. We don't really need pseudo-tty but it's handy when you want to use a bash prompt for example. I would keep it, as it's handy for debugging to just replace `yarn build` with `bash` or `sh` and you can inspect how things are inside a running container. Doesn't really break anything either.

lgierth: There's definitely a `-v` missing though

VictorBjelkholm: Most certainly, thanks @lgierth

ci/Jenkinsfile
```diff
@@ -1,5 +1,7 @@
 node(label: 'linux') {
-  sh 'npm run docker-build'
-  sh 'npm run docker-test'
-  sh 'archiveArtifacts artifacts: 'build/*.zip''
+  checkout scm
+  sh 'docker build -t ipfs-companion .'
+  sh 'docker run -i ipfs-companion yarn test'
```
lidel: Note to self: we may want to move tests after build when we have integration ones that run in real browser, but for now its fine as is 👍

VictorBjelkholm: Hm, I don't think so, we want to make sure that the application is correct before building it. A alternative solution could be to do unit tests + building sequentially in one pipeline and then run functional tests in it's own. If the functional tests fails, stop the other pipeline. 

We need to make some basic assertions before building though.

Issue #306, opened 2017-11-13
Reproducible Build


lidel:
Post-submission review process at Mozilla kicked in for [ipfs-companion](https://addons.mozilla.org/en-US/firefox/addon/ipfs-companion/) and  requires us to provide source code + build steps that will produce exactly the same code as one published to NPM:

> The following information is needed to complete your review:
> 
> 1) This version contains obfuscated, minified, concatenated or otherwise machine-generated code. Please provide the original sources, together with instructions on how to generate the final XPI. Source code must be provided as an archive and uploaded using the source code upload field, which can be done during submission or on the version page in the developer hub.
> 
> Please read through the instructions at https://developer.mozilla.org/en-US/Add-ons/Source_Code_Submission. If sources have not been provided within 7 days, this version will be rejected and removed from our queues.
More Information: https://blog.mozilla.org/addons/2016/04/05/improved-review-time-with-links-to-sources/
> 
> Established libraries (eg: JQuery, bootstrap) do not require the inclusion of their sources.
Please use non-custom versions of jquery-ui, as well as the stock version of jquery and others, that match the checksums in
https://github.com/mozilla/amo-validator/blob/master/validator/testcases/hashes-allowed.txt

**Problem 1** 
I am unable to replicate build results for two of four dependencies: `ipfs-api` and `is-ipfs`.

I downloaded https://github.com/ipfs/js-ipfs-api/releases/tag/v15.0.1 but was unable to produce the same file as https://unpkg.com/ipfs-api@15.0.1/dist/index.min.js -- after running `npm install && npm run build` minified code looks different.   Same goes for `is-ipfs`.

This may be due to different version of `node`, `npm` or lack of `package-lock.json` / `yarn.lock`.
Mozilla Reviewer is okay with using specific versions, as long as we provide step-by-step build instruction that result in correct file, matching the one from NPM.

**Problem 2** 
We have 7 days to provide "sources and build steps", otherwise [ipfs-companion](https://addons.mozilla.org/en-US/firefox/addon/ipfs-companion/) will be removed.

@diasdavid @dignifiedquire @fbaiodias – any ideas on how to address this?

dignifiedquire:
Created a test branch: https://github.com/ipfs/js-ipfs-api/pull/626 please test

olizilla:
[...]

PR #368, opened 2018-02-01
Fixes window.ipfs is added after page load

### LINKED ISSUE #362: window.ipfs injected after page loaded
Great work so far on `window.ipfs`, it's amazing to have IPFS embedded directly in the pages without having to build or do anything! Looking forward to perfecting it.

One issue I hit was with my test page from my previous experiments: https://github.com/VictorBjelkholm/js-ipfs-in-the-browser/blob/master/testpage.html

I've made two versions. One is the original which checks if `window.ipfs` is defined, and then does a `ipfs.id()` call. This does not work as the `window.ipfs` has not been injected when that code is being executed. https://ipfs.io/ipfs/QmW9EQSoL4WSPXy8UwZJjJ7xmrhiEY4UzWZ85LYSvf3e4j

A second version I made introduces a `setTimeout` of `500` milliseconds. This one works, as the script has had time to inject and load the script before the code is being executed. https://ipfs.io/ipfs/QmY49HH5pABacU5gSdXB12vQim6U2oXrTVefD3oTMvRCQv

If you look at my previous example, I define in the `manifest.json` that the script should be injected at the `document_start` (https://github.com/VictorBjelkholm/js-ipfs-in-the-browser/blob/de2bdc82d3dbe05060814ea94c666e6c3d290c76/manifest.json#L21), so pages will instantly be able to access `window.ipfs`web. I'm guessing ipfs-companion is not doing this, as we don't want to inject `ipfs.window` if the option is not turned on.

My thinking is that we should rather go the `manifest.json` way, and inject the content-script always, but only do the expose if the option is turned on. This would solve the problem.

alanshaw:
fixes #362

also fixes whitelist by default from [review feedback](https://github.com/ipfs-shipyard/ipfs-companion/pull/333#pullrequestreview-91336031)

see https://github.com/ipfs-shipyard/ipfs-companion/issues/362#issuecomment-362231167 for discussion

screen shot 2018-02-01 at 16 44 03


add-on/src/lib/ipfs-proxy/acl-whitelist.json
```diff
@@ -0,0 +1,30 @@
...
+  "object.stat",
+  "pin.ls",
+  "pubsub.subscribe",
+  "pubsub.unsubscribe",
+  "pubsub.peers",
+  "pubsub.ls",
+  "swarm.peers",
+  "swarm.addrs",
+  "swarm.localAddrs"
+]
```
alanshaw: Too permissive?

VictorBjelkholm: This is a whitelist right? Any other calls will by default be blocked? Think they are all fine, but I do worry about for example MFS being totally open for everyone with access to window.ipfs.

lidel: I'd remove all `*.ls` from whitelist for now. Adding / reading specific hash can be done without explicit permission, but listing internal node state should be require confirmation.

add-on/src/lib/options.js
```diff
@@ -13,7 +13,7 @@ const optionDefaults = Object.freeze({
   customGatewayUrl: 'http://127.0.0.1:8080',
   ipfsApiUrl: 'http://127.0.0.1:5001',
   ipfsApiPollMs: 3000,
-  ipfsProxy: false
+  ipfsProxy: true
```
lidel: I agree it is easier to keep it enabled during development, but we need to switch it off by default before this PR is merged :)

package.json
```diff
@@ -44,6 +46,7 @@
     "babel-preset-es2015": "6.24.1",
     "babel-preset-es2017": "6.24.1",
     "babelify": "8.0.0",
+    "brfs": "^1.4.4",
```
lidel: `^` :))

alanshaw: sorry!

VictorBjelkholm:
Tried it out, worked beautifully! Nice to see a IPFS app whose size is only 2.2K 👯

yarn.lock
```diff
@@ -5003,7 +5343,7 @@ level-js@^2.2.4:
     typedarray-to-buffer "~1.0.0"
     xtend "~2.1.2"
 
-level-js@timkuijsten/level.js#idbunwrapper:
+"level-js@github:timkuijsten/level.js#idbunwrapper":
```
VictorBjelkholm: Weird, somehow this changed from being a npm link (I think?) to being a github link, requiring git and therefore breaking the docker image.

alanshaw:
@lidel good to merge now?

yarn.lock
lidel: Fixed it in 1a968ee. More context: 

> Same thing happened to me after a `yarn cache clean` (trying to get a git branch dependency to update).
> 
> Removing the `node_modules` folder and running `yarn` again seemed to work around the issue.
> – https://github.com/yarnpkg/yarn/issues/2083#issuecomment-272424677

alanshaw:
eek, wtf?! I'm guessing that's because we're adding `window.Ipfs` as well. I think we could get rid of half of that by removing `dist/contentScripts/ipfs-proxy/page.js` after `dist/contentScripts/ipfs-proxy/content.js` is generated (page.js is inlined into content.js so isn't actually used). Would that be acceptable or should we remove `window.Ipfs` altogether?

lidel:
I am afraid removing `page.js` is not enough:  `content.js` itself is >4MiB. 
We need to remove `window.Ipfs` for now, or find a way to go under linter limit.

alanshaw:
I can bring `content.js` down to 2MB if I uglifyify page.js (this includes `window.Ipfs`) - is that cool?

lidel:
Yes, we can minimize the output as long we have reproducible builds (#306).

alanshaw:
@lidel what did you do to fix the yarn issue?

lidel:
@alanshaw  removing the `node_modules` folder and running yarn again seemed to work around the issue (I did it on a linux box, if that matters. idea from: https://github.com/yarnpkg/yarn/issues/2083#issuecomment-2724246770)

alanshaw:
🙄 that's what I did...

lidel:
Did you run via `npm run yarn-build`  to guarantee correct yarn is used? 
Perhaps you have an older version installed globally. 🤔

alanshaw:
I did yes. I'm not sure how we can avoid this when our dependencies have github dependencies e.g.

https://github.com/libp2p/js-libp2p-crypto/blob/71339e08e75f6c92485e22a95caf96164cbb1cad/package.json#L45

alanshaw:
@lidel I'm not sure your commit in https://github.com/ipfs-shipyard/ipfs-companion/commit/1a968eef41afd634622d7cd920bcdcf8a74d5e75 fixed the build - I still see lots of `github:` dependencies in that file.

lidel:
@alanshaw I've just added git to the image for now and it fixed the build.
(CI is being reworked in #372, but i did not want to port that PR here, just fixed alpine image so I can merge this PR)

PR #381, opened 2018-02-11
UX fixes around embedded js-ipfs and automatic mode

### LINKED ISSUE #373: Daemon is off but redirect is still enabled
I'm not being able to go to ipfs.io urls because the extension keeps redirecting me, even though the daemon is off.

image

Is this intended? 

I'm using version 2.1.0
image

### LINKED ISSUE #380: Pinning does not work on Firefox when using js-ipfs node
Tested with 888d3cf
> 2018-02-10-162729_1720x711_scrot

> 2018-02-10-162559_752x427_scrot

We should either fix it or hide pin controls for now.

lidel:
I started dog-fooding embedded mode in Firefox and fixed some UX bugs related to it.

- [x] suspend redirect to custom gateway (and automatic mode) when embedded node is active
- [x] redirect to public gateway instead of custom one when running js-ipfs
  > This works fine in Firefox/Chrome, but may need additional work in separate PR to get optimal UX under Brave (eg. we probably want to redirect IPFS resources to `ipfs://` instead of keeping URL of public gateway).  
- [x] fix UX when starting with external API being offline (close #373)
- [x] extract some reusable feature-detection functions
- [x] open public gateway instead of custom one when uploading a file
  using embedded js-ipfs under Firefox
- [x] separate redirect tests for embedded and external modes
- [x] make it work under Chrome 
- [x] Pinning does not work on Firefox when using js-ipfs node (close #380)
- [x] Move runtime feature detection to `feature-detector.js`  + tests

### Known Issues (to be addressed in separate PR)

- Rube Goldberg machine-like workflow breaks browser action menu #388


lidel:
I was able to make it work in Chrome by adding `const browser = require('webextension-polyfill')` to `add-on/src/lib/state.js` but I get `ReferenceError: chrome is not defined` when I run test suite (`yarn test`).

@alanshaw, @olizilla  did you see this error before by any chance?

alanshaw:
I've not see that before. Stack trace please?

lidel:
@alanshaw I added mentioned change to this PR, full stacktrace is at Jenkins: [PR-381/3/pipeline#step-7-log-10](https://ci.ipfs.team/blue/organizations/jenkins/IPFS%20Shipyard%2Fipfs-companion/detail/PR-381/3/pipeline#step-7-log-10)

lidel:
Fixed it by extracting feature detection code into `feature-detector.js` (0cc1ad1)
I don't like the name, but does the job + is finally testable.  

Do you see anything that could be improved before we merge this?

alanshaw:
I'll take a look at these as soon as I can

alanshaw:
Next time could you split changes for multiple issues like this into multiple PRs? It's much easier for me when reviewing to see which changes relate to which issues when they're separate.

lidel:
Yeah, my bad. I realized there is a size problem around 4th "fix" in this PR. 
It was supposed to be much smaller. :disappointed:

add-on/src/lib/ipfs-request.js
```diff
@@ -43,7 +49,7 @@ function createRequestModifier (getState, dnsLink, ipfsPathValidator) {
       }
       // Detect valid /ipfs/ and /ipns/ on any site
       if (ipfsPathValidator.publicIpfsOrIpnsResource(request.url)) {
-        return redirectToCustomGateway(request.url, state.gwURL)
+        return redirectToIpfsTransport(request.url, state, featureDetector)
```
alanshaw: I don't know what a `IpfsTransport` is, how about just `redirectToGateway` (could be custom or public)

add-on/src/lib/ipfs-request.js
```diff
@@ -55,7 +61,9 @@ function createRequestModifier (getState, dnsLink, ipfsPathValidator) {
 
 exports.createRequestModifier = createRequestModifier
 
-function redirectToCustomGateway (requestUrl, gwUrl) {
+function redirectToIpfsTransport (requestUrl, state, featureDetector) {
+  // TODO: redirect to `ipfs://` if inBrowserWithNativeProtocol() === true
+  const gwUrl = featureDetector.embeddedNodeIsActive() ? state.pubGwURL : state.gwURL
```
alanshaw: Nitpick but, I vote to keep this function simple, as it was, and perform the feature detection before it is called.

add-on/src/lib/ipfs-companion.js
```diff
@@ -251,23 +249,23 @@ module.exports = async function init () {
   }
 
   // TODO: feature detect and push to client type specific modules.
-  function getIpfsPathAndLocalAddress (hash) {
+  function getIpfsPathAndNativeAddress (hash) {
     const path = `/ipfs/${hash}`
-    if (state.ipfsNodeType === 'embedded' && browser && browser.protocol && browser.protocol.registerStringProtocol) {
-      return {path, localAddress: `ipfs://${hash}`}
+    if (featureDetector.embeddedNodeIsActive() && featureDetector.inBrowserWithNativeProtocol()) {
```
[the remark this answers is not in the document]
olizilla: Good question. Logically, if we've been able to registerStringProtocol handler then `ipfs://` addresses should work. I think when I wrote this I was only using the embedded js-ipfs node in the registerStringProtocol handler function, but now we support using either. It should be safe to remove it.

lidel:
@alanshaw  simplified things just like you sugested in https://github.com/ipfs-shipyard/ipfs-companion/pull/381#pullrequestreview-97812294:

- state-based checks are inlined
- remaining runtime checks are no longer functions but precomputed booleans

Ok to merge?

alanshaw:
🚀 :thumbsup:

PR #382, opened 2018-02-13
Adds custom proxy access dialog

### LINKED ISSUE #377: Custom dialog for window.ipfs ACLs (make it work under Firefox)
Gist from @alanshaw's TODO  list in PR https://github.com/ipfs-shipyard/ipfs-companion/pull/333#issue-157965898:

> * Add a custom confirm dialog.  Firefox does not support `window.confirm` from the background process so it fails - **all access to `window.ipfs` on Firefox will be automatically denied**, however as a temporary workaround, Firefox users can use the manage permissions page to enable permissions after they have been auto denied
>    * Another reason this is needed is that we want to be able to grant access to all (or a subset of) the API for a given origin, `window.confirm` can't cover these UX requirements

It may be useful to note how other extensions handle confirmation dialogs  (repost from https://github.com/ipfs-shipyard/ipfs-companion/pull/333#discussion_r163708569): 

> I've noticed and found how Greasemonkey displays confirmation dialog (under Firefox):
> 
> https://github.com/greasemonkey/greasemonkey/blob/9ce5dabc74496fa6a631d991302ad4fd80730f1d/src/bg/user-script-detect.js#L76-L99
> 
> (HTML page loaded in new browser window with `'type': 'popup'`)

alanshaw:
* `window.confirm` is not supported when called from a background script in Firefox
* We need a mechanism to grant allow/deny to all permissions for a given origin

screen shot 2018-02-13 at 15 55 25
screen shot 2018-02-13 at 15 53 25

resolves #377


package.json
```diff
@@ -72,7 +72,7 @@
     "sinon": "4.1.2",
     "sinon-chrome": "2.2.1",
     "standard": "10.0.3",
-    "uglifyify": "^4.0.5",
+    "uglifyify": "4.0.5",
```
lidel: @alanshaw  I think you forgot to update yarn.lock. If I run `npm run yarn-build` it modifies it:
```diff
--- a/yarn.lock
+++ b/yarn.lock
@@ -4374,7 +4374,7 @@ ipfs-block@^0.6.1, ipfs-block@~0.6.1:
   dependencies:
     cids "^0.5.2"
 
-ipfs-css@^0.1.0:
+ipfs-css@0.1.0:
   version "0.1.0"
   resolved "https://registry.yarnpkg.com/ipfs-css/-/ipfs-css-0.1.0.tgz#0337e7e3aeddf391bbd1b7b7fb427f98b6ec5ca3"
 
@@ -9315,7 +9315,7 @@ uglify-to-browserify@~1.0.0:
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/uglify-to-browserify/-/uglify-to-browserify-1.0.2.tgz#6e0924d6bda6b5afe349e39a6d632850a0f882b7"
 
-uglifyify@^4.0.5:
+uglifyify@4.0.5:
   version "4.0.5"
   resolved "https://registry.yarnpkg.com/uglifyify/-/uglifyify-4.0.5.tgz#49c1fca9828c10a5a8e8d70f191a95f7ab475911"
   dependencies:
```

This is why Jenkins failed:
- https://ci.ipfs.team/blue/organizations/jenkins/IPFS%20Shipyard%2Fipfs-companion/detail/PR-382/1/pipeline#step-14-log-29

alanshaw:
When you check the checkbox I'm implementing something like this:

```js
{
  origin: { '*': true }
}
```

Usually it would look like:

```js
{
  origin: { 'files.add': true, 'object.new': false /* ... */ }
}
```

There's some nuances to `*` that I'd like to discuss before continuing:

1. **What happens when we set access to `*`?**

    * Does it remove any existing permissions that have been granted? e.g. if I've already approved `files.add` and then I am asked about another permission for which I check the checkbox and click allow, does my permission object then look like `{ 'files.add': true, '*': true }` or `{ '*': true }`?

    * If it removes all existing permissions, what happens if I try to set a specific grant? e.g. if I have `{ '*': true }` and then set granted access to `files.add` to `false` should that fail or add the new grant? This is sort of hypothetical, because there's no way in the UI you can do this currently.

2. **What happens when we get granted access to a permission when `*` exists?**    

    * If a `*` grant exists is that the granted permission that the user gets or if there is a more specific permission, should that be returned instead? e.g. `{ '*': true, 'files.add': false }` - what happens when I try to get granted access to `files.add` - do I get `true` or `false`?

Right now I'm leaning towards when you set `*` it removes all existing access grants and if you try to set a more specific access grant it fails - you can't set `files.add: false` if `*: true` (incidentally set `files.add: true` if `*: true` is a noop). Since this most closely matches the UX i.e. the fail will never happen as there is no UI for the user to add a more specific grant if `*` is already granted.

olizilla:
Setting "allow all", I think it should remove more specific permissions, and just store `{ '*': true }`.

If you _hypothetically_ try to set a more specific permission, when `{ '*': true }` is set, then I'd be fine with that being a no-op, in as much as, the the existing permissions already allow what you're asking for, so change nothing in the acl and say ok to the user.

If it should never be possible to set a more specific permission when a wildcard is already set, then an error would be appropriate. Right now that feels like an implicit rule just because of the way the ui is structured, but I don't think it's a rule we really care about.

alanshaw:
@lidel @olizilla @VictorBjelkholm fonts from `ipfs-css` are not yet working but the main body of work here is done and I'd like to get some eyes on it so we can get it merged. Could you please take a look? 

I'll send a separate PR for getting the fonts working.

add-on/src/lib/ipfs-proxy/pre-acl.js
```diff
@@ -18,8 +16,8 @@ function createPreAcl (getState, accessControl, origin, permission, requestAcces
     let access = await accessControl.getAccess(origin, permission)
 
     if (!access) {
-      const { allow } = await requestAccess(origin, permission)
-      access = await accessControl.setAccess(origin, permission, allow)
+      const { allow, remember } = await requestAccess(origin, permission)
+      access = await accessControl.setAccess(origin, remember ? '*' : permission, allow)
```
lidel: Took me a moment to get it.. `remember` does not seem to be a good description for what is happening here :)
Perhaps `wildcard`  would be a more intuitive name than `remember`?

.gitignore
```diff
@@ -11,3 +11,5 @@ crowdin.yml
 add-on/dist
 coverage
 .nyc_output
+add-on/fonts/*
+add-on/*.css
```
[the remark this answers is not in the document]
alanshaw: 👌

add-on/src/lib/ipfs-proxy/pre-acl.js
alanshaw: 👍

.gitignore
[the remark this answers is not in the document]
lidel: Note to self: our landing page (#324,  #326) will re-use styles and fonts, eventually

alanshaw:
@lidel good to merge?

PR #386, opened 2018-02-19
Adds scoped permissions

### LINKED ISSUE #375: Isolate ACL for window.ipfs on IPFS paths (assume HTTP gateway)
As noted by @VictorBjelkholm  in https://github.com/ipfs-shipyard/ipfs-companion/issues/362#issuecomment-362567671: 

> The address for checking the permissions needs to be the full path, but without the hash and query parameters. Otherwise applications running on a gateway gets access to other apps running on the same gateway.

(adding as separate issue so we don't forget about it)

alanshaw:
🎥 https://youtu.be/v6wuFRDKVBU

This PR introduces scope based permissions - following a similar implementation as service worker registration scopes.

This means that allowing access to:
https://ipfs.io/ipfs/QmQxeMcbqW9npq5h5kyE2iPECR9jxJF4j5x4bSRQ2phLY4/

will allow access to:
https://ipfs.io/ipfs/QmQxeMcbqW9npq5h5kyE2iPECR9jxJF4j5x4bSRQ2phLY4/path/to/something

...but not:
https://ipfs.io/ipfs/QmWfVY9y3xjsixTgbd9AorQxH7VtMpzfx2HaWtsoUYecaX

or other hashes.

resolves #375

---

### Scopes

Permissions are scoped to the **origin and path** (and sub-paths) of the file from which the permission was requested.

Scoped permissions in `window.ipfs` work similarly to how they work for [service worker registrations](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register#Examples) except that the user cannot control the scope, and it is set to the origin and path from which the permission was requested.

Scope based permissions allow applications running on an IPFS gateway to be granted different permissions. Consider the following two web sites running on the ipfs.io gateway:

* https://ipfs.io/ipfs/QmQxeMcbqW9npq5h5kyE2iPECR9jxJF4j5x4bSRQ2phLY4/
* https://ipfs.io/ipfs/QmTegrragyzfFq6DSuUaPYoKzm4eRBj2tgQaDHC72dLLaV/

With [same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy) these two applications would be granted the same permissions. With scoped permissions, these applications will be given a different set of permissions.

e.g.

* Allow `files.add` to `https://domain.com/`
    * ...will allow `files.add` to:
        * `https://domain.com/file`
        * `https://domain.com/file2.html`
        * `https://domain.com/sub/paths`
        * `https://domain.com/sub/paths/files`
        * etc.
* Allow `files.add` to `https://domain.com/feature`
    * ...will allow `files.add` to:
        * `https://domain.com/feature/file`
        * `https://domain.com/feature/file2.html`
        * `https://domain.com/feature/sub/paths`
        * `https://domain.com/feature/sub/paths/files`
        * `https://domain.com/featuresearch/sub/paths/files` (note substring)
        * `https://domain.com/features.html` (note substring)
        * etc.
    * ...will cause additional prompt for `files.add` to:
        * `https://domain.com/`
        * `https://domain.com/files`
        * etc.

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -413,13 +413,8 @@
     }
   },
   "page_proxyAccessDialog_wildcardCheckbox_label": {
-    "message": "Apply to all permissions for $SCOPE$",
-    "description": "Label for the apply permissions to all checkbox (page_proxyAccessDialog_wildcardCheckbox_label)",
-    "placeholders": {
-      "scope": {
-        "content": "$1"
-      }
-    }
+    "message": "Apply this decision to all permissions in this scope",
+    "description": "Label for the apply permissions to all checkbox (page_proxyAccessDialog_wildcardCheckbox_label)"
   },
   "page_proxyAcl_revoke_all_button_title": {
     "message": "Revoke all permissions",
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -355,22 +355,22 @@
     "description": "Message displayed when no permissions have been granted (page_proxyAcl_no_perms)"
   },
   "page_proxyAcl_confirm_revoke": {
-    "message": "Revoke permission $PERMISSION$ for $ORIGIN$?",
-    "description": "Confirmation message for revoking a permission for an origin (page_proxyAcl_confirm_revoke)",
+    "message": "Revoke permission $PERMISSION$ for $SCOPE$?",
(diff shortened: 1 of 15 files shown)


lidel:
While I was at it, I made ACL dialog work on Android :sparkles: 

> 8fekisdl png

Issue #307, opened 2017-11-14
Provide API for other browser extensions


lidel:
[Interesting idea for WebExtension ecosystem](https://discourse.mozilla.org/t/extensions-providing-apis-to-other-extensions/21555?u=lidel):

> I think extensions working together by providing APIs to each other is a really cool thing. APIs can either be a `runtime.onMessageExternal` (or `onConnectExternal`) protocol or a global injected into a website that other extensions and user scripts can interact with to augment functionality.
> 
> I’ve started a list with documented APIs (I know of a few more, but no documentation for them, at least yet): https://github.com/freaktechnik/awesome-extension-apis

Quick brain dump:
- In theory, ipfs-companion could provide an API for other extensions
    - additional incentive for installing the add-on
    - decrease complexity of other extensions
- As a PoC we could provide a small sub-set of `interface-ipfs-core`
    - eg. read-only `files.get`,  `swarm.peers` etc.  
    - think about security considerations 
      (eg. should we avoid mutable actions until we have some kind of ACL?)
- Other extensions should be able to use the built-in management API to check if the ipfs-companion is installed, before request is sent, or just implement correct error handling

olizilla:
This seems at least partially related to #330 in as much as we're talking about how much of the ipfs api we can expose, and how to do so safely.

lidel:
Indeed, these two efforts will greatly overlap: everything we expose under `window.ipfs` should also be available to other extensions. 

The main difference would be authorization:
* Is a pop-up confirmation dialog enough/correct UI for this? 
* Should it grant a permanent permission to a specific extension? Does time-based one make any sense in this context?

lidel:
* Access to `ipfs.files` should be sandboxed the same way as when accessed via `window.ipfs` 
  - chroot could be created at `/dapps/webext/<extension_uuid>/`

travisperson:
This seems like it would be pretty simple to add this to an extension via the Ports. Create a client from `ipfs-postmsg-proxy`, and passing in custom options to communicate over the port would do it for the most part I believe.

Would need to update `getScope` to inspect the incoming port and pull out the extension `id` if it's there for scoping as mentioned above.

> Should it grant a permanent permission to a specific extension? Does time-based one make any sense in this context?

[...]

PR #391, opened 2018-02-22
[WIP] Companion chat

alanshaw:
This PR adds a chat function that all companion users can use to chat about IPFS.

It uses `ipfs.pubsub` and subscribes to a topic called `ipfs-companion:chat`. Messages are `cbor` encoded and look like `{ name, text }`.

screen shot 2018-02-22 at 12 02 20

screen shot 2018-02-22 at 10 14 56

### DIFF
--- a/add-on/src/lib/ipfs-client/embedded.js
+++ b/add-on/src/lib/ipfs-client/embedded.js
@@ -10,8 +10,11 @@ exports.init = function init () {
   node = new Ipfs({
     config: {
       Addresses: {
-        Swarm: []
+        Swarm: ['/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star']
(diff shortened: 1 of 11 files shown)


VictorBjelkholm:
Hm, my first reaction is, do we really want this in the same extension? Might be feature-bloat.

Otherwise, is it automatically starting when installing? In that case, we might want to have it under a preference-checkbox.

alanshaw:
No worries, it is just for fun. I just PR'd it in case anyone wanted to play. I'm going to make a demo/example from it to show off `window.ipfs`.

lidel:
@alanshaw this would also make a fantastic PoC/demo  for https://github.com/ipfs-shipyard/ipfs-companion/issues/307 (_Provide API for other browser extensions_) :)

PR #395, opened 2018-02-22
Allows embedded IPFS node to be configured

### LINKED ISSUE #394: Configure embedded IPFS node
There's currently no way to configure the embedded IPFS node. I want to enable the pubsub experiment but cannot.

alanshaw:
resolves #394

Simply adds a field in the options (if node type is embedded) that allows you to paste some JSON config. This allows the savvy user to configure their js-ipfs node as they like.

screen shot 2018-02-22 at 14 52 38

I've also removed the `is-js-ipfs-enabled` module as it is always enabled now (sorry that should probably be in another PR but is a small change so hopefully not too distracting).

This PR also upgrades choo to fix this problem with textareas: https://github.com/choojs/choo/issues/632


alanshaw:
@lidel can we deal with the linting warning as a separate issue please?

alanshaw:
@lidel @olizilla I added this because I needed to enable pubsub on the js-ipfs node, but this is real power user stuff and is error prone for new users. Would it be better to just have a checkbox to enable pubsub instead and leave the JSON config to the power users who are probably running their own daemon anyway?

lidel:
### Linter
Rule of thumb: we should never break build of `master`. It introduces friction for other people that want to work on own PRs.  

I propose we park this PR until linting is fixed in separate PR, and then you can rebase this PR on top of that. 

### JSON and UX
Indeed, a field with JSON in it is error prone. 
But I am not sure if we want to create a precedent of adding control for pubsub.  

Let's keep JSON-based config, but make sure the "Reset Everything" button restores config to safe defaults.  
That way of user breaks it, there is a way of fixing it.

We should also either add "(for Power Users)" to label description, or move it to "Experiments" section.

olizilla:
On the subject of config... I say keep it simple. It's be helpful as a dev to know that I can build an app that will run if the user installs ipfs-companion as it has a fairly predictable config. They don't have to install a local go node or desktop, the defaults in companion will work. 

I'm not sure who we are catering for in making an easy to use embedded node also be "power-user" configurable. A simpler story would be "embedded node with defaults" for the casual user, "external node with custom config" or just, "create an app that bundles it's own custom ipfs node" if your app or use case has custom requirements.

I'd be happy to start in that direction, and then see if anyone starts demanding more configurability of the embedded node and review.

alanshaw:
@lidel I'm happy for this PR to be blocked by the lint issue 👍 

I appreciate that I opened this PR but I agree with @olizilla on this one - I think the stories should be casual -> embedded, expert -> external. I think JSON config is at odds with the casual embedded story.

Here's the thing. If we don't have JSON config _and_ we don't have an enable pubsub checkbox then people **can't use pubsub with `window.ipfs`**.

Why?

Pubsub is disabled in `js-ipfs-api` when used in browsers. If you run your own daemon configured to pubsub then we have to use `js-ipfs-api`.

Embedded js-ipfs is not configured to enable the pubsub experiment by default.

lidel:
Hm.. now that I think about it, I'd say pubsub is marked as an experiment by js-ipfs for a reason :) 

I propose we keep the "embedded node JSON config field" field in current form, but move it to  to "Experiment" section (and update label to explicitly state that the config is for embedded node only).

That way that there will be no confusion that this feature is an opt-in experiment that can changed removed in future.

alanshaw:
@lidel rebased and lint error is now gone 😄

lidel:
Not sure if related to this PR, but also found: #407

alanshaw:
Ok so, you're saying that `"rekt"` as a swarm value throws when we create a new IPFS instance?

We can code around that particular issue, but it's going to be difficult to guard against all config issues. I can imagine a situation where users accidentally stop their node from communicating on the network or even cause it to error when particular methods are called due to the config they've copy/pasted from the internet or changed themselves. I don't think that's something we can, or even should be trying to guard against...like I said above, it's error prone and I think a checkbox to enable pubsub would be a safer option for casual users.

As a side note, is the reset button not working for you? It should reset back to the default config.

lidel:
I agree we can't (and we should not) guard against broken config. 
My point was that Reset button does not work as expected (the real expectation is "make it all work again"). I am sorry, should emphasize that better :)

### Replication steps
1. Set config with invalid `"Swarm": [ "rekt" ]` → node breaks
1. Press <kbd>Reset Everything</kbd> → the config is restored  to defaults
   (`Swarm` value on _Preferences_ screen is restored to `[ ]` ) 
1. .. but the node does not work. Even if you toggle between embedded and external node API client does not start correctly ([logs](https://user-images.githubusercontent.com/157609/36699467-ff8e863a-1b4c-11e8-93fd-a1b2a9d48f07.png)) and user is left with broken extension without  a way of fixing it. 

I would really like to keep this "full config" feature, as it encourages playful experimentation, but we need to make it possible to self-heal.

alanshaw:
@lidel ok I'm on it! 👍 

btw, how do you format a button in gfm?

lidel:
`<kbd>K</kbd>` → <kbd>K</kbd> :)

alanshaw:
Blocked by https://github.com/ipfs/js-ipfs/issues/1193#issuecomment-368843378

cvan:
@alanshaw hi there! it looks like this won't be blocked anymore thanks to your work in R ipfs/js-ipfs#1239 (though not upstreamed to a [patch release yet](https://github.com/ipfs/js-ipfs/pull/1239#issuecomment-372338949)). just wanted to make sure I'm following everything correctly. is that all correct? nice work, btw!

alanshaw:
Thanks, I think we're just waiting on it to be released now @cvan

alanshaw:
@lidel I've taken another look at this, updated the IPFS dependency and fixed the stability issues you were seeing.

There is now **validation of config** and **feedback to the user** and there's always **a way to reset config** and the extension acts gracefully when it fails to start an IPFS node.

I'm happy for this to merge if you are.

add-on/_locales/en/messages.json
```diff
@@ -169,6 +169,14 @@
     },
     "description": "A generic placeholder for error notification (notify_inlineErrorMsg)"
   },
+  "notify_startIpfsNodeTitle": {
+    "message": "Failed to start IPFS node",
```
lidel: We should change key names to reflect that message is for a failure:
- `notify_startIpfsNodeTitle`
- `notify_stopIpfsNodeTitle`

add-on/src/options/forms/experiments-form.js
```diff
@@ -87,6 +90,17 @@ function experimentsForm ({
           </label>
           <input type="checkbox" id="ipfs-proxy" onchange=${onIpfsProxyChange} checked=${ipfsProxy} />
         </div>
+        ${ipfsNodeType === 'embedded' ? html`
+          <div>
+            <label for="ipfsNodeConfig">
```
lidel: Now that we can safely recover any misconfiguration, lets move this back to the old place – below node type picker :)

add-on/_locales/en/messages.json
alanshaw: ha, oops I didn't mean to not indicate it was for a failure!

alanshaw:
@lidel looks good to you?

lidel:
:ok_hand:  
Thanks!

PR #402, opened 2018-02-26
Applies uglifyify transform in build:js script

alanshaw:
Resolves lint error for `ipfs-companion-common.js`


lidel:
Oops, `-g uglifyify` in `build:js` introduced #405

PR #411, opened 2018-03-05
Adds an example for fallback when window.ipfs is unavailable

alanshaw:
In the interests of enabling developers...


lidel:
fyi I noticed a broken link, I fixed it as a part of #412

PR #413, opened 2018-03-05
Fixes ipfs.swarm.peers call when using js-ipfs-api

alanshaw:
Currently the return type for `ipfs.swarm.peers` is different depending on whether we're talking directly to a js-ipfs or js-ipfs-api ([issue opened here](https://github.com/ipfs/js-ipfs/issues/1248)).

This PR updates the proxy to take this into account so that it doesn't error when passed a `PeerId` where it is expecting a `PeerInfo` (https://github.com/tableflip/ipfs-postmsg-proxy/commit/eb0aa43416ad79c4a84b0aaa48a035f54d4d6dae).

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "ipfs": "0.27.7",
     "ipfs-api": "18.1.1",
     "ipfs-css": "0.2.0",
-    "ipfs-postmsg-proxy": "2.8.3",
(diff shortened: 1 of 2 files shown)


diasdavid:
It would be great to actually fix the issue https://github.com/ipfs/js-ipfs/issues/1248

PR #359, opened 2018-01-22
Protocol handlers without web+ prefix in Firefox 59

lidel:
This PR implements changes described in https://github.com/ipfs-shipyard/ipfs-companion/issues/164#issuecomment-355708883, namely support for simplified (redirect-based) protocol handlers for  `dweb:` and `ipfs://` without `web+` prefix.  

It works perfectly fine with latest Firefox Nightly (59a1), but fails to load in Firefox <59:

```
$ web-ext run -s add-on/ --browser-console --verbose
[firefox/index.js][debug] Firefox stdout: 1516628484552	addons.webextension.<unknown>	ERROR	Loading extension 'null': Reading manifest: Error processing protocol_handlers.3.protocol: Value "dweb" must either: be one of ["bitcoin", "geo", "gopher", "im", "irc", "ircs", "magnet", "mailto", "mms", "news", "nntp", "sip", "sms", "smsto", "ssh", "tel", "urn", "webcal", "wtai", "xmpp"], or match the pattern /^(ext|web)\+[a-z0-9.+-]+$/
```

Due to this and a false-positive error in addon-linter (https://github.com/mozilla/addons-linter/issues/1783) I am parking this PR until Firefox 59 is released to Beta (it may even wait til Stable).

PR #422, opened 2018-03-13
Use constant publicGatewayUrl to avoid string repetitions and to facilitate updates

VanVan:
Even the default option is already defined, better to use it :)

### DIFF
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -75,7 +75,7 @@ function redirectToGateway (requestUrl, state) {
 // PROTOCOL HANDLERS: web+ in Firefox (protocol_handlers from manifest.json)
 // ===================================================================
 
-const webPlusProtocolHandler = 'https://ipfs.io/web%2B'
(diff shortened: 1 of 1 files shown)


VanVan:
I understand, but seems to be better to change a URL in two files than in more in case of update :)
IPFS success will break bandwidth of one public gateway one day :)

lidel:
No worries: before final request is sent, we swap URL of public gateway to the one from user options:

https://github.com/ipfs-shipyard/ipfs-companion/blob/bd3368dc544ba23938131393c8bb153be892d717/add-on/src/lib/ipfs-request.js#L90

This means even tho we use `ipfs.io` internally (due to `protocol_handler` API limitations), it won't be used for final request if user choose to use different gate. So no bandwidth will be wasted. 

But thank you for pointing this out, I agree it is not intuitive.
We should make the purpose of `const webPlusProtocolHandler` self-explanatory.

I suggest closing this PR, or parking it until #359 is merged.

lidel:
Closing this in favor of #501
Thanks again for raising this ambiguity  to our attention!

PR #427, opened 2018-03-21
fix: proxy access dialog display issue on Firefox

alanshaw:
I noticed this was still happening despite the previous "fix". I move the code for the fix into the page JS and it worked.


alanshaw:
I'm on 60 right now...I think somewhere along the line I switched to the beta channel...

lidel:
Yeah a lot of people switched to Beta for some reason.. :smiley:  
- https://bugzilla.mozilla.org/show_bug.cgi?id=1445672

PR #428, opened 2018-03-21
Decouple Page Action Updates

### LINKED ISSUE #426: Context update locks the UI when on "pin-in-progress" IPFS Path
I was tracking this bug for some time, and finally was able to reproduce.
If a big CID is being pinned, then UI is broken when opened while on the pinned IPFS Path:

> 2018-03-19-175247_1720x1421_scrot

A potential fix would be to do pin-status checks asynchronously.

lidel:
This PR aim to address the fact calculating pageActions for IPFS Paths is expensive (especially pin-related checks when pin is in progress).

- fixes and closes  #426 
  - calculating pageActions states is expensive (especially pin-related checks when pin is in progress)
    - pageActions are updated in a separate step to keep UI snappy
- experiment: add 800ms fade-in for dynamically added UI elements
  - it may be subjective, but it eased the feel of 'erratic updates' a bit
    - people do not focus on updates happening during fade-in 
  - we can revert this change later, or replace basic fade-in with better animation
  - demo (gif may render slower than realtime, but gives a good idea what is happening): 
    > peek 2018-03-21 13-52

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -87,10 +87,6 @@
     "message": "Non-IPFS resource",
     "description": "Default label for icon hidden in Page Action menu (pageAction_titleNonIpfs)"
   },
-  "pageAction_statusPlaceholder": {
-    "message": "(checking status...)",
-    "description": "Placeholder label for status header in Page Action menu (pageAction_statusPlaceholder)"
-  },
   "contextMenu_uploadToIpfs": {
     "message": "Upload to IPFS",
     "description": "An item in right-click context menu (contextMenu_uploadToIpfs)"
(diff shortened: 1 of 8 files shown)


alanshaw:
Fade ins look great btw. Probably better in a separate PR next time!

lidel:
Good catch with queued calls (https://github.com/ipfs-shipyard/ipfs-companion/pull/428#issuecomment-375289079). I've added: 

https://github.com/ipfs-shipyard/ipfs-companion/blob/cb3da2bb78e860554dabcefdd16acd9173b423f8/add-on/src/popup/browser-action/store.js#L206-L208

I agree that in the long term we probably need to move pin logic to the background page (and do expensive updates via a separate message) but that is a deeper change, separate PR.

Is it ok to merge this PR now?

lidel:
Removed the _"(checking status...)"_ from header of `pageAction`, it how shows instantly. 
This PR paired with changes from #430 makes UI feel really snappy, let's merge this.

PR #429, opened 2018-03-21
Fix options_ui in OSX Chrome

lidel:
This PR aims to fix UI bug reported by @alanshaw:

screen shot 2018-03-21 at 11 33 10

## Notes on `options_ui` in Chrome
- fix is to either specify `min-width` and `overflow` controls, or explicitly open `options_ui` page in a new tab
  - `open_in_tab` is [being deprecated](https://developer.chrome.com/extensions/optionsV2#step-1), so we would have to do it manually
    > screenshot-2018-3-21 options - google chrome

### DIFF
--- a/add-on/src/options/options.css
+++ b/add-on/src/options/options.css
@@ -1,3 +1,9 @@
+html {
+  /* allocate proper width in Chrome */
+  min-width: 650px;
+  overflow-x: hidden;
+  overflow-y: auto;
+}
 form * {
   font: caption;
   font-size: 1em;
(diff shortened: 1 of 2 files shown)


alanshaw:
@lidel did this work for you? Mine still doesn't scroll.

I've had to add width and height to the body element to get scrolling working again. However that messes with how it's displayed on Firefox. So, I added a detection which adds an `is-chrome` class to the `<html>` element so that we can apply the styles only on chrome.

VictorBjelkholm:
Note: the last commit that triggered a jenkins build was cancelled as I'm upgrading jenkins + workers. You'll see the build failed for now but as soon as everything is back, I'll retry this build. Sorry for the inconvenience

lidel:
@alanshaw I failed to mention that my Chrome looked fine all the time  :) problem seems to be specific to OSX build of Chrome. Thanks for fixing it!

I noticed and fixed [double scrollbar](https://user-images.githubusercontent.com/157609/37781416-f3ab6e12-2df0-11e8-9564-fba2a35e2d9d.png) in Firefox, should not impact your fix for OSX, but please double check.  

Is it ready to merge?

alanshaw:
Yep, all good :thumbsup:

PR #430, opened 2018-03-22
fix: time till browser action dropdown appears

alanshaw:
`factor-bundle` pulls common dependencies out from our page scripts to minimise download size of the extension.

This is a good thing since we don't want to include multiple versions of large dependencies such as IPFS (for example).

However, it introduces the overhead that every page using the common bundle has to read and parse it, even if some of the dependencies aren't used.

For the browser action, the only "significant" shared dependency is `choo`, which is only 4kb. The browser action is reading a ~4mb JS file just for a single dependency. _This_ is what is causing the short delay in opening the dropdown.

I've taken the browser action script out of the factor bundle and built it separately, meaning that we have a duplicate `choo`, but I think it's worth the 4kb hit for the performance improvement.

Try it out, it is 🐎 💨


lidel:
Wohoo, indeed, we are snappy again! :rocket: :sparkles: :1st_place_medal: 

I applied the same fix for page-action and quick-upload screens, which also displayed slight delay.
To minimize size, I created a separate shared bundle for use in GUI pages.

What is interesting now is that most of heavy "backend" code stays embedded directly in the background page (+ content scripts related to proxy):

```
-- ipfs-companion/add-on/dist --
    3.1 MiB [##########] /background
    2.6 MiB [########  ] /contentScripts
  136.0 KiB [          ]  ipfs-companion-gui-common.js
  116.0 KiB [          ] /popup
   68.0 KiB [          ]  ipfs-companion-common.js
   64.0 KiB [          ] /pages
   60.0 KiB [          ] /options

--- ipfs-companion/add-on/dist/popup ---
   68.0 KiB [##########] /browser-action
   24.0 KiB [###       ] /page-action
    4.0 KiB [          ]  quick-upload.js
    4.0 KiB [          ]  logo.js
    4.0 KiB [          ]  heartbeat.css
    4.0 KiB [          ]  quick-upload.html
    4.0 KiB [          ]  quick-upload.css
```

I did not switch options and acl/proxy pages as there is no visual slowdown and I think proxy-related  pages pull in ipfs-postmsg-proxy anyway, which defeats the purpose of a lean bundle for snappy UIs.

@alanshaw  is there anything you would like to do here, or are we good for merge?  
I am really excited for this to land! :)

PR #431, opened 2018-03-23
feat: scope mfs calls to an application path

alanshaw:
https://youtu.be/KvgxAafphPg

See https://github.com/ipfs-shipyard/ipfs-companion/issues/330#issuecomment-367651787 for discussion.

To avoid conflicts, each app gets it's own MFS directory where it can store files. When using MFS functions (see [docs](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#mutable-file-system)) this directory will be automatically added to paths you pass. Your app's MFS directory is based on the **origin and path** where your application is running.

e.g.

* `files.write` to `/myfile.txt` on `https://domain.com/`
    * writes to `/dapps/https/domain.com/myfile.txt`
* `files.write` to `/path/to/myfile.txt` on `https://domain.com/feature`
    * writes to `/dapps/https/domain.com/feature/path/to/myfile.txt`
* `files.read` from `/feature/path/to/myfile.txt` on `https://domain.com/`
    * reads from `/dapps/https/domain.com/feature/path/to/myfile.txt`
* `files.stat` to `/` on `https://domain.com/feature`
    * stats `/dapps/https/domain.com/feature`
* `files.read` from `/../myfile.txt` on `https://domain.com/feature`
    * reads from `/dapps/https/domain.com/feature/myfile.txt` (no traverse above your app's root)

### DIFF
--- a/add-on/src/lib/ipfs-proxy/pre-mfs-scope.js
+++ b/add-on/src/lib/ipfs-proxy/pre-mfs-scope.js
@@ -1,0 +1,65 @@
+const Path = require('path')
+const DEFAULT_ROOT_PATH = '/dapps'
+
+// Creates a "pre" function that is called prior to calling a real function
+// on the IPFS instance. It modifies the arguments to MFS functions to scope
+// file access to a directory designated to the web page
+function createPreMfsScope (fnName, getScope, rootPath = DEFAULT_ROOT_PATH) {
+  return MfsPre[fnName] ? MfsPre[fnName](getScope, rootPath) : null
+}
+
+module.exports = createPreMfsScope
+
+const MfsPre = {
+  'files.cp': srcDestPre,
+  'files.mkdir': srcPre,
+  'files.stat': srcPre,
+  'files.rm': srcPre,
+  'files.read': srcPre,
+  'files.write': srcPre,
+  'files.mv': srcDestPre,
+  'files.flush': optionalSrcPre,
+  'files.ls': optionalSrcPre
+}
+
+// Scope a src/dest tuple to the app path
+function srcDestPre (getScope, rootPath) {
+  return async (...args) => {
+    const appPath = await getAppPath(getScope, rootPath)
+    args[0][0] = appPath + safePath(args[0][0])
+    args[0][1] = appPath + safePath(args[0][1])
+    return args
+  }
+}
+
+// Scope a src path to the app path
+function srcPre (getScope, rootPath) {
+  return async (...args) => {
+    const appPath = await getAppPath(getScope, rootPath)
+    args[0] = appPath + safePath(args[0])
+    return args
+  }
+}
+
+// Scope an optional src path to the app path
+function optionalSrcPre (getScope, rootPath) {
+  return async (...args) => {
+    const appPath = await getAppPath(getScope, rootPath)
+    if (Object.prototype.toString.call(args[0]) === '[object String]') {
+      args[0] = appPath + safePath(args[0])
+    }
+    return args
+  }
+}
+
+// Get the app path for a scope, prefixed with rootPath
+const getAppPath = async (getScope, rootPath) => rootPath + scopeToPath(await getScope())
+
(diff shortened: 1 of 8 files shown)


olizilla:
This looks great, I love the video. Can we special case the https / http level to remove the colon before url-encoding? We're going to end up presenting these to the user in the new UIs and having the url-encoded version is gonna look surprising.

add-on/src/lib/ipfs-proxy/pre-mfs-scope.js
```diff
@@ -0,0 +1,101 @@
...
+
+// Get the app path (create if not exists) for a scope, prefixed with rootPath
+const getAppPath = async (getScope, getIpfs, rootPath) => {
+  const appPath = rootPath + scopeToPath(await getScope())
+  await getIpfs().files.mkdir(appPath, { parents: true })
+  return appPath
+}
+
+// Turn http://ipfs.io/ipfs/QmUmaEnH1uMmvckMZbh3yShaasvELPW4ZLPWnB4entMTEn
+// into /http%3A/ipfs.io/ipfs/QmUmaEnH1uMmvckMZbh3yShaasvELPW4ZLPWnB4entMTEn
```
lidel: I agree with https://github.com/ipfs-shipyard/ipfs-companion/pull/431#issuecomment-375748250,  we should add `.replace(':\/\/','/')` at the beginning of the `scopeToPath` pipeline to change first (and only first) occurrence of `://` into `/`

add-on/src/lib/ipfs-proxy/pre-mfs-scope.js
```diff
@@ -0,0 +1,101 @@
...
+// Creates a "pre" function that is called prior to calling a real function
+// on the IPFS instance. It modifies the arguments to MFS functions to scope
+// file access to a directory designated to the web page
+function createPreMfsScope (fnName, getScope, getIpfs, rootPath = DEFAULT_ROOT_PATH) {
+  return MfsPre[fnName] ? MfsPre[fnName](getScope, getIpfs, rootPath) : null
+}
+
+module.exports = createPreMfsScope
+
+const MfsPre = {
```
lidel: Question; how will it behave when a new `ipfs.files.foo` is added ? 
Will it not work, or will it work with no path scoping?
If the latter is the case,  we should explicitly block executions of new methods  in scoped `ipfs.files` API to avoid surprises in future.

alanshaw: Yeah, I mean, you'd need to update the `ipfs-postmsg-proxy` dependency to allow the new method to be used in the first place, but no it wouldn't be scoped automatically unless you did some work here.

lidel: Indeed, being dependent on `ipfs-postmsg-proxy` makes this safe enough for now.
Unless its trivial to add, I propose we add additional safeguards in separate PR.

alanshaw:
@lidel good now?

hacdias:
@alanshaw will we have a special case for WebUI? WebUI should be able to access everything. How will we do that?

alanshaw:
@hacdias yes, if accessed _through_ companion we can privilege the `window.ipfs` object to not scope MFS or ask permission. We can do the same with desktop if we load WebUI into it.

If you're just loading (our new, yet to be built) WebUI from the gateway in the browser with ipfs companion installed (i.e. https://ipfs.io/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ) then MFS stuff _will_ be scoped. You might want to load from /ipns/webui.ipfs.io to get around this, and also the problem of different versions of the app being scoped to different folders.

Check out this comment (and the backstory) for a bit more context: https://github.com/ipfs-shipyard/ipfs-companion/issues/330#issuecomment-376517818

hacdias:
@alanshaw gotcha. It makes sense. I like the `/ipns/webui.ipfs.io` idea. Always the same link **and** we could easily update the version without waiting for go-ipfs.

PR #433, opened 2018-03-23
fix: object count

alanshaw:
* Is now camel case: `NumObjects` -> `numObjects`
* Means that quick upload page now shows "and sharing 9497 files" again
* Removes `state.repoSize` as it's unused

### DIFF
--- a/add-on/_locales/de_DE/messages.json
+++ b/add-on/_locales/de_DE/messages.json
@@ -309,18 +309,6 @@
     "message": "Browser-Erweiterung, die den Zugriff auf IPFS-Ressourcen vereinfacht",
     "description": "Extension description in the Manifest file (manifest_extensionDescription)"
   },
-  "quickUpload_subhead_peers_and_files": {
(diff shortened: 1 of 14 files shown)


lidel:
Thank you for noticing this! I was about to merge this, but I feel this PR is a good opportunity to revisit two things :thinking:

##  1) What SHOULD we show on "Share" screen?

Having peer count at the upload screen adds some value. 
It makes it more engaging and encourages user to upload (by confirming connectivity is up and running):

screenshot_5

If we fix/restore object count, it will look like this:

screenshot_4

And I am not sure if this additional info belongs there or sends the right message these days. 
Especially if user has nothing in `ipfs.files` and is still seeing "hundreds of files" being shared. 

I am really conscious about scaring users with such random info without proper context. 
Perhaps we should remove  object count until we are able to make it clickable to provide more info via WebUI (IPLD Explorer,  File Browser or some kind of "Stats" screen). Peer count cant stay. 

Thoughts?

## 2) do we need to poll for repo stats every 3 seconds all the time?

AFAIK repo stats call is bit expensive, and this data is shown only on the upload screen.

alanshaw:
@lidel repo stats removed

PR #437, opened 2018-04-02
Portuguese Translation [WIP]

hacdias:


hacdias:
@lidel so I think the European Portuguese translation is done. Although, there are some terms I'm not sure how to exactly translate them (such as 'pin') but I based myself on other 'pin' translations I've seen around in Chrome and such. There are some terms I understand conceptually, but that are hard to translate.

Over time, I can improve this translation (which definitely can be).

Issue #347, opened 2017-12-21
Release Checklist for v2.2.0


lidel:
We've been producing betas for a while. I feel there should be a new release at the beginning of a new year with a main goal to deliver recent UI improvements. 

Let's go with a solid checklist this time. It may grow, as I want to create a template for future re-use:

## New Features

- [x] `window.ipfs` (https://github.com/ipfs-shipyard/ipfs-companion/issues/330)
  - [x] Isolate ACL for window.ipfs on IPFS paths (assume HTTP gateway) (#375)
  - [x] Custom dialog for window.ipfs ACLs (make it work under Firefox) (#377)
  - [x] Decide if window.ipfs should be injected into all_frames (#392)

## Pre-Release Checklist

- [x] all tests pass
  - [x] manual check: Firefox for Android
- [x] locales
  - [x] sync translations with [Crowdin](https://crowdin.com/project/ipfs-companion)
  - [x] resolve any issues / questions raised at [Crowdin](https://crowdin.com/project/ipfs-companion)
- [x] verify that `ipfs-postmsg-proxy`,  `ipfs` and `ipfs-api` deps are  up to date
- [x] close [known bugs](https://github.com/ipfs-shipyard/ipfs-companion/issues?utf8=%E2%9C%93&q=is%3Aopen+label%3A%22upstream+bug%22%2Cbug) that are not blocked by something external 
  - [x] error while removing IPFS Pin (https://github.com/ipfs-shipyard/ipfs-companion/issues/338)
  - [x] investigate odd behavior when removing pin under Firefox (https://github.com/ipfs-shipyard/ipfs-companion/issues/338)
  - [x] investigate random error on Android, when running in background (~ once a day)
  - [x] window.ipfs injected after page loaded #362 
  - [x] Daemon is off but redirect is still enabled #373
  - [ ] Restarting ipfs daemon sometimes freezes the browser #360
  - [x] ReferenceError: can't access lexical declaration `offlinePeerCount' before initialization #378
  - [x] Some locales do not load #434
- [x] before release, tag 'release candidate' for users to test against
  - if bugs are found/fixed, do another release candidate

## Post-Release

- [x] do the reproducible build dance
- [x] write Release Notes
- [x] publish
  - [x] AMO 
  - [x] Chrome Store
- [x] write a blogpost at blog.ipfs.io (https://github.com/ipfs-shipyard/ipfs-companion/issues/344)
- [x] save this checklist as a template for future reuse :)

[...]

PR #438, opened 2018-04-03
feat: upgrades ipfs-postmsg-proxy to 2.13.0

alanshaw:
This PR updates `ipfs-postmsg-proxy` to version 2.13.0 which adds support for `ipfs.name.*` and `ipfs.stats.*`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "ipfs": "0.28.2",
     "ipfs-api": "18.2.0",
     "ipfs-css": "0.2.0",
-    "ipfs-postmsg-proxy": "2.13.0",
+    "ipfs-postmsg-proxy": "2.13.1",
     "is-ipfs": "0.3.2",
     "is-svg": "3.0.0",
     "lru_map": "0.3.3",
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "ipfs": "0.28.2",
     "ipfs-api": "18.2.0",
     "ipfs-css": "0.2.0",
-    "ipfs-postmsg-proxy": "2.10.0",
+    "ipfs-postmsg-proxy": "2.13.0",
     "is-ipfs": "0.3.2",
     "is-svg": "3.0.0",
     "lru_map": "0.3.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1502,6 +1502,10 @@ circular-json@^0.3.1:
   version "0.3.3"
   resolved "https://registry.yarnpkg.com/circular-json/-/circular-json-0.3.3.tgz#815c99ea84f6809529d2f45791bdf82711352d66"
 
+class-is@^1.1.0:
+  version "1.1.0"
+  resolved "https://registry.yarnpkg.com/class-is/-/class-is-1.1.0.tgz#9d3c0fba0440d211d843cec3dedfa48055005825"
+
 class-utils@^0.3.5:
   version "0.3.6"
   resolved "https://registry.yarnpkg.com/class-utils/-/class-utils-0.3.6.tgz#f93369ae8b9a7ce02fd41faad0ca83033190c463"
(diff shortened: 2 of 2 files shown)


lidel:
@alanshaw I suspect we have the BigInt bug here again :thinking: 

This does not work in Firefox (no error, no message in console), but works fine in Chrome (`totalIn/Out` printed to the console):

```
ipfs.stats.bw((err, stats) => console.log(stats))
```

Are you able to replicate the problem?

alanshaw:
hmm, I'm not seeing this...

screen shot 2018-04-03 at 20 15 08

alanshaw:
ah, yes I can replicate when talking to js-ipfs

hacdias:
@alanshaw does it talk directly to the core or through js-ipfs-api?

alanshaw:
If the toggle is set to external, it'll talk via js-ipfs-api to a go-ipfs or js-ipfs daemon.
If the toggle is set to embedded it talks directly to a js-ipfs node running in the extension.

hacdias:
@alanshaw @lidel what is the 'BigInt bug' you've had before?

lidel:
@hacdias s/BigInt/big.js/ :) It was also related to stats, so I suspect it might be somehow connected.

See comment at:
- https://github.com/tableflip/ipfs-companion/blob/6e2a790abe7894aa77c8bf550c9db66fd23e162a/add-on/src/lib/ipfs-companion.js#L172-L176
  (the code was removed some time ago, but I was able to dig it up from git history)

alanshaw:
Not here, the issue is that the code has not made it into a release yet. [Current js-ipfs (0.28.2) does not have `stats.bw`](https://github.com/ipfs/js-ipfs/blob/f7cf10b34296e8e6716c10b88ba573e3379a358f/src/core/components/stats.js).

You see `undefined` because the sample code is ignoring the `err` in the callback.

alanshaw:
If a function is **not proxied** then when it is called `undefined is not a function` will be thrown.

There's no way to detect whether the function the user is calling is a legitimate function on the IPFS node across the `postMessage` boundary. For a couple of reasons:

1. There's no listener setup to receive calls to random function names the proxy does not know about
2. If talking to a daemon, the bundled js-ipfs-api module might support the function, but there's no guarantee the daemon that the user is running is a version that supports that function. We can't detect that without calling the function and receiving the error.

If a function **does not exist on the node** but is proxied (i.e. supported by go-ipfs and not js-ipfs) then yes we could send back a "special" error but _only_ if we're talking directly to a js-ipfs node. However we can't do that if the node is running as a daemon and js-ipfs-api supports the function, but the node does not.

In the former case (direct to js-ipfs) the promise will reject with `undefined is not a function` (this is what happened above but you didn't see the error because it was ignored in the callback).

In the latter case (daemon via js-ipfs-api) you'll probably get some other error related to HTTP (hopefully 404).

I don't think there's anything sensible we can do in code here. We just need to keep the proxy and js-ipfs modules up to date.

lidel:
Ok, got it. It will be a part of release process then. 
To ensure we don't forget about it I've added an item to release checklist template ([example](https://github.com/ipfs-shipyard/ipfs-companion/issues/347)). 

----

Going back to `ipfs.stats.bw` specifics.. :man_cook: 

We've identified js-ipfs situation, but I re-tested it again and it also does not work in _Firefox_ with.. _external_ node (go-ipfs v0.4.14).

#### Firefox (silence)

screenshot

#### Chrome (ok): 

screenshot_2


I've run `npm run yarn-build` against this PR to make sure I have everything up-to-date.
@alanshaw  any idea why it worked for you in https://github.com/ipfs-shipyard/ipfs-companion/pull/438#issuecomment-378364748 and not for me?

alanshaw:
@lidel can you log the error on Firefox?

lidel:
No error in page console (nor in global browser console)

I might've missed something, but it looks like callback just does not execute. 
Also, it depends on how I build the extension:
-  `npm run build`  – error described in https://github.com/ipfs-shipyard/ipfs-companion/pull/438#issuecomment-378757864 
-  `npm run watch`  – works ok

The only difference I see is the lack of  `uglifyify`  in `watch` scripts.
And it turns out the use of `uglifyify` in `build:content-scripts:*` is fine.
What causes the problem is`uglifyify` in `build:js:backend` :thinking:   – when removed, error is gone.

alanshaw:
🙄 ok, thanks for the investigation. I'll take another look now

alanshaw:
@lidel would you mind checking now? It's working for me.

PR #355, opened 2018-01-10
[WIP] Fix popup render issue in Chrome on Mac

lidel:
We need to fix https://github.com/ipfs-shipyard/ipfs-companion/issues/318 on our end, as it is unknown when/if potential fixes will land in Chrome:
- https://bugs.chromium.org/p/chromium/issues/detail?id=728174
- https://bugs.chromium.org/p/chromium/issues/detail?id=730958
- https://bugs.chromium.org/p/chromium/issues/detail?id=457887 (1yr old)
- https://bugs.chromium.org/p/chromium/issues/detail?id=428044 (4yrs old)

I compiled a list of potential workariund to try, starting with the most promising one:
- [x] https://bugs.chromium.org/p/chromium/issues/detail?id=428044#c16 (`requestAnimationFrame`)
- [x] https://bugs.chromium.org/p/chromium/issues/detail?id=428044#c13
- [x] https://github.com/gorhill/uMatrix/issues/423#issuecomment-299214607

@olizilla / @alanshaw   give it a try in a spare moment (I don't have a Mac so it is a stab in a dark 🙃 )

PR #441, opened 2018-04-04
Fix popup render issue in Chrome on Mac

### LINKED ISSUE #318: Popup render issue in Chrome

lidel:
This PR is a rebased version of https://github.com/ipfs-shipyard/ipfs-companion/pull/355, hopefully closes #318 

I do not own a Mac, so re-assigning to @olizilla  who noticed it in https://github.com/ipfs-shipyard/ipfs-companion/issues/318#issuecomment-378579068

### DIFF
--- a/add-on/src/popup/browser-action/store.js
+++ b/add-on/src/popup/browser-action/store.js
@@ -48,13 +48,8 @@ module.exports = (state, emitter) => {
     // fix for https://github.com/ipfs-shipyard/ipfs-companion/issues/318
     setTimeout(() => {
       document.body.style.height = window.innerHeight + 1 + 'px'
-      setTimeout(function () {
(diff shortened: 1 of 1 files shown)


alanshaw:
@lidel we didn't need that extra render so I removed. I also simplified the code a little and shortened the timeouts. Looks like the animation in takes 150ms so as long as we set a height on the body before that ends and remove it afterwards then we're good. I found that the popup "jumped around" a bit less with the shorter timeout (but still appeared correctly eventually). Hope that's ok!

lidel:
:+1:  lets merge it and do RC2.

PR #442, opened 2018-04-06
fix: multi permisison dialogs for the same permission

alanshaw:
Fixes https://youtu.be/k4TMpAXToJc

Permission requests are asynchronous so it's possible for a user to request the same permission multiple times before a response is received. This PR deals with this case.

You can use the following code to check this PR is working:

```js
Promise.all([ipfs.id(), ipfs.id()]).then(console.log).catch(console.log)
```

You should see 2 dialogs in master, and 1 dialog in this branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "lru_map": "0.3.3",
     "mime-types": "2.1.18",
     "p-queue": "2.3.0",
-    "piggybacker": "^2.0.0",
+    "piggybacker": "2.0.0",
     "prundupify": "1.0.0",
     "tachyons": "4.9.1",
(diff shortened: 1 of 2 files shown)


alanshaw:
@lidel is the issue that the second (and other) dialogs appear blank?

lidel:
Yeah.

alanshaw:
@lidel fix for the issue is here: https://github.com/ipfs-shipyard/ipfs-companion/pull/443

lidel:
@alanshaw  perfect, tested together with #443 works as expected :+1: will merge this shortly.

alanshaw:
🤦‍♂️ sorry, always with the hats

Issue #455, opened 2018-04-13
Improve error handling for window.ipfs


lidel:
Question from IRC:
> will the UX of `if (err.message === 'User denied access to ipfs.files.add') { // Fallback` be able to be improved?

For a better context, we currently suggest this in `docs/window.ipfs.md`:

```js
if (window.ipfs) {  
  try {
    await ipfs.add(Buffer.from('=^.^='))
   } catch (err) {
     if (err.message === 'User denied access to ipfs.files.add') {
       // Fallback
     } else {
       throw err
     }
   }

   const data = await ipfs.cat('QmS5RzFV9v4fucVtjATPxoxABgEmNhhduykzUbQeGyyS3N')
   console.log(data.toString()) // =^.^=
} else {
  // Fallback
}
```

Initial ideas:

- Perhaps instead of strring we could return a generic error  + the name of denied API in separate field? It would make it easier to handle programmaticallly. 
- We could expose "generic error message" as an additional attribute of window.ipfs object, so that dapps do not need to hardcode any string.

JonKrone:
I also noticed this dev difficulty when reading [the window.ipfs docs](https://github.com/ipfs-shipyard/ipfs-companion/blob/master/docs/window.ipfs.md#error-messages), I agree we should provide some common Error type or utils to detect them.

```javascript
const accessErrorSymbol = Symbol.for('IPFS Access Error') // or smth

const isAccessError = error => accessErrorSymbol in error

function AccessError(message, permission) {
  const error = new Error(message)
  error[accessErrorSymbol] = true
  error.permission = permission
  return error
}

// companion usage:
if (access popup is closed)
  throw new AccessError('Failed to obtain access response to ${permission} at ${scope}', permission)

// user-land:
try {
 const { CID } = await ipfs.files.add(buf)
 await ipfs.pubsub.publish(friendPool, CID)
} catch (err) {
  if (isAccessError(err) && err.permission.match(/pubsub/)) {
    return // non-critical error, continue user flow
  } else {
   // app-specific handling
  }
}
```

I think it's good practice for libraries to use Symbols to prevent naming collisions with consumers but we could alternatively do something else.

Because of the security concerns with revealing the availability of `window.ipfs` in https://github.com/ipfs-shipyard/ipfs-companion/issues/451, I think the `User disabled access to IPFS` error will no longer exist, right? We would only reveal `window.ipfs` if they have that permission turned on?

lidel:
[...]

Issue #444, opened 2018-04-06
Support and reuse ipfs.types and ipfs.util


lidel:
## Background 

I requested something similar some time ago (), and it just landed :rocket: 
Since last week a set of data types and utils is being exposed directly from the IPFS instance under `ipfs.types` and `ipfs.util`:

- [`ipfs.types.Buffer`](https://www.npmjs.com/package/buffer)
- [`ipfs.types.PeerId`](https://github.com/libp2p/js-peer-id)
- [`ipfs.types.PeerInfo`](https://github.com/libp2p/js-peer-info)
- [`ipfs.types.multiaddr`](https://github.com/multiformats/js-multiaddr)
- [`ipfs.types.multibase`](https://github.com/multiformats/multibase)
- [`ipfs.types.multihash`](https://github.com/multiformats/js-multihash)
- [`ipfs.types.CID`](https://github.com/ipld/js-cid)
- [`ipfs.types.dagPB`](https://github.com/ipld/js-ipld-dag-pb)
- [`ipfs.types.dagCBOR`](https://github.com/ipld/js-ipld-dag-cbor)
- [`ipfs.util.crypto`](https://github.com/libp2p/js-libp2p-crypto)
- [`ipfs.util.isIPFS`](https://github.com/ipfs-shipyard/is-ipfs)

Source: https://github.com/ipfs/interface-ipfs-core/pull/252, [SPEC/TYPES.md](https://github.com/ipfs/interface-ipfs-core/blob/0cf94b852ecf22e897ee39f2a9c5250825384950/SPEC/TYPES.md), [SPEC/UTIL.md](https://github.com/ipfs/interface-ipfs-core/blob/0cf94b852ecf22e897ee39f2a9c5250825384950/SPEC/UTIL.md)

### Availability Status

- [x] js-ipfs-api ([v20.0.0](https://github.com/ipfs/js-ipfs-api/releases/tag/v20.0.0)): , https://github.com/ipfs/js-ipfs-api/pull/732
- [ ] js-ipfs (**waiting for release**): https://github.com/ipfs/js-ipfs/pull/1297

## Tasks

- [ ] switch to js-ipfs-api and js-ipfs with support for these endpoints
- [ ] switch to bundled types and utils 
- [ ] remove duplicate dependencies
- [ ] expose `ipfs.types` and `ipfs.util` under `windows.ipfs` (if feasible)

lidel:
[...]

Issue #454, opened 2018-04-13
Refine security and UX constraints for window.ipfs


lidel:
_(This is a continuation of endeavor to expose IPFS API as window.ipfs, started in #330)_

IPFS Companion [v2.2.0 shipped](https://blog.ipfs.io/35-ipfs-companion-2-2-0/) with `window.ipfs` enabled by default. 

It has access controls in place, but is time to refine it it both in terms of **UX** (not nagging user when we can sandbox/scope instead) and **security** (sniffing/spoofing/fingerprinting).

Feel free to provide general feedback in this issue, we will extract specifics when needed.

### Tasks

- [ ] **Document a coherent framework for defining security constraints for window.ipfs**
      It could be just a section in `window.ipfs.md`. The goal is to have a set of guidelines, our 'rule of thumb' to follow when a new API is added. Should simplify decisions such as: should API X be exposed under `window.ipfs`? does it require sandboxing/chroot?
  - [x] Improving security of window.ipfs (low hanging fruits) #484
  - [ ] Support explicit whitelist of exposed APIs https://github.com/tableflip/ipfs-postmsg-proxy/issues/29
- [x] Enable users to disable creation of `window.ipfs`  (Firefox only)
       (Prevent websites from knowing I have IPFS running in my browser) #451
- [x] Chrome Error: Please use a browser with webcrypto support #475 
- [x] Disable window.ipfs.config.{set,replace}. Provide a capabilities api. #478
- [ ] [Improve] Performance of window.ipfs #460
- [x] Privileged API access for WebUI #514

### Open Questions

- [x] Restricting window.ipfs to Secure Context #476
- [ ] Should IPFS paths at different gateways share the same ipfs.files chroot? #453
- [x] How to improve error handling "User denied access to ipfs.foo" #455
- [x] How to decrease number of dialog windows? 
    - Idea to pre-emptively request permissions was suggested in https://github.com/ipfs-shipyard/ipfs-companion/issues/454#issuecomment-381245084

JonKrone:
I have a question about UX. Is there a supported way for an app to pre-emptively request permissions?

Say you `window.ipfs.files.add` a big video file, it takes 2 minutes, and then you want to `pubsub.publish` the CID to some channel. Right now, we would wait to ask for the `pubsub.publish` permissions until after the file has uploaded. This can be a surprising UX because you might have navigated away from the page or otherwise started doing something else, thinking that first all you needed to do.

[...]

PR #458, opened 2018-04-14
Feat: IpfsApiAccessError

JonKrone:
Use a common `Error` type, `IpfsApiAccessError`, when an app is denied access to a window.ipfs api.

Discussed in https://github.com/ipfs-shipyard/ipfs-companion/issues/455.

Remaining:
- [ ] Make this type available somewhere under `window.ipfs`. Perhaps the `ipfs.types` or `ipfs.utils` 
 from https://github.com/ipfs-shipyard/ipfs-companion/issues/444?

This also fixes some `path.sep` related test failures on Windows, lemme know what you think about that.


JonKrone:
> Make this type available somewhere under window.ipfs. Perhaps the `ipfs.types` or `ipfs.utils`
from #444?

Looking for thoughts here.

lidel:
### Browserify issue?

While it seems to work in tests, real life version injected as content script does not seem to produce Error with additional attributes. Try executing this against some website (and deny access to trigger error): 

```js
ipfs.files.add(Buffer.from('seed')).catch(e => console.log(`Error.permission=${e.permission}, scope=${e.scope}, message=${e.message}`))
```
I get:

screenshot_4

Are you able to reproduce?

### Namespace for proxy stuff

The `ipfs.types` sounds like an obvious place, but I am worried about unexpected consequences of polluting original API namespaces with proxy-specific stuff.

Perhaps a safer alternative is to use `window.ipfs.proxy` (idea suggested in https://github.com/ipfs-shipyard/ipfs-companion/issues/454#issuecomment-381278859) and expose it at `ipfs.proxy.types.IpfsApiAccessError`.

@olizilla @alanshaw do we have any strong opinions or preferred conventions for handling this?

alanshaw:
Yes, I posted my comments here https://github.com/ipfs-shipyard/ipfs-companion/issues/455#issuecomment-381562091

alanshaw:
On the posix stuff, I did check this originally because actually it needs to be posix.

The path module as included by browserify is https://github.com/substack/path-browserify/blob/master/index.js which seems to be using the posix versions and actually doesn't include a `path.posix`.

lidel:
@JonKrone to limit the number of moving pieces, let's go with version proposed by @alanshaw  in https://github.com/ipfs-shipyard/ipfs-companion/issues/455#issuecomment-381562091 (no new type, just a self-explanatory property):
```js
if (err.isIpfsProxyAclError) {
```

That way we don't have to change anything on the backend and we don't need to think where to put the new type.

JonKrone:
Hey guys, I'm sorry for the delay! I was totally swamped last week.

I pushed some changes a few days ago, thanks @alanshaw for the solution with ipfs-postmsg.

re: the posix path fix for tests on Windows: I explicitly imported `path-browserify`. We could alternatively manually normalize the paths, let me know what you think.

add-on/src/lib/ipfs-proxy/pre-mfs-scope.js
```diff
@@ -1,4 +1,5 @@
-const Path = require('path')
+// Use path-browserify for consistent behavior between browser and tests on Windows
```
lidel: I think explicitly using `path-browserify` (which provides posix version by default) is bit safer than assuming browserify-ing process will swap the implementation. 

Let's merge this, and if at some point `path-browserify` turns out to be not enough, we can swap it with own implementation / wrapper.

@alanshaw  does it sound ok to you?

docs/window.ipfs.md
```diff
@@ -38,19 +38,19 @@ if (window.ipfs) {
...
-
-   const data = await ipfs.cat('QmS5RzFV9v4fucVtjATPxoxABgEmNhhduykzUbQeGyyS3N')
-   console.log(data.toString()) // =^.^=
+    const [{ hash }] = await ipfs.add(Buffer.from('=^.^='))
+    const data = await ipfs.cat(hash)
+    console.log(data.toString()) // =^.^=
+  } catch (err) {
+    if (err.isIpfsProxyAclError) {
+      // Fallback
+      console.log(':(')
```
lidel: I would replace `console.log(':(')` with `console.log('Unable to get ACL decision from user :(', err)`

PR #461, opened 2018-04-17
feat: updates ipfs-postmsg-proxy to 2.15.0

alanshaw:
This PR upgrades ipfs-postmsg-proxy to 2.15.0 with support for the bootstrap API and the additional (but not yet spec'd) bitswap API methods.

This brings it as close to interface-ipfs-core compatible as it can be right now barring the resolution of https://github.com/ipfs/js-ipfs/pull/1297


lidel:
:+1:

PR #465, opened 2018-04-19
fix: window.ipfs streaming

alanshaw:
There were two issues at play here:

1. `pull-postmsg-stream` was using an old version of `postmsg-rpc` so wasn't getting the data from a message correctly and caused it to error when pulling over web extension port messaging
2. Companion was actively disallowing bi-directional messages. For pull streams to work, messages need to flow in both directions


alanshaw:
Both js-ipfs-api and (to some extent) js-ipfs have bugs that prevent adding files with pull streams:

https://github.com/ipfs/js-ipfs-api/issues/745
https://github.com/ipfs/js-ipfs/issues/1317

Issue #349, opened 2017-12-30
Add support for wrapping quick uploads in a directory.


falsechicken:
Maybe in the form of a checkbox or something in the quick upload UI. Would be nice for my uses because if I share a file to another user using a public gateway the filename is not preserved when they try and save the file. This is a bigger problem on Windows because it relies so much on file extensions.

nunofmn:
Ok, so I want to give a go at this one 👍 

So I will try to summarize everything that is needed here:

 - The [Files.add function](https://github.com/ipfs/interface-ipfs-core) doesn't implement the `wrap-with-directory`option.
 - In [go-ipfs](https://github.com/ipfs/go-ipfs) this functionality is implemented in the _HTTP API_ and _CLI_.
 - In [js-ipfs](https://github.com/ipfs/js-ipfs) this functionality is only implemented in _CLI_. (missing _HTTP API_ and _Core_).

So the following steps are needed:
 - [ ] Add `wrap-with-directory`option to **File.add** in: 
    - [ ] **js-ipfs** - _HTTP API_ and _Core_
    - [ ]  **js-ipfs-api**
 - [ ] Update the tests in [interface-ipfs-core](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#add)
 - [ ] Support option in **ipfs-companion**

I have a working implementation of this option in **js-ipfs** _HTTP API_, and I'm working on adding it to **js-ipfs** _Core_ also.

@lidel from an UX perspective do you think the checkbox field that @falsechicken mentioned is a good option, or could this functionality be activated by default?

lidel:
### Plan A

@nunofmn  sounds good to me 👍 the summary and list of steps are very helpful, thanks!
If you are able to add it to the API, that would be super beneficial to other projects, go for it.

### Plan B

Just for the record, here is a backup plan, if extending API turns out to be not feasible:

> I initially planned to upload file the old way + assemble wrapping directory manually via object API:
> 
> - `file add` → `<file_cid>`
> - `object  new unixfs-dir` → `<dir_cid>`
> - `object path <dir_cid> add-link <filename> <file_cid>`  → `<dir_with_file_cid>`
> - `pin <dir_with_file_cid>`
> 

### Checkbox for wrapping file with directory

I think we should have an explicit checkbox on the bottom of the upload form.
And it should be checked by default because most of users will want to preserve file name.

Reasons for preserving file name by default:
[...]

PR #479, opened 2018-05-16
feat: preserve file name, support sharing multiple files

lidel:
This PR implements #349  by switching to the latest `js-ipfs-api` which enables support for `wrapWithDirectory` option in `ipfs.files.add`.

## UI Preview

I did not want to do a redesign so just added options at the bottom:
screenshot_50

Not sure if these labels are clear enough, any suggestions?

## TODO before merge

**Add**

- [x] support sharing multiple files
- [x] add checkbox for controlling directory wrapping 
- [x] add checkbox for controlling automatic pinning (added while we are at it)
- [x] add locale keys

**Fix**

- [x]  disable controls when embedded node is used
- [x] solve `FILE_TOO_LARGE` error from `mozilla/addons-linter` caused by the size of the latest `js-ipfs-api`

**Optional (can be a separate PR)**

- [ ] wait for https://github.com/ipfs/js-ipfs/pull/1329 to be released (js-ipfs > v0.28.2) and switch to it, check fi wrapping works and unlock it for embedded node


lidel:
@alanshaw @olizilla  UI is temporary, but is there anything that could be improved before we merge it?

olizilla:
"Preserve the filename when adding a single item" is good wording for the label, but it leads to the question, when would I not want that? I vote for just doing the right thing here (wrap) and not troubling the user with the decision, as per https://github.com/ipfs-shipyard/pm-ipfs-gui/issues/15#issuecomment-373936917 I don't think wrapping by default would harm anyone here.

"Recursively Pin added data"... I think we just do this for all data the user adds. I don't see any harm in doing this and not asking the user.

lidel:
@olizilla Yeah it may be just me :upside_down_face:, but I  _often_ share ephemeral images and text snippets that don't need a name (or have an ugly one that  I'd have to change before sharing) nor permanence (I don't' care to keep it alive forever). 

I agree, the defaults are sane,  but I really want to give power user an option to opt-out without need of using commandline, so I just collapsed them under a tiny `upload options »` that reveals them on click:

screenshot_52

olizilla:
"Upload options" is good and clear.

olizilla:
I'll give this a try now

olizilla:
Using unicode glyphs as icons always seems like a good idea, but the cross browser rendering issues will drive you mad. I have tried.

**extra lefty ticks on FF on mac**
screenshot 2018-05-18 11 06 17

**no labels on Chrome on mac**
screenshot 2018-05-18 11 09 53

olizilla:
@lidel i'll see if I can track down why the labels are missing on chrome.

olizilla:
@lidel ah, the chrome label issue went away after a rebuild. meh.

Now the labels are hidden by default, I'm thinking they can be more informative like

- [x] `--wrap-with-directory` _Wrap files in a directory to preserve their filenames._
- [x] `--pin` _Pin files so they are retained when performing garbage collection on your local IPFS repo._
> See: https://ipfs.io/docs/commands/#ipfs-add for more info

lidel:
Switched to pure CSS checkmark that should look the same everywhere, updated labels.
I don't want to focus on it too much as the screen will be redesigned anyway, 
but it should look ok for now:
screenshot_55

@olizilla please check if it does not :fire:  on mac.
I'd like to merge this and release to beta ASAP  :)

add-on/src/lib/ipfs-companion.js
```diff
@@ -274,19 +276,25 @@ module.exports = async function init () {
...
-    result.forEach(function (file) {
+  async function uploadResultHandler (result) {
+    for (let file of result) {
       if (file && file.hash) {
         const {path, url} = getIpfsPathAndNativeAddress(file.hash)
-        browser.tabs.create({
-          'url': url
-        })
-        console.info('[ipfs-companion] successfully stored', path)
+        console.log('uploadResultHandler.file-' + result.indexOf(file), file)
```
olizilla: Remove debug logging or add `[ipfs-companion]` prefix

add-on/src/lib/ipfs-companion.js
```diff
@@ -274,19 +276,25 @@ module.exports = async function init () {
...
+        // open the wrapping directory (or the CID if wrapping was disabled)
+        if (result.length === 1 || file.path === '' || file.path === file.hash) {
+          await browser.tabs.create({
```
olizilla: What's the expected behaviour here? If i add a single file and let it be wrapped, I'm seeing it open the directory not the file. If i add multiple, it shows the directory.

add-on/src/popup/quick-upload.js
```diff
@@ -62,14 +79,41 @@ function quickUploadStore (state, emitter) {
...
+        <label for='pinUpload' class='flex items-center db relative mt1 pointer'>
+          <input id='pinUpload' type='checkbox' onchange=${onPinUploadChange} checked=${state.pinUpload} />
+          <span class='mark db flex items-center relative mr2 br2'></span>
+          ${browser.i18n.getMessage('quickUpload_options_pinUpload')}
+        </label>
+      </div>
+    `
+  }
+  return html`
+    <button class='f6 link dim bn bg-transparent moon-gray dib pa0 pointer ma0' style='color: #6ACAD1' onclick=${onExpandOptions}>
```
olizilla: Add an `mt3` here to get the same vertical space when the button transitions to the options.

add-on/src/lib/ipfs-companion.js
lidel: Yes, that is correct.

If wrapping is enabled or you are uploading more than one file, it will **always open wrapping directory** (the default).
If wrapping is disabled and a single file is uploaded, it will be opened directly. (requires explicit opt-in from user via upload options).
 
I feel it is a good default for now:
- Most of users want to keep file names and as a side effect will get a branded directory listing provided by gateway
  (Until we integrate ideas from https://github.com/tableflip/share-via-ipfs via separate PRs)
- Uploaded file can be a `.zip` file and browser can't show that. 
  (Opening the file itself will just close the tab and start the download, which makes it difficult for user to copy the URL)
- Mobile users are not forced to download the data, they just get directory listing and decide what do do with it (open the specific file, copy link to directory, copy direct link to file etc)

Issue #436, opened 2018-03-31
gateway redirect messes up CORS headers on firefox


fahrradflucht:
Browser: `Firefox 59.0.2`
AddOn Version: `IPFS Companion (RC1@fb21b6d) 2.2.0.8540` or `IPFS Companion 2.1.0`

Given [this webpage](https://gateway.ipfs.io//ipfs/QmYFfqDBW6D69ehZErmMeTNkJks7qWiADtavcbdM5qLQKy):
```HTML
<!DOCTYPE html>
<html>

<body>
    <button onclick="window.fetchFromPublicGateway()">Fetch from public gateway.</button>
    <button onclick="window.fetchFromLocalGateway()">Fetch from local gateway.</button>
    <script>
        window.fetchFromPublicGateway = () => {
            fetch("https://ipfs.io/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y")
                .then(res => res.text())
                .then(console.log)
                .catch(console.log)
        }

        window.fetchFromLocalGateway = () => {
            fetch("http://localhost:8080/ipfs/QmejvEPop4D7YUadeGqYWmZxHhLc4JBUCzJJHWMzdcMe2y")
                .then(res => res.text())
                .then(console.log)
                .catch(console.log)
        }
    </script>
</body>

</html>
```

when clicking on the public gateway button firefox will cancel the request and issue a same-origin-policy warning due to missing CORS headers even though the local gateway has the necessary CORS-headers set which can be checked clicking the other button (be sure to check which origin you are accessing the document when using the IPFS link).

I checked Chrome 65 for comparison and the issue isn't present.

lidel:
Thank you for the sample app, it made reproducing it much easier :+1: 

Upon inspection of `onBeforeRequest` hook Firefox indeed blocks XHR right after redirect:

> ![screenshot](https://user-images.githubusercontent.com/157609/38162850-0e0eedd4-34e9-11e8-94ec-7ec846a155f2.png)

Chrome does not block redirect, second call has updated `url` field:

> ![screenshot_1](https://user-images.githubusercontent.com/157609/38162860-3a783768-34e9-11e8-94b8-bd3c554b0ab1.png)

I'll look into it.

lidel:
It seems to be a bug in Firefox (see findings below).

I filled an upstream bug:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1450965

[...]

PR #494, opened 2018-06-04
Do not redirect XHR that would fail due to CORS bug in Firefox

lidel:
This PR disables gateway redirect for cross-origin XHRs.

- It fixes all the websites broken by [CORS false-positive bug in Firefox](https://github.com/ipfs-shipyard/ipfs-companion/issues/436#issuecomment-378254294) (samples below).
  -  It is a "lesser evil". There will be no gateway redirect for such requests, but at least "IPFS-enabled" websites that do cross-origin requests to different gateways will work.
- I believe we want the same behavior on all browsers, so initial version of this PR conforms to the Firefox limitation in Chrome as well.
  - For: Different handling on different browsers would be a nightmare for a developer, who assumes it "works the same everywhere". 
  - Against: We introduce artificial limitation for browsers other than Firefox.
- The full context is in https://github.com/ipfs-shipyard/ipfs-companion/issues/436#issuecomment-378254294 including a link to the upstream bug.

### Samples

Examples of sites broken in [v2.3.0](https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.3.0) by the bug:
- https://ipfs.github.io/public-gateway-checker/
- http://node2.e-mesh.net/hls/stream.html?gw=https://hardbin.com/ipfs (reported by @darkdrgn2k)
(open Console to see CORS error)

### Open Questions

I'd appreciate some quick feedback, before this gets merged and released: 
1. Is it OK to artificially limit Chrome just for sake of uniformity? 
2. Should we log a warning to the console on every skipped XHR, or is it too much?

### DIFF
--- a/add-on/src/lib/ipfs-companion.js
+++ b/add-on/src/lib/ipfs-companion.js
@@ -58,7 +58,7 @@ module.exports = async function init () {
       onCopyCanonicalAddress: () => copier.copyCanonicalAddress(),
       onCopyAddressAtPublicGw: () => copier.copyAddressAtPublicGw()
     })
-    modifyRequest = createRequestModifier(getState, dnsLink, ipfsPathValidator)
+    modifyRequest = createRequestModifier(getState, dnsLink, ipfsPathValidator, runtime)
     ipfsProxy = createIpfsProxy(() => ipfs, getState)
     registerListeners()
     await setApiStatusUpdateInterval(options.ipfsApiPollMs)
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -49,7 +49,7 @@ function createRequestModifier (getState, dnsLink, ipfsPathValidator) {
     // handle redirects to custom gateway
     if (state.redirect) {
       // Ignore preload requests
-      if (request.method === 'HEAD' && state.preloadAtPublicGateway && request.url.startsWith(state.pubGwURLString)) {
+      if (request.method === 'HEAD') {
         return
       }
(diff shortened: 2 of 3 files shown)


fahrradflucht:
Since this is a severe limitation on what single page applications on ipfs will be able to do in for example an offline situation I think logging a warning every time this happens is a good idea. I think both firefox and chrome provide ways to silence those logs if the are expected.

alanshaw:
If we use the `isFirefox` runtime check and issue a warning (only for Firefox) we wouldn't have to restrict for Chrome (and potentially others). It wouldn't add a great deal of complexity and there's precedent for that sort of thing already in the code base.

In combination with the log line the altered behaviour in Firefox remains visible and the restriction can be removed easily when the issue is fixed. Meanwhile by leaving in the functionality for Chrome you ensure the code that is doing it continues to work as expected.

lidel:
@alanshaw I initially did that, but I got worried about providing different redirect rules for different vendors, which feels like a bad developer experience.

So your suggestion is to limit this workaround to Firefox only and and be very explicit that it is a temporary workaround for an upstream bug limited to Firefox?

lidel:
Alright, switched to this PR to apply the workaround only when in Firefox runtime.
I feel this makes it very safe to merge and release to beta, as it restores broken functionality for Firefox users and does not change anything for Chrome.

lidel:
For anyone interested in this fix, it just got shipped to stable channel as [v2.3.1](https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.3.1)

PR #498, opened 2018-06-11
refactor: remove browserify, switch to webpack

lidel:
This PR switches the build from browserify to webpack.
I tried to make non-invasive and easy to review, so it does not change project layout. 

Immediate gains introduced by this PR:
- more control over build process, no surprises
- time required by subsequent `yarn build` decreased from ~100s to ~30s
- size of window.ipfs content script from 1.1M to 869K
- size of entire add-on/dist from 6.5M to 5.6M
- `yarn watch` produces human-readable code that is easier to debug

Long term goal is to use webpack for vendor-specific builds as well,
replacing current combination of browserify, jq-based json manipulation
and inlined shell scripting with more programmatic approach that
is easier to maintain.

## Tasks

This PR is ready for review & merge if all below are checked: 

- [x] remove browserify from build and deps
- [x] `yarn build` and `yarn watch` work as expected
- [x] there are no `.js` bigger than 4MB (limit introduced by addon-linter)
- [x] UI pages such as browser action have a separate, small bundle
      (we dont want js-ipfs and js-ipfs-api in common bundle, as it slows down UI popups)
- [x] `yarn watch` should generate human-readable code and/or source maps
- [x] `yarn release-build` produces the same output every time 

## Future Improvements 
These things are not included in this PR:
- cleanup final package (eg. ensure source `.js` are not copied to `/dist/`)
- use webpack to generate manifest
- use webpack to generate HTML entry points that load bundles, enabling further size optimizations


olizilla:
👀

webpack.config.js
```diff
@@ -0,0 +1,122 @@
...
+    uploadPage: './add-on/src/popup/quick-upload.js',
+    optionsPage: './add-on/src/options/options.js',
+    proxyAclManagerPage: './add-on/src/pages/proxy-acl/index.js',
+    proxyAclDialog: './add-on/src/pages/proxy-access-dialog/index.js'
+  },
+  optimization: {
+    /*
+    occurrenceOrder: true,
+    minimizer: [
+      new UglifyJsPlugin({
```
olizilla: Any reason why we can't enable uglify here?

webpack.config.js
```diff
@@ -0,0 +1,122 @@
...
+    ]
+  },
+  resolve: {
+    modules: [
+      'node_modules'
+    ],
+    alias: {
+      // These are needed because node-libs-browser depends on outdated
+      // versions
+      // zlib: can be dropped once https://github.com/devongovett/browserify-zlib/pull/18 is shipped
```
olizilla: https://github.com/devongovett/browserify-zlib/pull/18 is shipped 🚢

webpack.config.js
```diff
@@ -0,0 +1,122 @@
...
+      zlib: 'browserify-zlib',
+      // http: can be dropped once https://github.com/webpack/node-libs-browser/pull/41 is shipped
```
olizilla: https://github.com/webpack/node-libs-browser/pull/41 is shipped 🚢

webpack.config.js
lidel: When we run webpack with `-p` it is enabled by default with default options. 
I've left this section in case we want to customize options that are passed to uglify, but defaults seem to works just fine. Will probably remove it before merge.

lidel:
Added changes from #499 to this PR, which brings:

- `add-on/dist/` size decreased to `5.5M`
- shared webpack config as a base for all targets
- separates build targets for background page, GUI and content scripts
- bundled webext polyfill for chrome, removed two calls to `tabs.executeScript`
- detects and rebuilds ipfs proxy content script on changes (`yarn watch`)
- removes browserify libs and other leftovers

PR #499, opened 2018-06-13
refactor: move content script build to webpack

lidel:
This is a PR on top of #498 

- `add-on/dist/` size decreased to `5.5M`
- shared webpack config as a base for all targets
- separates build targets for background page, GUI and content scripts
- bundled webext polyfill for chrome, removed two calls to `tabs.executeScript`
- detects and rebuilds ipfs proxy content script on changes (`yarn watch`)
- removes browserify libs and other leftovers


lidel:
Actually, I'll just merge this into #498, it may be easier to diff there, as this PR does some cleanup.

Issue #491, opened 2018-06-01
New UI for Backend Control


lidel:
### Tasks

- [x] define what is missing and what elements need to be added to UI
- [x] design drop-down and toggle (browser_action)
- [ ] design backend list editor (Preferences) – see notes below
- [ ] implement changes
  - [x] move node type switch to Preferences
  - [x] ON/OFF toggle  https://github.com/ipfs-shipyard/ipfs-companion/issues/491#issuecomment-398828326
  - [ ] backend list editor (Preferences)
- [ ] migrate existing settings to a new item in "backend list"
- [ ] close all issues in [related milestone](https://github.com/ipfs-shipyard/ipfs-companion/milestone/14)

-----

### Background

<details>
  <summary>Click to expand historical notes on why this change was proposed</summary>

Right now we have a simple toggle between _Embedded_ node (browserified js-ipfs) and _External_ node (usually go-ipfs at IP and port [defined in Preferences](https://user-images.githubusercontent.com/157609/40840535-6c2136cc-65a7-11e8-83af-003737ce4ab9.png), accessed over HTTP API):

> ![screenshot_7](https://user-images.githubusercontent.com/157609/40840515-52ca52d0-65a7-11e8-98c7-ead6fe248600.png)

### Limitations
- No obvious one-click to "enable/disable" what IPFS Companion does. 
    - Multiple users raised the need for "off" switch.
    - Removal of all redirects of IPFS paths and `window.ipfs` object during web development requires going over multiple checkboxes in Preferences or disabling the entire extension via Addon Manager (was the case when troubleshooting [CORS bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1450965)).
   - No clear indication when "power-user ipfs mode" (with local node, API and `window.ipfs`)  is enabled, and when  "vanilla public" (no API) gateway is used. This makes troubleshooting of various issues harder than necessary.
- Switching between "backend" different nodes is a chore, one needs to change URLs in _Preferences_  each time.
   - Users raised a need for making this easier, namely having a predefined list of backends that one can pick. MetaMask network-picker was given as an example of what is missing in Companion: 
      > ![screenshot](https://user-images.githubusercontent.com/157609/40844298-d9736a0e-65b3-11e8-95fb-c1e89bf4d93d.png)
[...]

Issue #450, opened 2018-04-12
High CPU usage


samuelmaddock:
I installed the Chrome extension earlier today and found it to have high CPU usage. I don't think I was using any IPFS-compatible websites besides checking the ipfs.io blog.

Chrome: 65.0.3325.181
IPFS Companion: 2.2.0

<img width="574" alt="screen shot 2018-04-12 at 3 31 48 pm" src="https://user-images.githubusercontent.com/1656324/38699857-460d4370-3e67-11e8-859d-6745da03fad6.png">

lidel:
Thank you for reporting this. Was it running against external or embedded node?

samuelmaddock:
> Thank you for reporting this. Was it running against external or embedded node?

Embedded node

lidel:
Embedded node uses js-ipfs, which is not as mature as go-ipfs. I was able to confirm problem under Chromium: after a few minutes with embedded node the CPU usage is indeed [very high](https://user-images.githubusercontent.com/157609/38740392-b8a6c494-3f37-11e8-9a6b-e5edca4f6d65.png).

Let's keep this issue open until **upstream issues** are resolved:

- High CPU when publishing files with js-ipfs https://github.com/ipfs/js-ipfs/issues/1190
- Connection Closing https://github.com/ipfs/js-ipfs/issues/962

Also related:
- IPFS light clients https://github.com/ipfs/notes/issues/206

----

### Workarounds for lowering CPU usage

For the time being I suggest using go-ipfs which requires smaller amount of resources.
If you run on laptop or other mobile device, try additional settings suggested in https://github.com/ipfs/go-ipfs/issues/4137#issuecomment-322557758, namely:

> - you probably want to use `ipfs daemon --routing=dhtclient`
> - Its probably also worth turning off the reprovider with: `ipfs config Reprovider.Interval "0"` so your node doesnt rebroadcast all this things it has every 12 hours.

gedw99:
Why not just use the new WASM functionality of golang ?
It would keep the J's and golang versions aligned

lidel:
@gedw99 "just" is bit too optimistic :) We may investigate that route, but my intuition is that it won't be easy. There are low details (such as lack of raw socket API in web browser, storage) that require custom handling etc. Also, there is a huge benefit in having two independent implementations interop in the wild:  it verifies if the spec is clear enough.

[...]

PR #500, opened 2018-06-15
Feature: ON / OFF Toggle

lidel:
### TL;DR

> wait, we had no on/off switch before? :thinking: 

peek 2018-06-15 00-59

### Background

As noted in #491, multiple users raised the need for "off" switch:

> - No obvious one-click to "enable/disable" what IPFS Companion does. 
>     - Multiple users raised the need for "off" switch.
>     - Removal of all redirects of IPFS paths and `window.ipfs` object during web development requires going over multiple checkboxes in Preferences or disabling the entire extension via Addon Manager (was the case when troubleshooting [CORS bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1450965)).
>    - No clear indication when "power-user ipfs mode" (with local node, API and `window.ipfs`)  is enabled, and when  "vanilla public" (no API) gateway is used. This makes troubleshooting of various issues harder than necessary.

Another issue: a toggle between _External_ (js-ipfs-api) and _Embedded_ (js-ipfs) node types in the main UI:

- Turns out users rarely change this setting (and sometimes instinctively assume it is a sort of an on/off switch!)
- Our future integration with Brave will change the meaning of "Embedded" node, current UI does not translate to that well
- Finally, it oversells the state of js-ipfs running is browser context:
  - It is still WIP experiment with known issues in browser context: #450, and we should do a better job in communicating that to the user,
  - After bootstrap nodes and public gateways were separated, the UX of embedded node degraded severely. 
    - It will improve after [circuit-relaying](https://github.com/ipfs/js-ipfs/tree/master/examples/circuit-relaying) with automatic relay discovery is shipped, but will take time.  Until then we should adjust main UI and de-emphasize embedded node  it in favor of using external one.

### Changes

This PR delivers an initial part of #491:

- Moves status fields back to the space-header, resulting in two clear sections instead of three vague ones
- Replaces node type toggle with a global On/Off switch that enables/disables gateway redirect and all active API integrations.
  - Perfect for users who don't want to run js-ipfs all the time to save battery, want to disable redirect  or temporarily block access to API `window.ipfs`
  - Things that don't require access to API will still work in OFF state:
    linkify, copy IPFS Path, custom protocol handlers
    (they just redirect to public gw)
- Moves switch between _External_ (js-ipfs-api) and _Embedded_ (js-ipfs)  to _Preferences_ screen
- Hides unused options when embedded node is active (_Preferences_ screen)
- Changes fonts to ones from ipfs-css

### Open Questions

- Does the updated browser action popup look good on Mac? 
  (We had issues with pop-up in past, would be useful to review and confirm by a Mac user)


olizilla:
@lidel this is rad. 

How about instead of removing items that won't work currently we gray them out with a title attr like "Please enable IPFS Companion". I think it'll feel more stable if the menu remains the same height, but items fade in and out of a based on the state. That might reduce any issue caused by the resize events during initial render too.

I kinda like the stats in the top box, but now the menu footer feels a little "unconnected" I don't quite know the right words to explain that feeling, but with the stats at the bottom the box feels a little more solid / grounded. Maybe I'm wrong, what do you think?

lidel:
@olizilla  I think keeping and fading-out status items is a good idea,  just added a commit that does just that:



We still have to resize popup due to context-dependent menu items, but that is below the static header, so should be ok for now.

I want to keep this PR small and release to Beta for further feedback, as many people were interested in "OFF switch".

Things for separate PRs: 

- Perhaps we could reorder items so that it is less janky?

- Lack of footer does not really bother me, but perhaps we could restore balance by moving "Share via IPFS to the very bottom and styling it differently (darker + icon) ?

olizilla:
Yep, that is better. How about including the menu options, I think it'll look calmer if the box doesn't resize.

olizilla:
On Sun, 17 Jun 2018, 09:24 Alan Shaw, <notifications@github.com> wrote:

> Hey, I think it’s great to have an enable/disable for companion and given
> the requests for the feature there’s definitely an argument for having it
> accessible from the drop down and not just in the options page. I have a
> few concerns I just wanted to raise:
>
>    - naming - I’m not sure if there’s a good answer but could users
>    confuse on/off with actually starting/stopping the ipfs node?
>
>
This is a quick way for a user to disable all of the things companion does,
without forcing them to disable it in the extention settings, where they
are more likely to forget to turn it back on, as it's that removes it from
the toolbar buttons (in chrome at least)



>    - behaviour - if running an embedded node what happens - is it
>    actually stopped? Should the behaviour be more similar for both node types
>    i.e. don’t actually stop an embedded node?
>
>
It stops what companion is doing. If it's an embedded node, it stops that
too.


>    - importance - is on/off our number 1 feature? It certainly looks that
>    way atm. Perhaps a smaller toggle in the top right/left might be better? We
>    don’t *really* want people to disable companion, right(?), but it
>    sounds as though it might be necessary
>
>
It gives new users a quick way to see the difference that companion makes
over regular we browsing, so it gets top billing. For existing users it's
still a feature you want quick access to.



>    - sanity double check - is disabling the add-on in the browser
>    preferences effectively the same thing?
>
>
Similar, but it keeps the extension in view.


>    - embedded node relegation - I understand the argument that js-ipfs
>    has issues atm wrt stability/reliability but relegating this feature to
>    (only) the options screen is going to hugely impact it’s exposure to users
>    which in turn won’t help us make it more stable/reliable. Would adding
>    warning messages to clearly state that it’s experimental suffice?
>
>
The next part of this work is to add a button right after the on/off switch
that let's the use select which ipfs backend they want to use. This'll give
is space to give the right messaging around js-ipfs

olizilla:
By which I mean, they are all totally legit concerns, but I think it's
gonna be ok. Right now I'm finding the enable/disable toggle in the
extension settings page to be my most used feature, so it'll help me at
least.

There is more detail on "where are we going with this" in
https://github.com/ipfs-shipyard/ipfs-companion/issues/491

We only want to relegate js-ipfs a little bit, and put the option on it's
own page where we can let the user add extra ones and add one for the new
built in go-ipfs node in Brave, so the node toggle was gonna get a little
overloaded.

On 17 Jun 2018 09:48, "Oli Evans" <oli@zilla.org.uk> wrote:



On Sun, 17 Jun 2018, 09:24 Alan Shaw, <notifications@github.com> wrote:

> Hey, I think it’s great to have an enable/disable for companion and given
> the requests for the feature there’s definitely an argument for having it
> accessible from the drop down and not just in the options page. I have a
> few concerns I just wanted to raise:
>
>    - naming - I’m not sure if there’s a good answer but could users
>    confuse on/off with actually starting/stopping the ipfs node?
>
>
This is a quick way for a user to disable all of the things companion does,
without forcing them to disable it in the extention settings, where they
are more likely to forget to turn it back on, as it's that removes it from
the toolbar buttons (in chrome at least)



>    - behaviour - if running an embedded node what happens - is it
>    actually stopped? Should the behaviour be more similar for both node types
>    i.e. don’t actually stop an embedded node?
>
>
It stops what companion is doing. If it's an embedded node, it stops that
too.


>    - importance - is on/off our number 1 feature? It certainly looks that
>    way atm. Perhaps a smaller toggle in the top right/left might be better? We
>    don’t *really* want people to disable companion, right(?), but it
>    sounds as though it might be necessary
>
>
It gives new users a quick way to see the difference that companion makes
over regular we browsing, so it gets top billing. For existing users it's
still a feature you want quick access to.



>    - sanity double check - is disabling the add-on in the browser
>    preferences effectively the same thing?
>
>
Similar, but it keeps the extension in view.


>    - embedded node relegation - I understand the argument that js-ipfs
>    has issues atm wrt stability/reliability but relegating this feature to
>    (only) the options screen is going to hugely impact it’s exposure to users
>    which in turn won’t help us make it more stable/reliable. Would adding
>    warning messages to clearly state that it’s experimental suffice?
>
>
The next part of this work is to add a button right after the on/off switch
that let's the use select which ipfs backend they want to use. This'll give
is space to give the right messaging around js-ipfs

lidel:
@olizilla just pushed version that does not hide inactive menu items, but hides context-items in non-ipfs contexts:

#### Non-IPFS Context

> peek 2018-06-18 12-34

#### IPFS Context

> peek 2018-06-18 12-39

lidel:
As a temporary measure, would it be better to move the switch to the bottom or the top?


### Bottom
screenshot_5 

screenshot_38

alanshaw:
> I like top right corner better

+1

olizilla:
Top right looks neat, but I think that's partly because the labels have been removed. I feel it should have a label. I'd either leave it in the middle, or go top right an pull the label into the switch http://www.uiparade.com/portfolio/toggle-switch-2/

lidel:
@olizilla  I really want to avoid label, as it may bring too much focus to the toggle. 
Personally, I would not stress over label too much: even if user does not intuit the meaning behind toggle in browser action, "_Open Preferences_" is the only active menu item, and the toggle is there as well, but with a clear label and color hint:

> screenshot_9



----


If material switch is too ambiguous, what if we replace it with [ipfs-css/icons/glyph_power.svg](https://github.com/ipfs-shipyard/ipfs-css/blob/master/icons/glyph_power.svg)? 

Quick mockup (glyph version, but stroke one may be a better choice):

> screenshot_40

> screenshot_41

Another option is to move icon back to the middle, but utilize the space better, by adding other quick actions, such as quick access to _Preferences_:

> screenshot_42

olizilla:
I like it! The last one with the power and config icons in the middle is my favourite.

alanshaw:
> The last one with the power and config icons in the middle is my favourite.

+1 yes, do it

lidel:
Done! 
Is there anything to tweak before this can be merged and released to Beta? 

peek 2018-06-20 11-10

alanshaw:
🔥 HAWT

Just to check, does the cog open preferences? Do we need a menu item _in additon to_ the cog?

olizilla:
When the "power is off" I'd make the power button slightly brighter than the disabled options. Right now it looks like it might also be disabled. Perhaps make the power icon svg be`snow` white when the power is off.

I'd drop the "open preferences" option. Let's make the cog _the_ way to get to the extension settings.

I think we need to do more thinking on the "API is down" state, but that's for another PR.

Generally tho, this is looking great 👍 ✨

lidel:
Thanks! 

When in "power off' state, icons look more actionable now:
> screenshot_12

As suggested, dropped duplicate "Open Preferences" from the menu to decrease clutter.
Other popular extensions (such as uBlock) use small icons to save screen space, so it should not be a problem:
> screenshot_13
> screenshot_15




I've been thinking about using yellow/red accent color to indicate API being down when in "power on" state, but for now let's go with simple fade-out.

lidel:
Merged and released to Beta in [v2.3.1.9890](https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.3.1.9890) :rocket:

PR #501, opened 2018-06-21
feat: meaningful error in redirect-based URI handler

lidel:
## Context

This PR supersedes https://github.com/ipfs-shipyard/ipfs-companion/pull/422 and adds a useful error page for rare cases when redirect-based URI handler was unable to process URI.

Note that redirect-based handler is specific to Firefox  as Chrome does not support `manifest.json/protocol_handlers`. 

In my experience this specific type of error is displayed only when someone is playing with protocol handler and enters address by hand to the address bar and makes a typo or puts `foobar` as a CID.

## Before

> screenshot_2

Opening `ipfs://invalid-cid` redirected  to https://ipfs.io/#redirect/ipfs%3A%2F%2Finvalid-cid, but because `invalid-cid` is not a valid IPFS resource, it was not redirected further. 
The result was very confusing. Invalid CID just displayed our main page (seen above)

## Now

> screenshot_14

Opening `ipfs://invalid-cid` redirects  to https://gateway.ipfs.io/ipfs/QmXQY7mKr28B964Uj4ouq3fPgkNLqzaKiajTA7surAiQuD#ipfs%3A%2F%2Finvalid-cid which:
- displays meaningful error message and some helpful steps
- is a regular IPFS resource, so it does not require HTTP-based hosting and will be redirected to local gateway if IPFS Companion is active
  - Note: **we can't bundle this error page with browser extension**, because [`webRequest/onBeforeRequest`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/onBeforeRequest) with `<all_urls>`  ignores requests made to `moz-extension://*` (this seems to be a security feature to block extensions from interacting with each others pages)
- `gateway.ipfs.io` looks better in [protocol handler picker](https://user-images.githubusercontent.com/157609/41741285-eed46276-759a-11e8-8221-16173704f733.png)

## Next

I feel we should not stress too much about this error page due to its rare use. 
Design can be improved in a separate PR, but I would wait with that until our visual language matures.

### DIFF
--- a/add-on/manifest.firefox.json
+++ b/add-on/manifest.firefox.json
@@ -23,32 +23,32 @@
     {
       "protocol": "web+dweb",
       "name": "IPFS Companion: DWEB Protocol Handler",
-      "uriTemplate": "https://ipfs.io/#redirect/%s"
+      "uriTemplate": "https://gateway.ipfs.io/ipfs/QmXQY7mKr28B964Uj4ouq3fPgkNLqzaKiajTA7surAiQuD#%s"
     },
     {
       "protocol": "web+ipns",
       "name": "IPFS Companion: IPNS Protocol Handler",
-      "uriTemplate": "https://ipfs.io/#redirect/%s"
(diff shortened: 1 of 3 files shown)


olizilla:
Nice!

Issue #16, opened 2015-04-11
Support website hints to reroute HTTP query to IPFS


lidel:
#### Summary of Open Ideas
- [x] Detect host names with dnslink and redirect requests to IPFS (`/ipns/<fqdn>`) 
   -  PoC in  https://github.com/lidel/ipfs-firefox-addon/issues/16#issuecomment-180552903
-  _https://github.com/ipfs/notes/issues/92: Recognize IPFS gateways running on arbitrary websites_
    - [x] idea(1): run some kind of validation for URL that contain `/ipns/` or `/ipfs/` segment (risk of false-positives for IPNS namespace)  
    - [x] idea (2): read `TXT` DNS record and check if it contains `dnslink` (no false-positives)
- [x]  _https://github.com/ipfs/notes/issues/104 A hint that site assets with names being valid IPFS multihashes can be loaded from IPFS_
   - idea: [use _RFC 6249: Metalink/HTTP: Mirrors and Hashes_ as a way to provide hint via HTTP headers](https://github.com/ipfs/notes/issues/104#issuecomment-287631968)

Semi-related:
- https://github.com/ipfs/in-web-browsers/issues/55: IPFS runs as a Service Worker!

Fil:
What about URLs like `(whatever)/resource?hash::HASH`

browsers (or proxies) that don't know about `?hash::` would do the usual thing, others would be able to use ipfs, or download and check the signature, or even "seed" the resource — which means that a webmaster could get all the benefits of ipfs for their resources w/o installing ipfs themselves, just but by signing them

lidel:
Yes, that also sounds good! 
To make adoption easy, it should be self-explanatory and minimalistic, and `?ipfs:<path>` would be a perfect alternative for people who can't use `/.well-known/ipfs` (shared hosting etc).

Right now I plan to support (redirect request to a custom gateway, if enabled):
- `/.well-known/ipfs/<path>`
- `/.well-known/ipns/<path>`
- `<url>?ipfs:<path>`
- `<url>?ipns:<path>`

This add-on already supports ([v1.0.1](https://github.com/lidel/ipfs-firefox-addon/releases/tag/v1.0.1)) `ipfs:<path>` and `ipns:<path>` in address bar, so it would follow a single naming convention :sparkles:

Fil:
Wait. Reading https://tools.ietf.org/html/rfc5785#page-3 we need to make sure that we follow the rules

>   well-known URIs are not intended
[...]

PR #505, opened 2018-06-27
Redirect Opt-out via URL Hints

lidel:
**TL;DR**: this PR adds support for `x-ipfs-no-redirect` symbol as a way to disable gateway redirect for a single request, without disabling anything globally.

    https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR#x-ipfs-no-redirect

 ```js
 function isSafeToRedirect (request, runtime) {
   // Do not redirect if URL includes opt-out hint
   if (request.url.includes('x-ipfs-no-redirect')) {
    return false
  }
 ```

That is all. The rest is just some cleanup/refactoring without any functional changes.

### Background 

This PR adds an URL-based opt-out from global gateway redirect that website developers can rely on  to ensure request goes to the specific HTTP gateway.

Potential consumers of this feature:
- Public gateway preloaders (IPFS Companion already does this for uploads)
- Health checks (eg. https://ipfs.github.io/public-gateway-checker)
- Self-hosted clusters of IPFS gateways optimized for specific use case (eg. video streaming)

### Details

- IPFS Path normalization and potential redirect happens in [`webRequest.onBeforeRequest`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/onBeforeRequest).
  - This is where we skip redirect if a hint (`x-ipfs-no-redirect`) is present in URL.
- The opt-out hint can be anywhere, but we expect people to use it:
   - ad-hoc as a part of hash (cache-friendly way): https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR#x-ipfs-no-redirect
   - as one of query args: https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?foo=bar&x-ipfs-no-redirect
- Preloading Change: until now all HTTP HEAD requests were excluded from gateway redirect. This PR changes this behavior: restores consistent handling for all request types and requires explicit redirect opt-out for preload requests to work. 

Thoughts?

cc @lgierth @VictorBjelkholm @kyledrake

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,13 @@ Learn more at [ipfs.io](https://ipfs.io) (it is really cool, we promise!)
 
 #### Automagical Detection of IPFS Resources
 
-  Requests for IPFS-like paths (`/ipfs/$cid` or `/ipns/$peerid_or_fqdn-with-dnslink`) are detected on any website.  
-  If tested path is a valid IPFS address it gets redirected and loaded from a local gateway, e.g:  
-  `https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`  
-  → `http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`
+Requests for IPFS-like paths (`/ipfs/$cid` or `/ipns/$peerid_or_fqdn-with-dnslink`) are detected on any website.  
+If tested path is a valid IPFS address it gets redirected and loaded from a local gateway, e.g:  
+> `https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`  
+> → `http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`
+
+  It is possible to temporarily opt-out from redirect by suspending extension via global toggle,
+  or by including `x-ipfs-no-redirect` in the URL ([as a hash or query parameter](https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?x-ipfs-no-redirect#x-ipfs-no-redirect)).
 
 #### IPFS API as `window.ipfs`
 
--- a/README.md
+++ b/README.md
@@ -43,8 +43,9 @@ If tested path is a valid IPFS address it gets redirected and loaded from a loca
 > `https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`  
 > → `http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`
 
-  It is possible to temporarily opt-out from redirect by suspending extension via global toggle,
-  or by including `x-ipfs-no-redirect` in the URL ([as a hash or query parameter](https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?x-ipfs-no-redirect#x-ipfs-no-redirect)).
+It is possible to opt-out from redirect by
+a) suspending extension via global toggle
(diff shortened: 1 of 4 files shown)


VictorBjelkholm:
Shouldn't this be done as a `header` on the request instead of having to change the URL?

lidel:
@VictorBjelkholm  that was my first thought, but the idea got hit by API limitations quite fast:

- Gateway redirect happens in [`onBeforeRequest`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/onBeforeRequest). This hook does not provide access to headers, but enables us to do redirects.. 

- We can modify headers in [`onBeforeSendHeaders`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/onBeforeSendHeaders), but can't trigger redirect from there. :upside_down_face: 

But now that I think about it, what we can do is to detect header in `onBeforeSendHeaders`, store requestId as a skip flag somewhere and then check if skip flag exists in `onBeforeRequest`. 
Will look into this.

lidel:
@olizilla right now we detect IPFS paths on any website and redirect them. This means browser extension opportunistically upgrade transport to local IPFS node everywhere IPFS paths are used.

The problem this PR solves is that there is no opt-out mechanism for rare cases such as public gateway preloads and bug workarounds.

@VictorBjelkholm  I confirmed that opt-out via header is doable, but it comes at a price: it breaks [CORS preflight request](https://user-images.githubusercontent.com/157609/41974879-7985dd64-7a19-11e8-9fd4-4e8bf1b098a9.png).

There is no easy fix. If user has browser extension installed, we can detect the header and remove it before it is sent to gateway, but we can't do that for users without the extension.

This means opt-out via HTTP Header requires additional setup step from gateway administrators, namely  to include `x-ipfs-no-redirect` in `Access-Control-Allow-Headers`, and we can't assume everyone will have a public gateway under own control. 

I think we should either: 
- support opt-out only via URL
- support opt-out via both URL and Header, but leave it up to gateway admin to set up CORS, if they choose to use Header instead of URL hint

Thoughts?

olizilla:
It feels like we're dropping some useful info if we do... they are trying to tell us which gateway to use. Should we avoid re-writing IPFS urls that explicitly refer to a custom gateway? 

`/ipfs/<hash>` -> redirect to local ipfs gateway
`https://ipfs.io/ipfs/<hash>` -> redirect to local ipfs gateway
`https://dweb.video/ipfs/<hash>` -> leave it alone.

lidel:
Ok, time for some context:
What our browser extension does since 2017 is a global, opportunistic protocol upgrade that is "opt-in by default": it detects `/ipfs/<CID>` on _any_ website and redirects it to local HTTP2IPFS gateway.  This is enabled by _default_ for _every_ webpage,The gateway at  `ipfs.io` is not special: the same rules apply to all public gateways. As long your assets are under a path that starts with valid  `/ipfs/<CID>/*` it will be loaded over IPFS.

What this PR changes is to add optional "opt-out" mechanism: look at URL of request made to any public gateway and if it sees `x-ipfs-no-redirect` in will leave request as-is, won't redirect anything. **That is the only change proposed here:** an explicit opt-out mechanism for website and dapp developers to handle edge cases when they want to be sure request goes to the specific gateway.  


ps. Historical discussion about  opportunistic HTTP2IPFS upgrade on every website can be found in  https://github.com/ipfs-shipyard/ipfs-companion/issues/16. In short, I strongly think it should remain opt-out due to  [the tyranny of the default](https://pagefair.com/blog/2015/the-tyranny-of-the-default/).

lidel:
I think I was just not clear enough in my initial description what this PR does, sorry about that.

The change is really, really small:

**TL;DR**: this PR adds support for `x-ipfs-no-redirect` symbol as a way to disable gateway redirect for a single request ad-hoc, without disabling anything globally.

    https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR#x-ipfs-no-redirect

 ```js
 function isSafeToRedirect (request, runtime) {
   // Do not redirect if URL includes opt-out hint
   if (request.url.includes('x-ipfs-no-redirect')) {
    return false
  }
 ```

That is all. The rest is just some cleanup/refactoring without any functional changes.

olizilla:
I don't feel totally comfortable about supporting this. If it's just for companion, perhaps the token to check for should be `x-ipfs-companion-no-redirect`.

PR #511, opened 2018-07-02
Restore gateway redirect for CORS XHRs in Firefox

### LINKED ISSUE #436: gateway redirect messes up CORS headers on firefox

lidel:
This PR restores gateway redirect for CORS XHR in Firefox via late redirect in `onHeadersReceived` and closes #436.

Context for why CORS XHR were not redirected until now can be found in https://github.com/ipfs-shipyard/ipfs-companion/issues/436#issuecomment-378254294 and upstream [Bug #1450965](https://bugzilla.mozilla.org/show_bug.cgi?id=1450965). 

In short, `onBeforeRequest` can't redirect anything when CORS XHR is processed, otherwise it will trigger false-positive CORS error.  Good news is that `onHeadersReceived` is executed  long after CORS validation happens in Firefox, and allows us to cancel original connection and do a late redirect right after response headers arrive. 

This is the best we can do until upstream [Bug #1450965](https://bugzilla.mozilla.org/show_bug.cgi?id=1450965) is addressed.

Additional Notes: 
- The original outgoing request to the public gateway has to happen :( 
  I [raised privacy concerns](https://bugzilla.mozilla.org/show_bug.cgi?id=1450965#c4) in the upstream bug.
-  Good news is that we only need to read response headers before connection is  aborted. The overhead is minimal and I believe it is worth it: enables us to reach feature-parity with Chrome and removes dapp developer confusion that was caused by different redirect behaviors in Firefox and Chrome. 

----
Jean-François Millet, 1857, Oil on canvas
_Developers look for an efficient workaround in WebExtension API_
Jean-François Millet, 1857, Oil on canvas

### DIFF
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -68,7 +68,7 @@ function createRequestModifier (getState, dnsLink, ipfsPathValidator, runtime) {
         if (request.url.includes('/api/v0/add')) {
           for (let header of request.requestHeaders) {
             if (header.name === 'Connection') {
-              console.warn('[ipfs-companion] Executing "Connection: close" workaround for go-ipfs/issues/5168')
+              console.warn('[ipfs-companion] Executing "Connection: close" workaround for ipfs.files.add due to https://github.com/ipfs/go-ipfs/issues/5168')
               header.value = 'close'
               break
             }
@@ -176,7 +176,7 @@ function isSafeToRedirect (request, runtime) {
       const sourceOrigin = new URL(request.originUrl).origin
       const targetOrigin = new URL(request.url).origin
       if (sourceOrigin !== targetOrigin) {
-        console.warn('[ipfs-companion] Delaying redirect of CORS XHR until onHeadersReceived due to ipfs-companion/issues/436:', request.url)
+        console.warn('[ipfs-companion] Delaying redirect of CORS XHR until onHeadersReceived due to https://github.com/ipfs-shipyard/ipfs-companion/issues/436 :', request.url)
         onHeadersReceivedRedirect.add(request.requestId)
         return false
       }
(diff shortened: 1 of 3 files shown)


olizilla:
💯 bonus points to @lidel for the issue report on bugzilla with minimal working extension to demonstrate the issue, and art lolz in the PR.

add-on/src/lib/ipfs-request.js
```diff
@@ -65,7 +68,7 @@ function createRequestModifier (getState, dnsLink, ipfsPathValidator, runtime) {
         if (request.url.includes('/api/v0/add')) {
           for (let header of request.requestHeaders) {
             if (header.name === 'Connection') {
-              console.log('[ipfs-companion] Executing "Connection: close" workaround for https://github.com/ipfs/go-ipfs/issues/5168')
+              console.warn('[ipfs-companion] Executing "Connection: close" workaround for go-ipfs/issues/5168')
```
olizilla: nitpick: errors with urls in are great, let's keep it

PR #516, opened 2018-07-05
Fix XHR redirects when dnslink is enabled

lidel:
Dnslink lookup is disabled by default, but if user opted-in in latest version websites were broken due to false-positive late gateway redirects of XHR requests.

This PR  fixes dnslink regression introduced by https://github.com/ipfs-shipyard/ipfs-companion/pull/511
and adds tests to guard against the problem in future.

Some deeper refactoring is due, but that is something for separate PR.
I want to land this PR as soon as possible, so that dnslink works again.


etam:
Yep. Seems to be working fine. Thanks!

PR #520, opened 2018-07-10
feat: add take screenshot button

alanshaw:


lidel:
We could ship this as-is, but don't want to take space in menu item.  Maybe put it behind and icon or keyboard shortcut instead?

@hacdias what are plans for screenshots in Desktop? 
Will it be only shortcut-based or there will be an icon/menu item?

hacdias:
@lidel there is a shortcut on IPFS Desktop (CTRL/CMD+ALT+S). It takes a screenshot, adds it to screenshots/ dir and copies the link to the clipboard. It's already available and we'll keep it on the revamp branch 😄

jessicaschilling:
Is this a stale PR that should be closed?

lidel:
Ah, yes, we did this as a quick hack project, but turns out WebExtension APIs for browser viewport screenshots are super limited  and there are better tools for this built-in browsers now with better functionality:

https://support.mozilla.org/en-US/kb/firefox-screenshots

PR #523, opened 2018-07-12
feat: upgrade to js-ipfs 0.30

alanshaw:


add-on/src/lib/dir-view.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 const filesize = require('filesize')
-const mainStyle = require('ipfs/src/http/gateway/dir-view/style')
+const mainStyle = require('ipfs-http-response/src/dir-view/style')
```
lidel: Does this mean we can remove `ipfs/src/http/gateway/dir-view/` from companion repo?

package.json
```diff
@@ -87,9 +87,10 @@
     "file-type": "8.0.0",
     "filereader-pull-stream": "1.0.0",
     "filesize": "3.6.1",
-    "ipfs": "0.29.3",
-    "ipfs-api": "22.2.1",
+    "ipfs": "0.30.0",
+    "ipfs-api": "22.2.3",
     "ipfs-css": "0.5.2",
+    "ipfs-http-response": "^0.1.2",
```
lidel: `^` :upside_down_face:

add-on/src/lib/dir-view.js
alanshaw: No, that's a path in the ipfs module

package.json
alanshaw: 🤦‍♂️ every time

Issue #508, opened 2018-06-29
Get current CID for and IPNS address


olizilla:
I found myself trying to ask companion what the CID of an IPNS address was. URL bar page action option?

PR #532, opened 2018-07-16
feat: copy current IPFS path for IPNS address

lidel:
This PR implements context menu item to get  current CID for an IPNS address: 

screenshot_6

Closes https://github.com/ipfs-shipyard/ipfs-companion/issues/508

I am quite sure this needs to be reworded before merging. 

@olizilla  Any ideas how to label this feature?

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -71,6 +71,10 @@
     "message": "Copy Public Gateway URL",
     "description": "A menu item in Browser Action pop-up and right-click context menu (panel_copyCurrentPublicGwUrl)"
   },
+  "panel_copyResolveIpns": {
+    "message": "Copy IPNS Resolved to IPFS",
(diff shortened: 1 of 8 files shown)


olizilla:
This might be more intuitive if we change it to show the user the information rather than trying to frame it as a set of questions... like:

**IPFS Resource loaded via Custom Gateway**

<sup>URL</sup>
`http://ipfs.io/ipns/room.guardian`

<sup>PATH</sup>
`/ipns/room.guardian`

<sup>CID</sup>
`bafybafybafybase32yeahwoo`

---
- Copy URL
- Pin CID to local repo

olizilla:
I think URL in that menu should always show the canonical http url `https://ipfs.io/ip*`

We could use it to show other address forms too as they become supported... hello `ipfs://*`

lidel:
Let's continue brainstorming on this in https://github.com/ipfs-shipyard/ipfs-companion/issues/587

PR #536, opened 2018-07-22
Update gateway URLs to omit gateway. prefix

lgierth:
I noticed occurances of `gateway.ipfs.io` in my everyday usage of companion -- assuming these were leftovers I went and changed them to just `ipfs.io`.

### DIFF
--- a/add-on/manifest.firefox.json
+++ b/add-on/manifest.firefox.json
@@ -23,32 +23,32 @@
     {
       "protocol": "web+dweb",
       "name": "IPFS Companion: DWEB Protocol Handler",
-      "uriTemplate": "https://gateway.ipfs.io/ipfs/QmXQY7mKr28B964Uj4ouq3fPgkNLqzaKiajTA7surAiQuD#%s"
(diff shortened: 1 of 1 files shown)


ivan386:
Do not use ipfs.io instead of gateway.ipfs.io. It have dnslink record. This a problem when redirect this address to local gateway.

lidel:
@ivan386 the problem  you describe (dnslink clashing with gateway feature on same host) should  be fixed in the [latest beta](https://github.com/ipfs-shipyard/ipfs-companion/releases), see PR https://github.com/ipfs-shipyard/ipfs-companion/pull/518 for details.
It will land to stable soon.

lgierth:
> I kinda left that on purpose there upside_down_face, so user of redirect-based handler in Firefox gets an explicit hint that a HTTP gateway will be used, not the native protocol [...] I'd left it as-is until we move to dweb.link and subdomains

Understood, makes perfect sense :)

lidel:
Alright, we will change this after move to `dweb.link` :)

PR #537, opened 2018-07-23
feat: support page actions for CID subdomains

### LINKED ISSUE #526: Detect CID in subdomains
It is feasible to use `cidv1b32` as a subdomain, to take advantage of origin-based security perimeter in web browsers.
It should be fairly easy to detect if subdomain is a valid CID, and if so, display the same UI elements as for `/ipfs/<cid>` paths.

lidel:
Closes #526 

Test URLs: http://bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy.ipfs.dweb.link

### DIFF
--- a/add-on/src/lib/ipfs-path.js
+++ b/add-on/src/lib/ipfs-path.js
@@ -4,8 +4,19 @@
 const IsIpfs = require('is-ipfs')
 
 function safeIpfsPath (urlOrPath) {
+  if (IsIpfs.subdomain(urlOrPath)) {
+    urlOrPath = subdomainToIpfsPath(urlOrPath)
+  }
   // better safe than sorry: https://github.com/ipfs/ipfs-companion/issues/303
   return decodeURIComponent(urlOrPath.replace(/^.*(\/ip(f|n)s\/.+)$/, '$1'))
+}
+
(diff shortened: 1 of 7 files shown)


add-on/src/lib/ipfs-path.js
```diff
@@ -4,10 +4,21 @@
...
+  }
   // better safe than sorry: https://github.com/ipfs/ipfs-companion/issues/303
   return decodeURIComponent(urlOrPath.replace(/^.*(\/ip(f|n)s\/.+)$/, '$1'))
 }
 
+function subdomainToIpfsPath (urlString) {
+  const url = new URL(urlString)
+  const fqdn = url.hostname.split('.')
+  const cid = fqdn[0]
+  const protocol = fqdn[1]
```
lidel: Just realized this won't produce correct results for gateways under multi-leveled domains such as `cid.ipfs.something.co.uk`.
Easy fix is to switch to `is-ipfs`, probably something like:

```js
const { subdomainPattern } = require('is-ipfs')

const match = fqdn.match(subdomainPattern)
if (!match) throw new Error('no match for isIPFS.subdomainPattern')
const protocol = match[1]
const cid = match[2]
```

PR #552, opened 2018-08-06
fix: drag & drop uploads

### LINKED ISSUE #549: Drag n' Drop to add files to IPFS not working
firefox 61.0.1 on fedora 28 workstation
ipfs-companion Beta @ d46794a

Trying to drop a file just opens it in the browser.

lidel:
This PR restores drag&drop support and closes #549.
(Looks like it got removed during one of refactorings)

- directory upload is not possible at this time, we detect it and display error message instead
- empty file without extension looks the same as empty directory, so the error accounts for that possibility
- account for edge cases where the source of dragged object is not valid, which results in drag&drop event with empty file list

# Demo

drag-and-drop-demo

# Interop tests

- [ ] @olizilla  – would appreciate drag&drop smoke-test under Mac
- [x] @hacdias   – would appreciate drag&drop smoke-test under Windows

### DIFF
--- a/add-on/src/popup/quick-upload.js
+++ b/add-on/src/popup/quick-upload.js
@@ -5,6 +5,7 @@ const browser = require('webextension-polyfill')
 const choo = require('choo')
 const html = require('choo/html')
 const logo = require('./logo')
+const drop = require('drag-and-drop-files')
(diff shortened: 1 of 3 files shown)


lidel:
Ooops, forgot about this one :upside_down_face:  lets move it to Beta soon.

PR #553, opened 2018-08-08
[PoC] libdweb discovery and transport

alanshaw:

PR #557, opened 2018-08-26
Improved logger

lidel:
This PR tries to address the problem of noisy console logs in tests  and a chore of adding debug ones during development.

- switched from `console.log` to [debug](https://www.npmjs.com/package/debug) 
- light convention of default, debug and error loggers 
- hardcoded different color for each logger type
- ability to customize debug level of other ipfs packages using `debug` library via `localStorage.debug`
  - can be changed [here](https://github.com/ipfs-shipyard/ipfs-companion/compare/improved-logger?expand=1#diff-170909c5725b3676d266c4f14d69aba5R19) but  we could expose this on the _Preferences_ screen (either as input or checkbox to enable debug logger) to help users and developers in troubleshooting

### Known Issues / Open Questions

- [ ] `debug`  v3.1.0 does not work correctly in webextension context.
It was fixed in https://github.com/visionmedia/debug/pull/476 but not released yet, so this PR uses specific git revision instead of npm release.

This is just an initial stab to see how it could work. 

cc @alanshaw, @olizilla   Do we have already established logging conventions in other JS projects that should be applied here? Personal opinions are also useful  :)

### DIFF
--- a/add-on/src/lib/ipfs-companion.js
+++ b/add-on/src/lib/ipfs-companion.js
@@ -6,6 +6,7 @@ const { optionDefaults, storeMissingOptions } = require('./options')
 const { initState, offlinePeerCount } = require('./state')
 const { createIpfsPathValidator, urlAtPublicGw } = require('./ipfs-path')
 const createDnsLink = require('./dns-link')
+const { pleaseWait } = require('./please-wait')
 const { createRequestModifier, redirectOptOutHint } = require('./ipfs-request')
 const { initIpfsClient, destroyIpfsClient } = require('./ipfs-client')
 const { createIpfsUrlProtocolHandler } = require('./ipfs-protocol')
@@ -14,6 +15,12 @@ const createCopier = require('./copier')
 const createRuntimeChecks = require('./runtime-checks')
 const { createContextMenus, findUrlForContext } = require('./context-menus')
 const createIpfsProxy = require('./ipfs-proxy')
+
+// LOG
+// production: localStorage.debug = 'ipfs-companion:*,-*:debug'
+//        dev: localStorage.debug = 'ipfs-companion:*'
+localStorage.debug = 'ipfs-companion:*,-*:debug'
+const log = require('./log')('background')
 
 // init happens on addon load in background/background.js
 module.exports = async function init () {
@@ -35,7 +42,9 @@ module.exports = async function init () {
   const browserActionPortName = 'browser-action-port'
 
   try {
+    log('init(): %s', pleaseWait())
     const options = await browser.storage.local.get(optionDefaults)
+    log.debug('loaded options %O', options)
     runtime = await createRuntimeChecks(browser)
     state = initState(options)
     notify = createNotifier(getState)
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -97,6 +97,7 @@
   },
   "dependencies": {
     "choo": "6.13.0",
+    "debug": "https://github.com/visionmedia/debug/tarball/207a6a2d53507ec9dd57c94c46cc7d3dd272306d/debug.tar.gz",
```
alanshaw: non version number?

add-on/src/lib/ipfs-companion.js
```diff
@@ -15,6 +16,12 @@ const createRuntimeChecks = require('./runtime-checks')
 const { createContextMenus, findUrlForContext } = require('./context-menus')
 const createIpfsProxy = require('./ipfs-proxy')
 
+// LOG
+// production: localStorage.debug = 'ipfs-companion:*,-*:debug'
+//        dev: localStorage.debug = 'ipfs-companion:*'
+localStorage.debug = 'ipfs-companion:*,-*:debug'
```
alanshaw: Could this come from the manifest? Or some sort of user config/env var?

alanshaw: Btw, in my libdweb branch I had to do this to enable logging: https://github.com/ipfs-shipyard/ipfs-companion/pull/553/files#diff-fc43b78eb6e482a2ad1a6924959d9ec1R4

Does this definitely work in Chrome?

add-on/src/lib/log.js
```diff
@@ -0,0 +1,14 @@
+const root = 'ipfs-companion'
+
+const debug = require('debug')
+
+module.exports = (loggerName) => {
+  const log = debug(`${root}:${loggerName}`)
+  log.debug = debug(`${log.namespace}:debug`)
```
alanshaw: IMHO `log` and `log.error` are enough.

add-on/src/lib/ipfs-companion.js
alanshaw: 🙄 I'm really bad at reading the PR description...

lidel: > Could this come from the manifest? Or some sort of user config/env var? 

I hoped  we could override it via `web-ext run`'s  [`--pref`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/web-ext_command_reference#--pref) (or via env: `$WEB_EXT_PREF`) but this parameter works only for browser's settings from `about:config`.
WebExtension's _Preferences_ are now kept in separate storage thet can't be overridden on the fly without additional code.

As a workaround,  I was thinking about _Preferences_ screen in extension itself: adding a simple checkbox (that adds or removes `-*:debug`) or  an input which enables you to edit `localStorage.debug` (and enable logs from js-ipfs, libp2p etc).   Personally I'd go with input, as it gives more control to developers and power users who want to look under the hood.   As long we keep it in _Experiments_ section it should be fine. 
@alanshaw  what do you think? does it feel useful?

add-on/src/lib/log.js
lidel: Ok, if we just drop `log.debug` and use `log` everywhere,
should we set the default to:
-  hide non-errors by via something like  `localStorage.debug = '*:error'` 
- or print all Companion logs `localStorage.debug = 'ipfs-companion:*'` 

?

package.json
lidel: Yeah, fix for webextension context (https://github.com/visionmedia/debug/pull/476) was not released yet.

lidel:
Update: debug v4.x works correctly in webextension context!
Closing this PR as it got superseded by https://github.com/ipfs-shipyard/ipfs-companion/pull/712

PR #562, opened 2018-08-31
Workaround for 'invalid Read on closed Body' in files.add (webext in Chrome)

### LINKED ISSUE #480: Unable to upload to IPFS API: passing multiple files to ipfs.files.add sometimes fails
## Background

I was playing with `ipfs.files.api`  over `js-ipfs-api` and `go-ipfs` (PR #479) and noticed an odd edge case. Sometimes, for specific payloads, go-ipfs returns an error in the middle of payload:

```js
{"Name":"T-RexFaceEmoji.jpg_original","Hash":"QmR1G8Jq4rNiYwCKyNWEZhNS7i8EWrCpz2xDjKsB9ofi82","Size":"23723"}
{"Name":"FlyingSaucerEmoji.jpg_original","Hash":"QmVWxHcLVHYpFJP6R1ZLBZcpAZsxyFMfU7BU84D2LAMSed","Size":"13071"}
{"Message":"http: invalid Read on closed Body","Code":0,"Type":"error"}
```

Then, upon receiving this response, `js-ipfs-api` does not throw an error, but silently ignores the last line and returns a partial result list with only two first items.

## How to reproduce

I was able to reproduce it with brand new (dockerized) node and latest Companion from #479, so I assume the bug is real, but would be great if someone else was able to reproduce and comment if they got the same result:

1. Download three files from:
https://ipfs.io/ipfs/QmUGmyGCJEntjyxgcBmptkZ34k7PsgnCzXPoqwiWBDTK6L and save them on your disk
2. Start local instance of go-ipfs v0.4.15. 
   (I was able to reproduce the problem with ephemeral dockerized instance run via `docker run --rm -it  -p 8080:8080 -p 4001:4001 -p 127.0.0.1:5001:5001 ipfs/go-ipfs:v0.4.15`)
3. Try to upload the three files from step 1 (all at once, select them with <kbd>Ctrl</kbd>) via IPFS Companion's Quick Upload screen, you should get the error below:
screenshot_57

## How to fix

Not sure yet, but there are two bugs at play: 
1. go-ipfs returned `http: invalid Read on closed Body` for some reason
    - Upstream bug: https://github.com/ipfs/go-ipfs/issues/5168
      - Temporary workaround (requires custom build of go-ipfs): https://github.com/ipfs/go-ipfs/commit/aad8f813ee9042b8e2e994178923495ab108627e
2. ipfs-api ignored the error and returned a partial response (2 instead of 4 hashes)  as a valid result
    - Upstream bug: https://github.com/ipfs/js-ipfs-api/issues/797

For now, Companion checks if result list has a proper count and displays an error if things are fishy, but it should be addressed upstream.  

I am parking it for now and will  pick up in spare time. 
Until then, if anyone has any quick ideas, feel free to update this issue with any new findings.

lidel:
> Chrome does allow setting `Expect` via [`onBeforeSendHeaders`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/onBeforeSendHeaders) WebExtension API

This means we now have a workaround for https://github.com/ipfs-shipyard/ipfs-companion/issues/480 that works in both Firefox and Chromium.

Details: https://github.com/ipfs-shipyard/ipfs-companion/issues/480#issuecomment-417657758 & https://github.com/ipfs/go-ipfs/issues/5168#issuecomment-417112652

Closes #480

### DIFF
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -82,19 +82,43 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
         // There is a bug in go-ipfs related to keep-alive connections
         // that results in partial response for ipfs.files.add
         // mangled by error "http: invalid Read on closed Body"
-        // More info: https://github.com/ipfs/go-ipfs/issues/5168
-        if (request.url.includes('/api/v0/add')) {
+        // More info (ipfs-companion): https://github.com/ipfs-shipyard/ipfs-companion/issues/480
+        // More info (go-ipfs): https://github.com/ipfs/go-ipfs/issues/5168
+        if (request.url.includes('/api/v0/add') && request.url.includes('stream-channels=true')) {
+          let addExpectHeader = true
+          const expectHeader = { name: 'Expect', value: '100-continue' }
+          const warningMsg = '[ipfs-companion] Executing "Expect: 100-continue" workaround for ipfs.files.add due to https://github.com/ipfs/go-ipfs/issues/5168'
           for (let header of request.requestHeaders) {
-            if (header.name === 'Connection') {
+            // Workaround A: https://github.com/ipfs/go-ipfs/issues/5168#issuecomment-401417420
+            // (works in Firefox, but Chromium does not expose Connection header)
+            /* (disabled so we use the workaround B in all browsers)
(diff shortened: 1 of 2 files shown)


lidel:
Question: should we apply the  workaround only for API used by Companion at a time, or all IPFS APIs on the internet?

Issue #564, opened 2018-09-04
Landing pages flow & content


fsdiogo:
Hey @ipfs-shipyard/gui!

As we talked in the [GUI & In Web Browsers Sync](https://github.com/ipfs/in-web-browsers/issues/105), we need to define the flow and content of the landing pages for Companion.

## Welcome Page

After the extension is installed, the user would be presented with a page that showed something like a thank you message for installing Companion, and also checked if a node is already running, giving simple instructions how to run one if not. We could have some links about the dweb and some IPFS resources.

I think a split screen layout would work great for this:
<img width="1392" alt="welcome" src="https://user-images.githubusercontent.com/33324750/45045713-bfc2bd00-b06b-11e8-8ae2-833736f66dc3.png">

In the left side we would have the Companion related thank you message and node checks, in the right side we would have the links to resources or other apps.

The content is key and it will help with the design, so I'm asking you for some feedback about what should be present here.

## References
- Issue https://github.com/ipfs-shipyard/ipfs-companion/issues/324
- PR https://github.com/ipfs-shipyard/ipfs-companion/pull/326

lidel:
### Some Ideas for **left pane**

- Key feature: provide immediate feedback about connectivity status
   - If online, make it more engaging by displaying  "Connected to N Peers" and the heartbeat animation 
   - If offline, display two options: "Install go-ipfs" (in future we will replace this with "Install IPFS Desktop") or "Start Embedded JS-IPFS Node". 
     - We may even automate it further so it automatically switches to embedded node if HTTP API is down
   - IPFS API (after init) can be accessed like this (we can expose other methods/plumbing this way as well):
     ```js
     const ipfs = (await browser.runtime.getBackgroundPage()).ipfsCompanion.ipfs
     ```
- I wonder if we could re-use animation from https://ipfs.io :thinking: Thoughts?

### Unsure where to put

- (Highlight) a button/link to "Share files via IPFS"
   - It is a good tool to play with after install, if we make it smooth experience it will get user hooked 
- Other "killer apps": IPFS Files from new WebUI (IPFS Drive :P), PeerPad etc

### Some ideas for content for the **right pane**

- [One good onboarding video, but for now it could be a talk about IPFS project]
-  New to IPFS? Read [IPFS Documentation](http://docs.ipfs.io) to learn basic [concepts](https://docs.ipfs.io/guides/concepts/) and working with IPFS.
[...]

PR #565, opened 2018-09-04
feat: add welcome page

fsdiogo:
Check https://github.com/ipfs-shipyard/ipfs-companion/issues/564 for reference.


lidel:
@fsdiogo I've added a commit with  `browser.runtime.onInstalled` hook that opens the landing page on first install or when loaded as a temporary extension. It should make development easier :)
Note: I had to register this hook synchronously before anything else. Registering from async function  missed the  event.

fsdiogo:
Cool, thanks @lidel 🙂

fsdiogo:
Current status:

page

I have to put thumbnails and style them in the videos section, and link them to YouTube. I tried embedding the videos but got an error due to the extension blocking something, I don't remember what exactly, have to check that again.

After that I'll start working on the left side. Thoughts?

lidel:
### Onboarding links

It is very easy to digest, does not overwhelm user, I like it a lot!

Let's reorder onboarding links/sections  to follow the growth of user from noobie  to contributor :)

1. New to IPFS? 
1. Got questions?
1. Want to help?

### Videos

Youtube player was probably blocked by the default CSP, which is _very_ strict in WebExtensions in Firefox:
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_Security_Policy

It is possible to set more liberal CSP, but if you just embedd a thumbnail that opens a video in a new tab it should also be fine:

```html
<a href="https://www.youtube.com/watch?feature=player_embedded&v=YOUTUBE_VIDEO_ID_HERE
" target="_blank"><img src="https://img.youtube.com/vi/YOUTUBE_VIDEO_ID_HERE/0.jpg" 
alt="IMAGE ALT TEXT HERE" width="240" height="180" border="0" /></a>
```

### Left side

Here's an idea: What if when  HTTP API on default port is down (`peerCount`  === -1), we  let user _choose_ if they want to start embedded js-ipfs  or simply connect to a non-standard port/address?

@olizilla  created a very nice UI for this in WebUI, see screenshots in https://github.com/ipfs-shipyard/ipfs-webui/pull/715, we could basically copy the idea and add third option "start embedded js-ipfs".

Thoughts?

olizilla:
> I'm afraid that the instructions will take a lot of space, I'd like this page to be as simple as possible 🤔

That's a reasonable fear. This section of companion has 3 jobs:

- **catch the user after install**, to create a feeling of a coherent process from the promotion page through the install and then making it out the other side to a page that looks similar / familiar and welcoming.

- **ensure the users set up is working** and clearly guide them to install an ipfs daemon (most likely error) or update their api adress. Updating address is really for users that already know something about IPFS, so it can be given a lower visual priority. The page should make it **easy** to restart or go install a deamon, and **possible** to update your api address.

- **inspire and inform the user about what's possible** now they have a working set up.

Where the user has no api running we must still "catch" them, but the most important thing is getting them to a working set up. Given that, it feels like we could have a layout that gives a little space to creating the "catch" and the rest to "How to get this working well"

Where the user has a working set up, catching and inspiring can take up all the space, and we can drop the the "get it working" section. Power users who want to change the api they connect to can do so in the app settings.

When Desktop is updated, we'd message it like "go install desktop, then install companion" to increase the chances of the user being able to use companion right away, but we have to deal with both scenarios.

olizilla:
also of note, where you don't have working setup, we'd still pass through the "everything is working" flow once they fix the issue... the flow for a new user with no daemon is something like:

"the catch" -> "how to get it working" --> _...user goes off and installs a thing..._ -> "hurray you did it!" + "inspire them / what's next".

hacdias:
I like the page, but I was wondering if there aren't any more up to date videos to put there. That 'IPFS Alpha' videos seems a bit old. Or show some of the other apps/functionalities like Desktop or WebUI.

lidel:
I think we could grab something from [IPFS Talks Playlist](https://www.youtube.com/playlist?list=PLuhRWgmPaHtSgYLCqDwnhsQV6RxKDrkqb).

Eg. a short one hands-on (~10min) and a longer with the bigger picture.
For a longer one perhaps [this one from TEDxSanFrancisco](https://youtu.be/2RCwZDRwk48?list=PLuhRWgmPaHtSgYLCqDwnhsQV6RxKDrkqb)?
Is there a better recording of hands-on than the 'IPFS Alpha'?

fsdiogo:
@hacdias @lidel as discussed in today's sync, we'll keep the Alpha Demo for now.

fsdiogo:
## Everything cool:
ok

## Missing daemon:
not-ok

Thoughts? 🙂

fsdiogo:
@lidel new update based on your feedback, thanks!

- Added fonts
	- Used `montserrat` for the Companion title
	- Tried it in the titles but I think it looks better as is
- Added heartbeat animation and disabled logo to reflect the API state
- Fixed weird margins
	- The margin between the `You're all set!` and the bottom copy is less than the logo margin to differentiate those two parts (the extension title and the success state)
- Reordered resources and projects
- Fixed videos margin


## IPFS API 👍
ok

## IPFS API 👎 
nok

lidel:
@fsdiogo looks awesome!  :+1: :1st_place_medal: 

I've added some resources to onboarding steps in right pane and made tiny UI changes in https://github.com/ipfs-shipyard/ipfs-companion/pull/565/commits/dca79328ac2b92732a74e03e5dfb7a7b3d9a5574 (mostly to make it look better with fonts on Linux and for people with custom GTK/dark themes), please fetch it and check if it looks good on Mac. 

The only remaining thing before we merge this is adding `i18n`. 
Just move strings from page to `add-on/_locales/en/messages.json`,  keys probably should start with `page_landingWelcome_` or something like that.

fsdiogo:
Thanks @lidel 😄🎉

I don't know how it looks on Linux, but in macOS I think the titles looked better as they were previously.

With the labels (titles) in bold and uppercase I think it makes it a bit heavy, doesn't match the other labels (videos/projects) and takes attention away from the links:

bold

Usually in cases like this the different color is enough to capture the attention and it looks better in my opinion:

normal

If you really want the bold and uppercase we can lower the font size and see how it looks!

Alright, I'll add i18n to finish this 💪

lidel:
@fsdiogo  I agree:  it looked better before :upside_down_face: 

I think I have over-customized font rendering on my system and that is why I get different results :(
Please remove `ttu tracked` (it should restore previous version).

As for i18n, I  initially suggested  default lib from webextensions may not be enough, and to use `i18next` with ICU format and to try do the same thing Oli did in [ipld-explorer-components/pull/1](https://github.com/ipfs-shipyard/ipld-explorer-components/pull/1/files), but now  I see it relies on a helper specific to react, and I was unable to find something like that for choo.

However, **I think we can make it work with i18n from webextensions**, no need for i18next.

All we need is a very crude render-middleware function: 
```js
renderI18NLink('this <0>simple</0> notation for <1>multiple</1>links', [url1, url2])
```

that renders:

```html
this <a href="url1">simple</a> notation for <a href="url2>multiple</2>links
```

Would this work?

olizilla:
@fsdiogo this is great!

We should direct people to the new docs site in the "with these instructions" link. I think it should go here https://docs.ipfs.io/introduction/install/

fsdiogo:
> All we need is a very crude render-middleware function:
> ...
> Would this work?

Cool, that should work @lidel! I'll get on with it then 😄

Thanks @olizilla, I'll update the link 👍

add-on/src/landing-pages/welcome/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <title>IPFS Companion - Welcome</title>
```
lidel: Are we able to add JS  that sets title to value from i18n ?

add-on/src/utils/i18n.js
```diff
@@ -0,0 +1,62 @@
...
+const browser = require('webextension-polyfill')
+
+/**
+ * Renders a translated string with html anchors.
+ * @param {String} message The message to translate.
+ * @param {Array} links The array of hrefs.
+ * @param {Object} attributes HTML attributes to put in the anchor.
+ * @return {html} An HTML node with the translated string with anchors.
+ */
+const renderTranslatedLink = (message, links, attributes) => {
```
lidel: `*s` (we pass links, plural)

add-on/src/landing-pages/welcome/index.html
fsdiogo: I think we can leave this as is, but in the render of the `page.js` file we can override that to an i18n string! What do you think?

lidel: Sounds good. 
We could also remove static `<title>` and just set it in the render to avoid any visual jitter in title bar on slow machines.

fsdiogo:
Cool, the UI jitter is solved 👍 

The fix for the translated span broke something, I'm getting this:

ss

fsdiogo:
Fixed it, commit incoming.

Other than that I think it's ready to merge 🚀

fsdiogo:
Tested it in `Chrome` and `Firefox` with the latest commits and everything is good 👌

olizilla:
minor, but the vertical alignment of the left side is based on the height of the content on the right side rather than the height of the screen. On a 13" laptop screen the message on the left look a little low down and on a smaller screen the text gets cropped off the bottom. Would be great to fix that so it was centred on screen height, but could be a separate PR.

lidel:
Yeah, I think I see what you mean:

2018-09-17-141457_1269x718_scrot

@fsdiogo will you be able to fix this before merge? :)

fsdiogo:
I'm working on that, but having some issues. When the left side doesn't fit in the viewport I'm unable to have scroll there because I'm using `position: fixed`. Trying to find a solution.

fsdiogo:
Here's the thing: my idea was to fix the left side and give it a `min-height`. When the screen height was less than that `min-height`, the left side would have to scroll, but I only wanted one scroll for the entire page. That isn't possible. I don't know if I'm making myself clear 🤔 

Solution: as we don't have a lot of content on the left side, it's possible to keep it all on the viewport, for screens with smaller widths I can reduce the margins and logo sizes to keep it all in the user view wit media queries.

So the left side will always be fully visible with no scroll, and the right side content will have scroll (the page scroll).

Is that ok with you guys or were you thinking of another solution?

lidel:
> So the left side will always be fully visible with no scroll, and the right side content will have scroll (the page scroll).

I think it would be ok with me, as long responsive reflow on narrow screens (eg. mobile) still works, because it is more important. 

I mean this (right side get moved below left side on narrow viewport):

> 2018-09-18-162344_490x762_scrot

olizilla:
Yep i get you. I'd do what you can to make the content on the left fit the available space on small screens without needing a scrollbar. If the left still overflows vertically, then it should probably get it's own scrollbar like https://jsbin.com/lehemay/1/edit?html,css,output

fsdiogo:
@lidel, yeah, I can make it responsive in mobile as it is now!

@olizilla I thought of that too, but honestly I really don't like the idea of having two individual scrolls. So that would be my last resort.

PR #585, opened 2018-09-18
Improvements for Adding to IPFS via Context Menu

### LINKED ISSUE #579: Bug: Adding images to ipfs sometimes adds the entire page instead
For example, try right clicking the github logo and adding it to IPFS.

lidel:
(Closes #579, https://github.com/ipfs-shipyard/ipfs-companion/issues/592, https://github.com/ipfs-shipyard/ipfs-companion/issues/599)

I played a bit with the concept from #579 and it is not enough to change the order.
Instead we should have clear, separate labels for each context. So far I've added a separate menu item for adding link destination and kept common label for images,videos and audio as "Add This Object to IPFS".

Screenshot below illustrates the most extreme case is when you select part of a page and then click on an image that is also a link (so you get actions for all those contexts):

> screenshot_28

@alanshaw  @olizilla  would appreciate quick review of _"Add ... to IPFS"_ menu  labels, do they sound ok or awkward? How we could improve it to make English locale better?

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -87,16 +87,20 @@
     "message": "Add Selected Text to IPFS",
     "description": "An item in right-click context menu (contextMenu_AddToIpfsSelection)"
   },
-  "contextMenu_AddToIpfsRawCid": {
-    "message": "Add This Object to IPFS",
-    "description": "An item in right-click context menu (contextMenu_AddToIpfsRawCid)"
+  "contextMenu_AddToIpfsImage": {
+    "message": "Add this Image to IPFS",
+    "description": "An item in right-click context menu (contextMenu_AddToIpfsImage)"
   },
-  "contextMenu_AddToIpfsKeepFilename": {
-    "message": "Add This Object to IPFS (Keep Filename)",
-    "description": "An item in right-click context menu (contextMenu_AddToIpfsKeepFilename)"
+  "contextMenu_AddToIpfsVideo": {
+    "message": "Add this Video to IPFS",
+    "description": "An item in right-click context menu (contextMenu_AddToIpfsVideo)"
+  },
+  "contextMenu_AddToIpfsAudio": {
+    "message": "Add this Audio to IPFS",
+    "description": "An item in right-click context menu (contextMenu_AddToIpfsAudio)"
   },
   "contextMenu_AddToIpfsLink": {
-    "message": "Add This Link to IPFS",
+    "message": "Add Link Destination to IPFS",
     "description": "An item in right-click context menu (contextMenu_AddToIpfsLink)"
   },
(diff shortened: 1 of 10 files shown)


alanshaw:
More thoughts:

1. What does "add link" do?
2. Unless there's a good reason to not keep the filename I'd have just a single option for "Add object..." which keeps the filename

lidel:
Ad https://github.com/ipfs-shipyard/ipfs-companion/pull/585#pullrequestreview-156694580) Good ideas. It is possible to provide different label for image, video and audio, so I did just that (see screenshot below).

Ad 1) It executes `addFromURL` using `href` from `<a>` tag that was context-clicked

Ad 2) There are cases when users want raw CID, but I agree it should not be the default. Removed it for now, will think how to re-introduce it.

> screenshot_8

alanshaw:
👍 nice!

How about "Add Linked Content to IPFS" instead of "Add Link Destination to IPFS"?

olizilla:
What do "copy canonical" and "copy public gateway" do in this context? I assume they refer to the page not the selection, but it's not clear.

I think it'd be easier to reason about if it was

- Add selection to IPFS
- Copy page CID _bafy..._
- Copy page IPFS url _https://ipfs.io/..._

lidel:
I've restored option to get "raw CID"  as "Copy Direct CID" (it resolves IPFS/IPNS path to raw CID)
Note: until Page Actions are redesigned in #587  the same action labels are used in popu-menus: 

| Context Menu  | Page Action (Firefox only) | Browser Action | 
| ------------- | ------------- | ------------- |
| screenshot_50 | screenshot_48  | screenshot_49  |

How does this look?

@olizilla  yes, that one was an extreme edge case, usually it is page or image.

Depending on what is clicked, "Copy" actions will look for base URL in this order: 
1. image|video|audio (if clicked on any instead of  page)
2. link (if clicked on any instead of a page)
3. page

Do you think we should have separate labels for each context?

olizilla:
ERH GERD, SORRY I PRESSED THE UPDATE BRANCH BUTTON ACCIDENTALLY. 

revoke my permissions at once.

😿

olizilla:
anyhoo

`ETOOMANYOPTIONS`.

Is it possible to add the entire selection to IPFS regardless of what's in it? Could it always be "Add selection to IPFS" even if it's images and text?

"Direct CID" is tricky, as direct doesn't mean anything in this context. I think its more intuitive if we tell the user what the CID will be for (page, selection).

> I would not put truncated values in menus, it does not tell much and makes them look cluttered.

It might be a terrible idea, as it could make them look cluttered, but I mildly disagree that it wouldn't tell the user much. _Showing_ what they'll get is really helpful if you don't know what a CID is, or what "canonical" means.

PR #586, opened 2018-09-19
fix: make global toggle checkbox less ambiguous

olizilla:
_what will happen if I uncheck the box?_
screenshot 2018-09-19 10 04 52

This PR is just a tiny tweak so the checkbox label says what it will do if it is checked.

- [ ] **Global toggle: Suspend all IPFS integrations**

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>


olizilla:
I haven't pushed anything to transifex while it awaits your judgement @lidel

olizilla:
Ah yes, that's the one.

PR #590, opened 2018-09-20
Ship the new WebUI

lidel:
This PR aims to replace old Web UI with the bundled one.
It is important to get this right, as we will use similar approach for [ipfs-share-files](https://github.com/ipfs-shipyard/ipfs-share-files).

Remaining work:

- [x] webui uses the API object from Companion
  - `ipfsCompanion.ipfs` in background page
- [x] Ability to bundle arbitrary release or revision of webui
  - Current approach is bit crude: we just do regular build in `node_modules/..` subdir  and copy artifacts. Refinements will be introduced in following commits
- [x] <del>webui shows IPFS Companion as API Provider</del>
    - we decided to show "ipfs-companion" for now: https://github.com/ipfs-shipyard/ipfs-webui/issues/787
- [x] Avoid rebuilding webui every time Companion is built
  - Ideally we would not build it ourselves, but have tagged releases with prebuilt version 
  - If that does not pass Mozilla review, we can fallback to building it once on dependency install   
     - research `postinstall` step
- [x] Reload bundled webui when API backend changes 
- [x] Display webui version 
   - https://github.com/ipfs-shipyard/ipfs-webui/issues/842 | PR: https://github.com/ipfs-shipyard/ipfs-webui/pull/844 
- [x] Use Companion's API in Chrome
   - Upstream fix needs to be bubbled up to webui first: https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/14

Nice to have, but can be separate PR:
- [ ] Companion updates older webui to the bundled version
- [x] Research decreasing extension size
   - Companion size grew from ~4MB to ~10MB
     - Until https://github.com/mozilla/web-ext/issues/940 lands we remove maps manually: size is now under 7MB

## Screenshot (check location bar)

> screenshot


lidel:
Any thoughts on how to improve the bundling process? 
So far I would really like to keep webui "a black box" and ideally just include pre-built release version. 
But perhaps there is a better way?

lidel:
I created some orchestration for fetching prebuilt webui from IPFS:

https://github.com/ipfs-shipyard/ipfs-companion/blob/b170b221dd3c628eb33f34a967797dcdad32609d/package.json#L26-L31

- It tries to fetch it via `ipfs` but if command is not on the path or yields error, fallback will fetch CID from public gateway
- It works on windows ;)

add-on/src/lib/ipfs-client/index.js
```diff
@@ -26,18 +26,24 @@ async function destroyIpfsClient () {
...
       client = null
-      await _reloadIpfsClientDependents()
+      await _reloadIpfsClientDependents() // sync (API stopped working)
     }
   }
 }
 
+function _isWebuiTab (url) {
+  const bundled = !url.startsWith('http') && url.includes('/webui/index.html#/')
+  const ipns = url.includes('/webui.ipfs.io/#/')
```
lidel: @olizilla FYI this will detect webui loaded from:
- `https://webui.ipfs.io`
- `http://127.0.0.1:8080/ipns/webui.ipfs.io`

olizilla: `#/` will only be added by the app after the initial bundle has loaded (unless the user has bookmarked it explicitly)...

lidel: Yes, that was on purpose to remove false-positive matches on random websites and avoid double-reload when quickly switching between backends.

lidel:
Looks stable enough to ship to Beta. Let's smoke-test it there.

PR #616, opened 2018-11-08
fix: CORS false-positive in Chrome 71

### LINKED ISSUE #615: Can't Connect to Local Node in Chrome with Beta versions v2.5.1.11970 or v2.5.1.12010
As of v2.5.1.11970 (Beta) and v2.5.1.12010 (Beta) IPFS Companion can't connect to my local node under Chrome, and yet Firefox works just fine.

- Switching to the custom gateway instantly shows that the gateway is offline and switches to the internal js-ipfs node by default. 
- Disabling automatic switching keeps the extension in external mode, showing the gateway and API address correctly but version and peer appears offline, and fails to work.
- As tested on Firefox with the extension as well as just by directly accessing the local node's API from the same Chrome instance, everything is working fine on my end.

Versions:

- Xubuntu 18.04 LTS
- Chrome `70.0.3538.77 (Official Build) (64-bit)`
- Firefox `63.0 (64-bit)`
- Go-IPFS `0.4.18`
- IPFS Companion
  - `v2.5.1.11970 (Beta)`
  - `v2.5.1.12010 (Beta)`

When I enabled error logging for the extension in Chrome I get the following:

- Fresh install with default automatic external/internal switching enabled:
  - :warning: `Error while ipfs.swarms.peers: SyntaxError: Unexpected number in JSON at position 4`
  - Chrome's extension error log doesn't hang.
- Thereafter, with automatic mode disabled and switched to external node (but with DNSLink enabled by default):
  - :exclamation:  `Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/`
    - Context: `dist/background/background.html`
    - Stack Trace: `dist/bundles/backgroundPage.bundle.js:2 (readDnsLinkFromTxtRecord)`
  - Chrome's extension logging page immediately hangs the tab, and reaches 100% CPU usage.
  - This is clearly related to the DNSLink functionality which has also always worked for me even before it was enabled by default. 
    - Disabling DNSLink removes the error and doesn't hang the error logging page.
- Thereafter, with DNSLink disabled, but still on the custom node.
  - Just the JSON error as before, and nothing else.

lidel:
This PR restores [this workaround](https://github.com/ipfs-shipyard/ipfs-companion/blob/7a6fac1d9dc344b426120e121e158d220d377be5/add-on/src/lib/ipfs-request.js#L153-L161) for localhost API closes #615

(Fast tracking as this needs to land in Beta and Stable)

### DIFF
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -108,8 +108,8 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
     onBeforeSendHeaders (request) {
       const state = getState()
 
-      // Skip if IPFS integrations are inactive or request is marked as ignored
-      if (!state.active || isIgnored(request.requestId)) {
+      // Skip if IPFS integrations are inactive
+      if (!state.active) {
         return
       }
 
@@ -150,10 +150,9 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
             request.requestHeaders.push(expectHeader)
           }
         }
-        // For some reason js-ipfs-api sent requests with "Origin: null" under Chrome
-        // which produced '403 - Forbidden' error.
+        // For some reason js-ipfs-api sends requests with "Origin: null" under Chrome
+        // which produces '403 - Forbidden' error.
         // This workaround removes bogus header from API requests
-        // TODO: check if still necessary
         for (let i = 0; i < request.requestHeaders.length; i++) {
           let header = request.requestHeaders[i]
           if (header.name === 'Origin' && (header.value == null || header.value === 'null')) {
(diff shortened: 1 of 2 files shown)


lidel:
To clarify, our current approach is to remove `Origin: null` from requests made to API URL. 
This removes the need for users to add `Access-Control-Allow-Origin: "null"` to go-ipfs config:

> It may seem safe to return Access-Control-Allow-Origin: "null" , but the serialization of the Origin of any resource that uses a non-hierarchical scheme (such as data: or file: ) and sandboxed documents is defined to be "null". Many User Agents will grant such documents access to a response with an Access-Control-Allow-Origin: "null" header, and any origin can create a hostile document with a "null" Origin. The "null" value for the ACAO header should therefore be avoided.
> 
>> The simple string comparison of CORS as applied to "null" is controversial. Some believe that "null" should be treated as a keyword token indicating the lack of an Origin, which, when tested, should never compare as equal to another "null" Origin. (As is the case with null values in SQL, for example.) It is unwise to build systems which rely on the "null" equals "null" comparison as this behavior may change in the future. 
> 
> – https://w3c.github.io/webappsec-cors-for-developers/#avoid-returning-access-control-allow-origin-null

Issue #589, opened 2018-09-20
Meta: window.ipfs v2


lidel:
(This is a placeholder issue, will be updated in future)

The goal: **`window.ipfs` v2** ships and is production ready 
(tl;dr: injected on-demand, expose only requested APIs)

# TL;DR

Future-proof way of using `window.ipfs` is:

```js
if (window.ipfs) {
  try {
    const ipfs = await window.ipfs.enable({commands: ['add','cat']})
    const [{ hash }] = await ipfs.add(Buffer.from('=^.^='))
    const data = await ipfs.cat(hash)
    console.log(data.toString()) // =^.^=
  } catch (err) {
    // find more code samples at https://git.io/fjAu8
...
```
 
# Implementation Status

https://github.com/ipfs-shipyard/ipfs-companion/pull/619:
- [x]  scaffolding for `window.ipfs.enable`
- [x] decide if display prompt for `.enable()` (no params)  and keep no-dialog for things like `ipfs.cat` or  display no initial dialog but prompt every api instead
  - **A:** no change for now, logic stays the same
- [x] Implement Bulk Permission Prompt at `window.ipfs.enable` (https://github.com/ipfs-shipyard/ipfs-companion/pull/619)
  - [x] `window.ipfs.enable({commands: ['id','version'] })` 
  - [x] update docs
  - [x] tests
- [x] UX+API for requesting permissions in bulk 
  - UI remains the same to keep this PR small
- [x] add support for future opt-in `experiments`
  - separate namespace for optional experimentation without any stability guarantees
  - PoC `await window.ipfs.enable({ experiments: { ipfsx: true } })` will return [`ipfsx`](https://github.com/alanshaw/ipfsx) version of the API
- [x] add deprecation warning to API calls executed on `window.ipfs`

### Open Questions

- [x] Should we rename  it to `window.ipfsProxy.enable`? (to avoid breaking websites that manually detect `window.ipfs` and execute `window.ipfs.<command>`) or is it enough to deprecate old use for 3 months  and then remove support for old ways?
  - We will keep the name and go with deprecation period (2019-04-01), as noted in https://github.com/ipfs-shipyard/ipfs-companion/pull/619#issuecomment-449409199
- [x] now that we have `window.ipfs.enable` should we require permission from user for all commands? (remove acl-whitelist for things like `dag.get` ?)
[...]

PR #619, opened 2018-11-14
Add window.ipfs.enable(opts) (Bulk Permission Prompt)

lidel:
> Context and main discussion: #589 

This PR adds `await window.ipfs.enable(opts)` as the new way of obtaining IPFS API Proxy object. When provided with a list of commands it will request access rights in bulk and return a promise with IPFS API instance or throw an error (details below).

The goal is to move away from synchronous way of accessing API instance and provide UX incentive  for developers to use `window.ipfs.enable` instead of old API.

This change is backward compatible, old API remains to be exposed on `window.ipfs` for the time being.

> 

#### ACL Management

When called without any arguments, command will just return API instance equal to the old `window.ipfs` or throw an error if IPFS Proxy is disabled in Preferences.

When called  with options object (eg. `{ commands: ['id','peers'] }`) ACLs for specified commands will be validated:

- if any of the commands is denied or blocked, function will throw
- if any of the commands require user approval, user will be presented
with a single prompt dialog that lists all requested permissions and URL
that requests them
  - if user approves, ACLs are saved and future calls will not trigger
  prompt
  - if user denies, ACLs are saved and an error is thrown for current
  and all future executions (unless user removed scope from blacklist)

To keep change set small I'd like to merge this as-is and work on "Future TODOs" (below)  in separate PRs. 

### Tasks

- [x]  scaffolding for `window.ipfs.enable`
- [x] decide if display prompt for `.enable()` (no parms)  and keep no-dialog for things like `ipfs.cat` or  display no initial dialog but prompt every api instead
  - **A:** no change for now, logic stays the same
- [x] Implement `window.ipfs.enable({commands: ['id','version'] })` 
  - [x] update docs
  - [x] tests
- [x] UX+API for requesting permissions in bulk 
  - UI remains the same to keep this PR small
- [x] add support for future opt-in `experiments`
  - separate namespace for optional experimentation without any stability guarantees
  - PoC `await window.ipfs.enable({ experiments: { ipfsx: true } })` will return [`ipfsx`](https://github.com/alanshaw/ipfsx) version of the API
- [x] add deprecation warning to API calls executed on `window.ipfs`

### Open Questions

- [x] Should we rename  it to `window.ipfsProxy.enable`? (to avoid breaking websites that manually detect `window.ipfs` and execute `window.ipfs.<command>`) or is it enough to deprecate old use for 3 months  and then remove support for old ways?
  - We will keep the name and go with deprecation period (2019-04-01), as noted in https://github.com/ipfs-shipyard/ipfs-companion/pull/619#issuecomment-449409199
- [x] now that we have `window.ipfs.enable` should we require permission from user for all commands? (remove acl-whitelist for things like `dag.get` ?)
  - We will remove whitelist, and nothing will be allowed without explicit permission from the user (rationale: https://github.com/ipfs-shipyard/ipfs-companion/pull/619#issuecomment-450900594)

### Future TODOs (separate PRs)

- [ ] Check if it is feasible to disable `window.ipfs` injection via manifest in Chromium
  - now that we have async `window.ipfs.enable` we should revisit 
  - context why we did it in the first place: https://github.com/ipfs-shipyard/ipfs-companion/issues/362#issuecomment-362231167
- [ ] Look if it is feasible to disable `window.ipfs` injection via manifest in Chromium
  - now that we have async `window.ipfs.enable` we should revisit 
  - context why we did it in the first place: https://github.com/ipfs-shipyard/ipfs-companion/issues/362#issuecomment-362231167
- [ ] thin content script, ipfs proxy client is lazy-loaded
  - [ ] stop exposing methods on `window.ipfs`
  - [ ] minimize the size of content script responsible for `window.ipfs`
  - [ ] lazy-init IPFS Proxy client on first call to `window.ipfs.enable()`
- [ ] improve UX of permission dialog
- [ ] remove Buffer - https://github.com/ipfs-shipyard/ipfs-companion/issues/637
- [ ] add support for Uint8Array/ArrayBuffer natively – https://github.com/ipfs/interface-ipfs-core/issues/404
- [ ] standardize error messages to follow js-ipfs/js-ipfs-http-client (after discussion in https://github.com/ipfs/js-ipfs/pull/1746 is finalized and merged)


add-on/src/contentScripts/ipfs-proxy/page.js
```diff
@@ -1,7 +1,40 @@
...
+  return {
+    enable: async (opts) => {
+      // (This should be a lazy-load)
+      // Send message to proxy server for additional validation
+      // eg. trigger user prompt if a list of requested capabilities is not empty
+      // or fail fast and throw if IPFS Proxy is disabled globally
+      await require('postmsg-rpc').call('proxy.enable', opts)
+      // Additional client-side features
+      if (opts && opts.experiments) {
+        if (opts.experiments.ipfsx) {
```
lidel: The idea is to enable opt-in experiments such as ipfsx:
```js
let ipfsx = await window.ipfs.enable({
  commands: ['add','files.addPullStream'],
  experiments: { ipfsx: true }
})
```

alanshaw:
> * Should we rename  it to `window.ipfsProxy.enable`? (to avoid breaking websites that manually detect `window.ipfs` and execute `window.ipfs.<command>`) or is it enough to deprecate old use for 3 months  and then remove support for old ways?

I go for "deprecate old use for 3 months and then remove support for old ways" and use something like https://www.npmjs.com/package/deprecate to help communicate it to developers. Perhaps we could search github for projects using it and open an issue for them? `window.ipfs` is a desireable namespace to have - it's catchy, and honestly, we can't swtich namespace everytime we want to make a breaking change 😆 

> * now that we have window.ipfs.enable should we require permission from user for all commands? (remove acl-whitelist for things like dag.get ?)

What would be the reason for this change?

lidel:
Good points for keeping `window.ipfs` – you are right, it is still marked as "experiment", so we are off the hook for breaking it, as long we deprecate it over time.

Before we merge this, I will add deprecation notice to commands exposed directly on `window.ipfs`

> > * now that we have window.ipfs.enable should we require permission from user for all commands? (remove acl-whitelist for things like dag.get ?)
> 
> What would be the reason for this change?

My thinking was around enabling "privacy by default", where dapps have no access to IPFS node unless user granted access.

Rationale: 
- We added acl-whitelist as a proviacy compromise to avoid DoSing user with multiple dialogs. Adding `window.ipfs.enable` removes the need for such compromise. 
    - To clarify what I mean by that: due to the way IPFS works there are no "safe" commands. Even a simple read-only  `dag.get` can be abused to uniquely identifying user (by requesting unique CID and then finding user's multiaddr via `findprov <cid>`).

lidel:
Thanks, I've added deprecation warning to `window.ipfs.<cmd>` (I think its better to be noisy about this, so its printed on every call): 

> 

Does the text look ok?

> Presumably you'd need permission to **add** something first to become a provider for that content?

Yeah, someone else could pre-generate unique CIDs and just request them and see who is the _second_ provider. 

Anyway, I removed ACL whitelist and updated cods in https://github.com/ipfs-shipyard/ipfs-companion/pull/619/commits/756b177bf8875410476f6a3e0c47245cef5dbf15. 
Everything requires an explicit approval from the user now. 

I'd like to ship it to Beta this week and continue in separate PR – any final feedback before we merge this?

lidel:
Continued in https://github.com/ipfs-shipyard/ipfs-companion/issues/589

PR #642, opened 2018-12-17
[WIP] New Share Screen (ipfs-share-files)

lidel:
This PR adds bundled and minimized version of [ipfs-share-files](https://github.com/ipfs-shipyard/ipfs-share-files) and opens it in addition to the old Share screen (in a new tab). After app is ready we will remove the old screen, for now we open both:

> 2018-12-17-014732_1279x948_scrot

Thanks to [this code in ipfs-redux-bundle](https://github.com/ipfs-shipyard/ipfs-redux-bundle/blob/v4.1.2/src/companion/index.js) the Share App is able to detect it runs inside of IPFS Companion and use its API instance directly. 

@fsdiogo FYI to update to a new version of ipfs-share-app we will just change the CID in `build:share` script:

https://github.com/ipfs-shipyard/ipfs-companion/blob/eaf97847b0abe4a21258c9b665698576797a6c56/package.json#L33

## TODO

- Feature parity with old screen
  (Get things to the point where users won't have a feeling of functional regression)
  - [ ]  Ability to copy gateway link to the wrapping directory https://github.com/ipfs-shipyard/ipfs-share-files/issues/69
- [ ] Sort out pinning
   - Detect ipfsCompanion and add uploaded content to MFS?
- [x] Fix direct links https://github.com/ipfs-shipyard/ipfs-share-files/issues/68
- [x] Fix overflow issue https://github.com/ipfs-shipyard/ipfs-share-files/issues/67


jessicaschilling:
@lidel Is there any point in keeping this PR around?

lidel:
Closing. This PR mostly acted as a PoC, but was never meant to be:

Since then we've had issues at browser extension stores related to build reproducibility (that is why we don't bundle webui on Firefox) and bundling this would be similar source of trouble.

- In the future, we will refine the sharing user story, but will probably do the most of it within webui
- Revitalization of share.ipfs.io can be tracked in https://github.com/ipfs-shipyard/ipfs-share-files/issues/75

Issue #648, opened 2018-12-27
Warn about mixed content issues when changing the gateway


da2x:
The options UI should display a warning about content being blocked because of mixed-content protections when the “Custom Local Gateway” option doesn't match the pattern `https://.*/|http://127\.\d+\.\d+\.\d+:\d+/` (any HTTPS origin, or any origin on the loopback interface). This only needs to be visible when the pattern doesn't match.

Mockup: [/ipfs/QmezHbfHAg7NyzrbqgeRpvJJPx5h7eikHb77HLBwkm5jps](https://cloudflare-ipfs.com/ipfs/QmezHbfHAg7NyzrbqgeRpvJJPx5h7eikHb77HLBwkm5jps)

Configurations like #19/#328/#604 and #645 all indicate that such a warning would be useful.

PR #650, opened 2019-01-02
Add warning about mixed content issues

lidel:
> Closes https://github.com/ipfs-shipyard/ipfs-companion/issues/648

This PR checks if URL would trigger mixed content error and if check is positive displays an inline warning message:

> 

@olizilla does the copy look ok?

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -239,6 +239,10 @@
     "message": "URL of preferred HTTP2IPFS Gateway",
     "description": "An option description on the Preferences screen (option_customGatewayUrl_description)"
   },
+  "option_customGatewayUrl_warning": {
+    "message": "Warning: IPFS content can be blocked from loading on HTTPS websites unless your gateway URL starts with “http://127.0.0.1” or “https://”",
+    "description": "A warning on the Preferences screen, displayed when URL does not belong to Secure Context (option_customGatewayUrl_warning)"
+  },
   "option_useCustomGateway_title": {
     "message": "Use Custom Gateway",
     "description": "An option title on the Preferences screen (option_useCustomGateway_title)"
--- a/add-on/src/options/forms/gateways-form.js
+++ b/add-on/src/options/forms/gateways-form.js
@@ -5,6 +5,10 @@ const browser = require('webextension-polyfill')
 const html = require('choo/html')
 const { normalizeGatewayURL } = require('../../lib/options')
 
+// Warn about mixed content issues when changing the gateway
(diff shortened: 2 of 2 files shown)


add-on/_locales/en/messages.json
```diff
@@ -239,6 +239,10 @@
     "message": "URL of preferred HTTP2IPFS Gateway",
     "description": "An option description on the Preferences screen (option_customGatewayUrl_description)"
   },
+  "option_customGatewayUrl_warning": {
+    "message": "Warning: IPFS content can be blocked from loading on HTTPS websites unless your gateway URL starts with “http://127.0.0.1” or “https://”",
```
olizilla: Is good. We can drop the "Warning:" prefix as it seems pretty clear from the copy, and use `will` to say it more clearly.

```suggestion
     "message": "IPFS content will be blocked from loading on HTTPS websites unless your gateway URL starts with “http://127.0.0.1” or “https://”",
```

add-on/src/options/forms/gateways-form.js
```diff
@@ -5,6 +5,10 @@ const browser = require('webextension-polyfill')
 const html = require('choo/html')
 const { normalizeGatewayURL } = require('../../lib/options')
 
+// Warn about mixed content issues when changing the gateway
+// https://github.com/ipfs-shipyard/ipfs-companion/issues/648
+const secureContextUrl = /^https:\/\/.*|^http:\/\/127.0.0.1/
```
da2x: This should include `^http:\/\/\[\:\:1\]/` (IPv6 loopback address) as it’s treated the same way as 127.0.0.1 in browsers.

add-on/_locales/en/messages.json
```diff
@@ -239,6 +239,10 @@
     "message": "URL of preferred HTTP2IPFS Gateway",
     "description": "An option description on the Preferences screen (option_customGatewayUrl_description)"
   },
+  "option_customGatewayUrl_warning": {
+    "message": "IPFS content will be blocked from loading on HTTPS websites unless your gateway URL starts with “http://127.0.0.1” or “https://”",
```
da2x: This should include `http://[::1]`.

da2x:
Sorry, but I forgot about IPv6 in #648. It’s one more valid address that is whitelisted as secure by browsers. The inclusion of a third address does make the warning message messy, though. Not sure how to improve it without reformatting it as a list or maybe even linking to a wiki page with valid addresses and more information?

lidel:
Good catch @da2x! I've added it and it's not that bad actually (a technical person will know what's up, a non-technical won't need to touch this field anyway).

>

PR #658, opened 2019-01-14
update zh_CN

alx696:
翻译补全。


lidel:
Thank you for contributing translation.
To avoid duplicated work, we don't accept direct PRs: translations need to be submitted using Transifex project at https://www.transifex.com/ipfs/ipfs-companion/
You can find more information at [/docs/localization-notes.md](https://github.com/ipfs-shipyard/ipfs-companion/blob/master/docs/localization-notes.md) and [ipfs/i18n](https://github.com/ipfs/i18n)

PR #670, opened 2019-01-27
Add bundle:brave npm script and Brave manifest

bbondy:
These changes makes it so a bundle for Brave can be created which have chrome.socket APIs exposed. 

Note that `key` manifest JSON property is needed to get a consistent
extension ID when developing across machines.  Since the APIs are only whitelisted for this extension ID, this is needed. 

If you're wondering where the `key` manifest JSON property value can be obtained from, you can get it by installing IPFS Companion from the Chrome store and looking in your profile directory at the written manifest.  For example, on macOS it would be here: `~/Library/Application\ Support/BraveSoftware/Brave-Browser-Dev/Default/Extensions/nibjojkomfdiaoajekhjakgkdhaomnch/2.7.0_0/manifest.json`

This will not work with the current live release of Brave, however you can build a new `ipfs` branch that I created to get it working. We can continue to work within that `ipfs` branch until we're ready to ship something.

I have a branch of brave/brave-browser and brave/brave-core named `ipfs`.

To get a build of Brave with this branch, follow the build-instructions here:
https://github.com/brave/brave-browser/wiki

But after cloning checkout the branch `ipfs`

If you already have a build of Brave, you can just switch to the `ipfs` branch in the root of your `brave/brave-browser` clone.
Then run `npm run init` one time so the dependency brave-core (https://github.com/brave/brave-core) which is mounted in `src/brave` gets updated to use its `ipfs` branch.
Then run `npm run build` to build the changes.

Load unpacked but bundle IPFS companion with `npm run build bundle:brave`.

### DIFF
--- a/add-on/manifest.brave.json
+++ b/add-on/manifest.brave.json
@@ -1,0 +1,15 @@
+{
+  "sockets": {
+    "udp": {
+      "send": "*",
+      "bind": "*"
+    },
+    "tcp": {
+      "connect": "*"
+    },
+    "tcpServer": {
+      "listen": "*:*"
+    }
+  },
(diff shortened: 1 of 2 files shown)


bbondy:
Sorry I just noticed you have a similar PR here:
https://github.com/ipfs-shipyard/ipfs-companion/compare/feat/brave-build-with-chrome-sockets

You can just add the `key` property to yours if you prefer and close this one.

PR #684, opened 2019-02-17
Avoid creating world-writable build results

bmwiedemann:
World writable files are a severe security issue
because any process on the system can modify them.

Without this patch, the build dir and everything within
had 777 or 666 as permission bits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "ci": "run-s ci:*",
     "ci:install": "npx yarn@1.12.3 install --frozen-lockfile || npx yarn@1.12.3 install --frozen-lockfile",
     "ci:test": "npx yarn@1.12.3 test",
-    "ci:build": "./ci/update-manifest.sh ; npx yarn@1.12.3 build ; chmod -R ugo+rwX build/ add-on/",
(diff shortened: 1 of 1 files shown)


bmwiedemann:
This bug was added in commit 72c56da2ff and modified multiple times since.

lidel:
Thank you for reporting this. 

There is additional complication with this PR: `ci:build` aims to run inside of Docker to produce reproducible artifacts in `build` (a mounted directory from host). Removing writable bit makes it impossible to remove `build` by regular user, which breaks cleanup at CI:

```console
$ npm run release-build && rm -rf build
...
rm: cannot remove 'build/firefox/ipfs_companion-2.7.5.zip': Permission denied
```

I decided to remove problematic surface entierely by running docker build under normal user (which removed the need for permission escalation). 

Closing this one, as it is superseded by https://github.com/ipfs-shipyard/ipfs-companion/pull/685 which [`removes chmod` from mentioned line](https://github.com/ipfs-shipyard/ipfs-companion/pull/685/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R61).

PR #688, opened 2019-02-26
chore: update to Web UI 2.4.0

fsdiogo:
Check the [release notes](https://github.com/ipfs-shipyard/ipfs-webui/releases/tag/v2.4.0) for more info.

### DIFF
--- a/add-on/src/lib/state.js
+++ b/add-on/src/lib/state.js
@@ -24,7 +24,7 @@ function initState (options) {
   state.dnslinkPolicy = String(options.dnslinkPolicy) === 'false' ? false : options.dnslinkPolicy
   // store info about 'blessed' release of Web UI
   // which should work without setting CORS headers
-  state.webuiCid = 'QmXc9raDM1M5G5fpBnVyQ71vR4gbnskwnB9iMEzBuLgvoZ' // v2.3.3
+  state.webuiCid = 'Qmb6o8HKrouK2WXVFS8753BArqnx5xz68tk6779ncPxpd8' // v2.4.0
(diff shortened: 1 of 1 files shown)


lidel:
@fsdiogo Is my understanding correct that the only problem with 2.4.0 is opt-out based on the presence of DNT header, as noted in https://github.com/ipfs/go-ipfs/pull/6025 ?

IPFS Companion is able to manually add DNT header to requests sent to Web UI, effectively disabling it by default and making analytics an opt-in.

https://github.com/ipfs-shipyard/ipfs-companion/blob/6578886bb90abd648dca519d5a0e50e0f839ae57/add-on/src/lib/ipfs-request.js#L131-L134

That way we could ship this without waiting for [v2.5.0](https://github.com/ipfs-shipyard/ipfs-webui/issues/978).

Does it make sense? Or is a patch realease landing this week?

fsdiogo:
> @fsdiogo Is my understanding correct that the only problem with 2.4.0 is opt-out based on the presence of DNT header, as noted in ipfs/go-ipfs#6025 ?

Yes, that is correct.

I think it would be best to just cut a patch release so Companion won't have to do that and everything is streamlined.

I'll release a new Web UI version today and open a PR here!

fsdiogo:
Actually, it's best to wait per:

> I'm gonna merge this now, so we get opt-in deployed to webui.ipfs.io now, and open an new PR for the rest.

_Originally posted by @olizilla in https://github.com/ipfs-shipyard/ipfs-webui/pull/981#issuecomment-467845102_

Issue #689, opened 2019-02-28
[META] Upgraded UI for IPFS Companion


lidel:
#### <del>ETA April 2020:</del> :rofl: 
Next steps for this need to be an analysis of the entire discussion chain and breaking what hasn't already been addressed out into more discrete, workable parts.

#### Original issue text:
Let's use this issue as an open space for discussing how we could improve current UI,
namely the main menu displayed after clicking on Browser Action.

@ipfs-shipyard/gui

lidel:
I've been looking into ways we could make actions specific to Active Tab more prominent.

Initials mockups:

| A | B  | C | 
| ---- | ---- | ---- |
| ![screenshot_2019-02-28 53101099-68d17700-3529-11e9-890b-0fd64b8a52e1 gif gif image 744 x 570 pixels](https://user-images.githubusercontent.com/157609/53589308-5ab9d100-3b8f-11e9-92ed-585421df4e27.png)  | ![updates-2019-02-23--21-30-03](https://user-images.githubusercontent.com/157609/53291434-9fbabd80-37b3-11e9-896f-7cd316973484.png) | ![companion-active-tab](https://user-images.githubusercontent.com/58871/53421062-dfacbb00-39d4-11e9-9316-c06dd559c6c7.png) |

@hugomrdias in https://github.com/ipfs-shipyard/ipfs-companion/pull/687#issuecomment-467542577
> I made a prototype which packs lots of ideas out of this PR scope
> 
> https://codesandbox.io/s/14y8r7qjo3
> 
> but basically i like toggles with explicit label and tooltip explaining everything like i'm 5. 
> I prefer label texts that don't change with state.

| Default | Expanded Node Info |
| ---- | ---- |
| [![screenshot_2019-02-28 codesandbox](https://user-images.githubusercontent.com/157609/53588657-c13def80-3b8d-11e9-9717-0a3a4f4468aa.png)](https://codesandbox.io/s/14y8r7qjo3) | [![screenshot_2019-02-28 codesandbox 1](https://user-images.githubusercontent.com/157609/53588711-e599cc00-3b8d-11e9-9364-41e8dc4a308b.png)](https://codesandbox.io/s/14y8r7qjo3) |

lidel:
### New UI by Hugo

@hugomrdias posted well-thought mockup of complete rework in https://github.com/ipfs-shipyard/ipfs-companion/pull/687#issuecomment-467542577
[...]

PR #687, opened 2019-02-20
Per-site Redirect Opt-out

lidel:
> This PR aims to:
> - close https://github.com/ipfs-shipyard/ipfs-companion/issues/647 
> - close https://github.com/ipfs-shipyard/ipfs-companion/issues/633

### Icon-based global redirect toggle 

To avoid miss-clicks, confusion, and to disincentivise disabling global redirect, we moved global toggle from Browser Action menu to the utility icon row, and used custom "redirect" icon:

| Enabled  | Disabled |
| ---- | ---- |
| on-2019-01-20--22-59-30 |  | 

> When you start seeing the face above, you cannot unsee it.

### Redirect opt-out per site

- menu item in Active Tab section enables user to disable gateway redirect on current website
- when clicked on regular site toggles redirect for current FQDN and all its subdomains
- when clicked on /ipns/<fqdn>/ (DNSLink) website, toggles redirect for <fqdn>
- after redirect preference changes for current website, the tab is reloaded
- DNSLink websites are reloaded to with URL change between IPNS path and original URL
  > peek 2019-02-20 16-05
  > _Redirect opt-out on DNSLink website loaded from local gateway restores original URL_
- Redirect preference applies not only to request with FQDN of the active tab, but also to all subresource requests that have it in `originUrl` (Firefox) or `Referer` header (Chrome). This means toggle is useful to restore functionality of complex websites such as d.tube (which uses IPFS subresources from various subdomains):
  >  peek 2019-02-20 16-08
  > _Redirect opt-out on regular website: functionality is restored thanks to automatic reload_

#### Opt-out Test URLs

Use below to test the per-site opt-out:

- http://docs.ipfs.io (DNSLink website at `/ipns/docs.ipfs.io`)
- https://d.tube/#!/v/elenasteem/890wiapc (regular website with custom JS)
  - video won't load if redirect is active on initial load
  - disabling redirect for `d.tube` restores playback (note: additional toggle does not impact current video, as data is cached by the website)

### Other
- added subtle animation to utility icons
- global redirect icon will enable integrations if clicked when in suspended state
- menu items specific to the Active Tab are in a section marked with additional border 
  (just a prototype, needs refinement)

## TODO

- [x] fix tests 
- [x] improve UX of "Active Tab" section in browser Action menu
- [x] minimize `redirect-icon.js` (100K right now)
- [x] expose per-site opt-outs as editable field on _Preferences_ screen
  - mvp: one FQDN per line, sort lexicographically
[...]

PR #694, opened 2019-03-12
Show IPFS Actions on DNSLink Sites

lidel:
> cc #689 

When we introduced option to opt-out from redirect per site (#687 ) it came with a side effect of removing IPFS context actions. 

This PR:

-  Adds context actions on DNSLink sites (when redirect is disabled)
   Peek 2019-03-12 13-56
    >  This adds context actions such as "Copy IPFS path", "Copy CID", "Pin" to DNSLink websites without redirect. 
    > 
    > It includes refactoring of ipfsPathValidator to expose high level resolvers used in various places by the Companion:
    > 
    > - `resolveToPublicUrl`: always return a meaningful, publicly accessible URL that can be accessed without the need of IPFS client.
    > - `resolveToIpfsPath`: return a valid IPFS path that can be accessed with IPFS client.
    > - `resolveToImmutableIpfsPath`: same as resolveToIpfsPath, but the path is always immutable /ipfs/
    > - `resolveToCid`: returns direct CID without anything else
- Adds a bunch of tests
- Tweaks behavior of pin/unpin via browser action menu
  > In the past we closed menu after toggling pin status, but now we have a nice visual toggle switch and other uses of the switch do not close the menu, so we unified behavior and toggle action no longer closes the popup menu.
- Works around missing dnslink resolver under js-ipfs
  > js-ipfs v0.34 does not support DNSLinks in `ipfs*resolve` methods yet: ipfs/js-ipfs#1918
  > This temporary workaround detects known errors and falls back to using path from DNSLink cache. It won't be as fresh as original, but at least resolv will work.

### DIFF
--- a/add-on/src/lib/copier.js
+++ b/add-on/src/lib/copier.js
@@ -40,8 +40,8 @@ function createCopier (notify, ipfsPathValidator) {
     },
 
     async copyRawCid (context, contextType) {
+      const url = await findValueForContext(context, contextType)
       try {
-        const url = await findValueForContext(context, contextType)
         const cid = await ipfsPathValidator.resolveToCid(url)
         await copyTextToClipboard(cid, notify)
       } catch (error) {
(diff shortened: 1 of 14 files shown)


lidel:
@ericronne  yes, it will have correct colors (when #692 is merged)

PR #705, opened 2019-04-02
manifest.common.json: disallow running in incognito mode

### LINKED ISSUE #704: How to handle Incognito Mode?
> For extension developers, Firefox now fully supports the value `not_allowed` for the [manifest `incognito` key](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/incognito).  As with Chrome, specifying `not_allowed` in the manifest will prevent the extension from running or receiving events from private windows.
> – https://blog.mozilla.org/addons/2019/03/26/extensions-in-firefox-67/

Mikaela:
I am not entirely sure this came to the correct spot as there are no examples that I could see and I think `incognito` could logically be a permission. 

Requested in https://github.com/ipfs-shipyard/ipfs-companion/issues/704#issuecomment-478929763. Closes #704

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,7 +7,7 @@
       "@babel/preset-env",
       {
         "targets": {
-          "firefox": 65,
+          "firefox": 67,
           "chrome": 70
         }
       }
--- a/add-on/manifest.common.json
+++ b/add-on/manifest.common.json
@@ -58,5 +58,6 @@
     }
   ],
   "content_security_policy": "script-src 'self'; object-src 'self'; frame-src 'self';",
-  "default_locale": "en"
+  "default_locale": "en",
+  "incognito": "not_allowed"
 }
--- a/add-on/manifest.firefox.json
+++ b/add-on/manifest.firefox.json
@@ -8,7 +8,7 @@
   "applications": {
     "gecko": {
       "id": "ipfs-firefox-addon@lidel.org",
-      "strict_min_version": "65.0"
+      "strict_min_version": "67.0"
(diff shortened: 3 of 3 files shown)


Mikaela:
Done, however this may drop `privacy.resistfingerprinting` users to old version until Firefox ESR updates to 67 or higher thanks to [Firefox bug 1394448: Cannot install Addon withprivacy.resistFingerprinting==true](https://bugzilla.mozilla.org/show_bug.cgi?id=1394448).

As a workaround I think I am switching back to the beta channel which I left only because it doesn't work with Firefox Sync resulting me either not having IPFS Companion at all or having two of them.

lidel:
Thank you @Mikaela 

You raised a valid concern: we don't want to force people to switch to different browser channel if they want to use Companion Beta. Companion Beta should always work with Stable Firefox.

I believe a safe way to ship this is to park this PR until Firefox 67 lands in Stable by the end of this quarter.  Firefox 68 will become ESR in July soon after that: 

release-overview-high-res d3c9788f2dd1

Does this sound ok?

Mikaela:
I think it sounds ok (but I am staying in the beta as I am curious enough and following enough issues).

I think you may be overestimating my skills/abilities/knowledge on Firefox/webextensions/development and spent a long time staring at documentation after you asked me to look into this in #704 (without which I wouldn't have done anything).

lidel:
Firefox 68 landed  to ESR channel, let's merge this! :)

PR #726, opened 2019-06-03
feat(ux): suggest ipfs-desktop when no node is found

lidel:
This PR adds info about IPFS Desktop to the welcome page introduced in #565.
The Welcome Screen appears once, after initial install from add-on store.
I also fixed/tweaked some links to give a better onboarding experience.

### Before

TERMINAL OR DIE

### After

Terminal is life, but now there are options for everyone! :^)

fixed-2019-06-03--14-06-27

@olizilla @ericronne  feel free to tweak copy if it can be worded in a  better way.

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -556,12 +556,8 @@
     "description": "Install steps title (page_landingWelcome_installSteps_title_desktop)"
   },
   "page_landingWelcome_installSteps_cli_install": {
-    "message": "<0>Follow these instructions</0>.",
+    "message": "Follow <0>these instructions</0>, then start an IPFS daemon in your terminal:",
     "description": "Install steps copy (page_landingWelcome_installSteps_install)"
-  },
-  "page_landingWelcome_installSteps_cli_run": {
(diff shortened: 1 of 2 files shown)


fsdiogo:
Wondering if we should switch the order, to push people towards IPFS Desktop?

And then something like `The manual way` with the terminal instructions below.

olizilla:
Good idea @lidel! I shall be getting a TERMINAL OR DIE tattoo, but in the meantime, I'm with @fsdiogo and I think we should promote installing IPFS Desktop to manage your daemon, with the command line as a for experts kinda option.

ericronne:
I love favoring newbies by nudging them toward desktop. The word "daemon" might be too technical for them, too. Perhaps we say, "IPFS does not seem to be running." or something similar for the title, and then offer links to "Get IPFS Desktop" or go the manual/expert route, as @fsdiogo suggests?

lidel:
I moved IPFS Desktop as "the default" way and simplified language:




Thoughts?

lidel:
:+1:  ?

PR #732, opened 2019-07-10
fix(window.ipfs): stop injecting window.Buffer

### LINKED ISSUE #637: Do not inject window.Buffer
I found that some websites such as zhihu.com will use window.Buffer, Buffer injected by ipfs-companion would break the normal behavior of these websites. Buffer can be set elsewhere.

https://github.com/ipfs-shipyard/ipfs-companion/blob/50db9c08d5bf0e5297981ff474da07585aa8a8ce/add-on/src/contentScripts/ipfs-proxy/page.js#L6

lidel:
This PR removes code responsible for injection of `window.Buffer` when `window.ipfs` is enabled. 
We did that as a convenience tweak to make it easier to write demos/pocs, but polluting global scope introduced some issues on websites (Closes #637). 

Question: should we 
- (A) ship it now..
- (B) park until we remove  old `window.ipfs` altogether in September:
  https://github.com/ipfs-shipyard/ipfs-companion/blob/4ca9a0ce1842a9cf5979623bbf416493a18c2b2f/add-on/src/contentScripts/ipfs-proxy/page.js#L37 
- (C) park until we add API support for String and Uint8Array/ArrayBuffer natively (https://github.com/ipfs/interface-ipfs-core/issues/404)

### DIFF
--- a/add-on/src/contentScripts/ipfs-proxy/page.js
+++ b/add-on/src/contentScripts/ipfs-proxy/page.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const _Buffer = Buffer
 const { assign, freeze } = Object
 
 // TODO: (wip) this should not be injected by default into every page,
@@ -46,7 +45,4 @@ function createWindowIpfs () {
   return freeze(proxyClient)
 }
 
-// TODO: we should remove Buffer and add support for Uint8Array/ArrayBuffer natively
-// See: https://github.com/ipfs/interface-ipfs-core/issues/404
-window.Buffer = window.Buffer || _Buffer
 window.ipfs = window.ipfs || createWindowIpfs()
--- a/docs/examples/window.ipfs-fallback.html
+++ b/docs/examples/window.ipfs-fallback.html
@@ -1,4 +1,9 @@
 <!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8"/>
+</head>
+<body>
 <div id="output"></div>
 <script>
   (() => {
@@ -39,3 +44,5 @@
       .catch(log)
   })()
 </script>
+</body>
+</html>
--- a/docs/window.ipfs.md
+++ b/docs/window.ipfs.md
@@ -68,7 +68,7 @@ if (window.ipfs && window.ipfs.enable) {
 }
 ```
 
-Note that IPFS Companion also adds `window.Buffer` if it doesn't already exist.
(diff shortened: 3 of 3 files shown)


add-on/src/contentScripts/ipfs-proxy/page.js
```diff
@@ -46,7 +45,4 @@ function createWindowIpfs () {
   return freeze(proxyClient)
 }
 
-// TODO: we should remove Buffer and add support for Uint8Array/ArrayBuffer natively
-// See: https://github.com/ipfs/interface-ipfs-core/issues/404
-window.Buffer = window.Buffer || _Buffer
```
alanshaw: I think we should add support for `TypedArray` and `ArrayBuffer` before we ship this.

I'd be interested to know if [this](https://github.com/ipfs-shipyard/js-ipfs-http-client-lite/blob/master/src/add/normalise-input.js) covers all the bases.

lidel: Yeah, I've been thinking about it and (C) feels like a right way to do it.

Something I realized is that the first thing people do when they play with API is to add some text (`String`) to IPFS.  What if we remove friction of converting to bytes and accept `String` input and convert it internally to a typed array?

afaik something like https://encoding.spec.whatwg.org would do:

```js
const enc = new TextEncoder() // implicit utf-8
const bytes = enc.encode('This string will be converted to a Uint8Array'))
```

alanshaw: Yes, I've wanted this for a while and allowed it in https://github.com/ipfs-shipyard/ipfsx/blob/master/API.md#example-1

alanshaw: The tricky bit with it is destinguishing between content and a path in the file system, which I think is supported somewhere

hugomrdias: yeah thats the problem addFromFs

lidel: Is `addFromFs` the only place?
I think it is quite intuitive special case where `String` is interpreted as file path. It could run the generic `normalizeInput` as every other command and then convert bytes back to file path internally.

hugomrdias: Previously we had everything in the add, addReadable, addPullstream when we introduced addFromFs i think we kinda separated the `/file/path` input to this method only need further investigation

lidel:
Superseded by https://github.com/ipfs-shipyard/ipfs-companion/pull/777

PR #737, opened 2019-07-18
refactor: use Web UI provided by IPFS node

### LINKED ISSUE #736: Use Web UI version from IPFS Node
## Context

Every IPFS Node provides `:5001/webui` endpoint on the API port. It redirects to the latest 'blessed' CID that can be loaded from `:5001/ipfs/{cid}`. For security reasons only hardcoded CIDs can be loaded from API port.

Right now IPFS Companion ships with hardcoded version of Web UI, and that version can be bigger than the one at `:5001/webui`.

Companion can't load it from API port as it is not on the whitelist, so it loads it from the Gateway instead. That comes at a price of having two separate Origins and falling under CORS limitations. We don't want people to add CORS headers, as it would allow allow any website loaded from Gateway port to execute arbitrary requests on API port. Instead, we remove CORS check on webui path by supressing `Origin` header in requests made from `:8080/ipfs/{blessed_webui_cid}` (Gateway) to `:5001` (API). 
That way we keep CORS check from anything else on Gateway port.

Why we did this? It made it possible for us to iterate and ship new Web UI without waiting for new go-ipfs release (which took multiple months at the time).

## Proposed change

We are now in maintenance mode where regular go-ipfs releases are expected and webui itself is much more stable now.

It is time to:

- now: switch to `/webui` provided by go-ipfs (or js-ipfs) as the new default in Companion
  - PoC in PR: https://github.com/ipfs-shipyard/ipfs-companion/pull/737
- future: provide opt-in option for Companion users to load latest Web UI from `:5001/ipns/webui.ipfs.io` (requires ipns path to be added to API port whitelist in GO and JS)
  - PoC PR for Companion adding opt-in option: https://github.com/ipfs-shipyard/ipfs-companion/pull/738
    - PoC PR with discussion for go-ipfs: https://github.com/ipfs-shipyard/ipfs-companion/issues/736
    - PoC PR with discussion for js-ipfs: https://github.com/ipfs/js-ipfs/pull/2281

cc @olizilla  (https://github.com/ipfs-shipyard/ipfs-webui/issues/1038#issuecomment-494000620)

lidel:
This PR changes the way "Open Web UI" menu option works.

- Before: we kept CID of latest version and opened it from gateway port + executed a lot of magic to ensure it can talk to api port in secure way.

- After: we removed all the magic and just open the version provided on api port
  (default: `http://127.0.0.1:5001/webui/`)

Closes #736

### DIFF
--- a/add-on/src/lib/ipfs-client/index.js
+++ b/add-on/src/lib/ipfs-client/index.js
@@ -6,11 +6,9 @@ const debug = require('debug')
 const log = debug('ipfs-companion:client')
 log.error = debug('ipfs-companion:client:error')
 
-const browser = require('webextension-polyfill')
 const external = require('./external')
 const embedded = require('./embedded')
 const embeddedWithChromeSockets = require('./embedded-chromesockets')
-const { webuiCid } = require('../state')
 
 let client
 
@@ -33,7 +31,7 @@ async function initIpfsClient (opts) {
 
   const instance = await client.init(opts)
   easeApiChanges(instance)
-  _reloadIpfsClientDependents(instance) // async (API is present)
+  preloadWebui(instance, opts)
   return instance
 }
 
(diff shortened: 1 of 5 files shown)


lidel:
Superseded by https://github.com/ipfs-shipyard/ipfs-companion/pull/853

Issue #353, opened 2018-01-07
Support for Microsoft Edge?


rogersachan:

lidel:
Good question :)  There is an open task in #276 for smoke-testing it under MS Edge, but nobody picked it up yet (most of our community seems to be using Firefox/Chrome).

@rogersachan  do you have MS Edge and some spare time on your hands?  If so, try loading [ipfs_companion-2.1.0.zip](https://github.com/ipfs-shipyard/ipfs-companion/releases/download/v2.1.0/ipfs_companion-2.1.0.zip) (see [this instruction](https://docs.microsoft.com/en-us/microsoft-edge/extensions/guides/adding-and-removing-extensions))  and let us know if there were any errors (AFAIK Edge does not support notifications API yet, so you may need to inspect Browser Console). 

It would be an important data point telling us where MS Edge stands.

rogersachan:
![image](https://user-images.githubusercontent.com/7173984/34651264-ff2cd90e-f39b-11e7-8e72-817cf7993047.png)
I get an error before I can even load this extension. There are no other errors or messages in the console.

rogersachan:
![image](https://user-images.githubusercontent.com/7173984/34651319-a5a727c6-f39c-11e7-8202-89d7c247134a.png)
Updating background with this in order to fix the issue gives me another error.
![image](https://user-images.githubusercontent.com/7173984/34651326-c41a6736-f39c-11e7-9087-7b4d58825b9d.png)

rogersachan:
Deleting both 'browser_style' fields allowed me to import the extension. I'll start testing it as soon as I can.

rogersachan:
Attempting any of the links in #312 just sends me to the search engine.

rogersachan:
The background page is now giving me this error: 
```
Unable to initialize addon due to error Error: Invocation of form storage.local.get(object) doesn't match definition storage.local.get(optional string or array or object keys, function callback).
```

lidel:
This is a _very_ good feedback that cleared a lot! 👍 🔍 
Thank you @rogersachan!

Different handling of manifest keys is inconvenient, but can be managed. 
But the last error is a show-stopper.

[...]

PR #743, opened 2019-07-27
Add link to edge alongside the other browsers

rogersachan:
The link goes to the page where Edge can be downloaded for Windows 7/8/10 and macOS. The IPFS extension was shown to work in https://github.com/ipfs-shipyard/ipfs-companion/issues/353


lidel:
Thank you! :+1:

PR #744, opened 2019-07-30
Use updated versions of the AMO and Chrome Web Store badges

rogersachan:
These badges reflect the new logos of both add-on/extension stores.


rogersachan:
@lidel I unfortunately don't have any hardware suitable for pinning these images. If you can do it feel free to do so and close this pull request.

lidel:
@rogersachan I updated URLs. In the future, do not hesitate and use gateway URLs:  IPFS project  will take care of pinning :)

Thank you!

PR #745, opened 2019-07-31
[Moved to #937] feat: add permalink and IPFS path copy options

### LINKED ISSUE #740: Add "Copy Permalink" Options
The dropdown has three copy items, as can be seen here:

Screen Shot 2019-07-25 at 2 56 07 PM

Unfortunately, when accessing a resource through an IPNS system, such as DNSLink, the top two both copy links that use IPNS, rather than the underlying IPFS CID. If I want a link that continues to work, regardless of IPNS changes, I have to hit 'Copy CID' and build it manually. It would be convenient if options were available for creating CID-based links instead of IPNS-based ones.

DeedleFake:
Closes #740.

As this currently is, it introduces a UI bug in the pop-up in Firefox that's accessed by clicking the icon on the far-right, as opposed to the one in the URL bar. The contents of the pop-up are too large for it, and it's not scrollable, so the bottom of the menu becomes inaccessible. I'm not familiar enough with how extensions work to fix that without hunting through the codebase for stuff that looks like it might have to do with it, so if someone could let me know what I should take a look at for fixing that it would be quite helpful.

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -83,6 +83,10 @@
     "message": "Copy IPFS Path",
     "description": "A menu item in Browser Action pop-up and right-click context menu (panelCopy_currentIpfsAddress)"
   },
+	"panelCopy_currentIpnsAddress": {
+		"message": "Copy IPNS Path",
+		"description": "A menu item in Browser Action pop-up and right-click context menu (panelCopy_currentIpnsAddress)"
+	},
   "panelCopy_copyRawCid": {
     "message": "Copy CID",
     "description": "A menu item in Browser Action pop-up and right-click context menu (panelCopy_copyRawCid)"
@@ -91,6 +95,10 @@
     "message": "Copy Public Gateway URL",
     "description": "A menu item in Browser Action pop-up and right-click context menu (panel_copyCurrentPublicGwUrl)"
   },
+	"panel_copyCurrentPermalink": {
+		"message": "Copy Permalink",
+		"description": "A menu item in Browser Action pop-up and right-click context menu (panel_copyCurrentPermalink)"
+	},
   "pageAction_titleIpfsAtPublicGateway": {
     "message": "IPFS resource loaded via Public Gateway",
     "description": "A tooltip displayed over Page Action in location bar (pageAction_titleIpfsAtPublicGateway)"
--- a/add-on/src/lib/context-menus.js
+++ b/add-on/src/lib/context-menus.js
@@ -56,12 +56,16 @@ const contextMenuAddToIpfsKeepFilename = 'contextMenu_AddToIpfsKeepFilename'
 // Add X to IPFS
 const contextMenuAddToIpfsSelection = 'contextMenu_AddToIpfsSelection'
 // Copy X
-const contextMenuCopyCanonicalAddress = 'panelCopy_currentIpfsAddress'
+const contextMenuCopyCidAddress = 'panelCopy_currentIpfsAddress'
+const contextMenuCopyCanonicalAddress = 'panelCopy_currentIpnsAddress'
 const contextMenuCopyRawCid = 'panelCopy_copyRawCid'
 const contextMenuCopyAddressAtPublicGw = 'panel_copyCurrentPublicGwUrl'
+const contextMenuCopyPermalink = 'panel_copyCurrentPermalink'
+module.exports.contextMenuCopyCidAddress = contextMenuCopyCidAddress
 module.exports.contextMenuCopyCanonicalAddress = contextMenuCopyCanonicalAddress
 module.exports.contextMenuCopyRawCid = contextMenuCopyRawCid
 module.exports.contextMenuCopyAddressAtPublicGw = contextMenuCopyAddressAtPublicGw
+module.exports.contextMenuCopyPermalink = contextMenuCopyPermalink
 
 // menu items that are enabled only when API is online
 const apiMenuItems = new Set()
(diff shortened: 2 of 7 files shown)


lidel:
## Problem

Indeed, when we add two additional menu entries it goes over the max height browser allows:

| :anger:  `/ipns/` | :heavy_check_mark:  `/ipfs/` |
| ---- | ---- | 
| 2019-08-02-124708_3440x1440_scrot | 2019-08-02-124720_3440x1440_scrot |


I am afraid that is the hard limit:

>  In chrome the popup window dynamically changes (grows / shrinks) to suit changes to the content up to a maximum size of 800 x 600 px.
- https://bugzilla.mozilla.org/show_bug.cgi?id=1215025
- https://discourse.mozilla.org/t/what-is-the-max-size-for-a-browser-action-popup/16074

Also, now that I see all the Copy options, it feels the flat list does not work anymore: menu becomes too noisy.

## Looking for solutions 

We can either drop some items (eg. path ones), or change the UI of this to take less vertical space:


| default state | (A) details on hover  | (B) details on click |
| ---- | ---- | ---- | 
| default state shown when browser action is opened | clickable options are displayed inline, when user hovers over the "Copy" row | "Copy" is clickable and on click entire "Active Tab" section is replaced with "Copy" options | 
| 2019-08-02-130716_3440x1440_scrot |  2019-08-02-130716_3440x1440_scrot | 2019-08-02-124720_3440x1440_scrot | 

Thoughts? cc @ericronne

lidel:
This needs some design love, looping in design shop :)

**tl;dr** is: "Copy.." options take too much space vertically, details above.

jessicaschilling:
@lidel Looks like this got dropped at some point. To recap -- are you thinking something like this? 

image

lidel:
@jessicaschilling yes, you are reading my mind :ok_hand:  

IIRC the remaining problem here is that on a mutable website such as  IPNS or DNSLink http://en.wikipedia-on-ipfs.org/ this PR would add two items for copying an immutable link/path for the current snapshot: 

> Copy shareable link: https://en.wikipedia-on-ipfs.org/wiki/ 
> **Copy immutable permalink** https://bafy..foo.ipfs.dweb.link/wiki/
> Copy content path: /ipns/en.wikipedia-on-ipfs.org/wiki/ 
> **Copy immutable path**: /ipfs/bafy..foo/wiki/ 
> Copy CID: bafy..bar

Problems: 

- if we add one more item  the menu gets too tall and browser cuts the item at  bottom of the list.  Last time I checked there was no API that let us control height reliably across all vendors, so parked this PR until we rework the header of this menu to take less space.
- still, we simply have too many Copy.. items ([mock](https://user-images.githubusercontent.com/157609/62366891-69365b00-b528-11e9-9705-dc26e8f95126.png))

jessicaschilling:
They do fit into three categories:
- copy a link
- copy a path
- copy a CID

We could go down the path of making it a two-step process (copy a link: shareable or permalink?), which isn't ideal, but does reduce the initial cognitive burden down to three choices, which is a net improvement.

jessicaschilling:
@lidel Here's a thought. What if we just had a single "Copy ..." item that would launch a modal with those 3-5 options, and the user could click to copy the one they want? This keeps the menu tidy, and gives us a little room in the modal to qualify what each of those options actually means (particularly useful in situations where you as a newbie might not know if you'd rather copy the shareable link or the permalink).

The drawback is it knocks *every* copy action into a two-click process. However, that's also in line with how many (most?) "share" flows work in the general wild.

Thoughts?

lidel:
I think the idea of modal is ok, but still trying to find a way to make this work without introducing different screen/modal/huge state change :sweat_smile: 

What if we collapse all items into a single line with three buttons (or even icons?)

```
Enable on example.com
----------------------
Copy:  link, path, CID
----------------------
Pin IPFS resource
```

..and then when user hovers on "link" we show a value that will be copied when clicked (tooltip, or some sort of subtle overlay):

```
Copy:  [link], path, CID
 ________/\____________________________
| Click to copy a shareable link:
| https://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi.ipfs.dweb.link/
```

@jessicaschilling would appreciate a reality check with this idea.

jessicaschilling:
@lidel LOVE this idea. I'd go with words, rather than icons, to avoid another layer of decoding (not sure we can create meaningful icons to differentiate between link/path/CID.

So ... next steps? @DeedleFake -- I realize this started out way back as your PR 😊 Are you game for implementing these suggestions? Thanks so much!

DeedleFake:
I can take a stab at it when I get a chance, sure.

jessicaschilling:
@DeedleFake - thanks so much! Please reach out if we can be helpful. 😊

jessicaschilling:
Just a note: https://github.com/ipfs-shipyard/ipfs-companion/pull/907 may mean we don't need this. Watch this space.

jessicaschilling:
Quick note, cc @lidel and @DeedleFake: https://github.com/ipfs-shipyard/ipfs-companion/pull/907 may get us back the vertical height we need to make this work as originally proposed (i.e. adding two menu items for "Copy immutable permalink" and "copy immutable path"). Is it possible to look this back over again after the new Companion release?

DeedleFake:
Sure. Sorry that I haven't been able to get to it yet, by the way.

jessicaschilling:
No worries at all 😊 thanks!

jessicaschilling:
Hi @DeedleFake -- you can see the UI changes in stable channels for both Chrome and Firefox now! Are you able to take a look and see if there's room?

jessicaschilling:
Closing this in favor of https://github.com/ipfs-shipyard/ipfs-companion/pull/937, since some significant changes to Companion code since this PR was originally made.

@DeedleFake -- happy to have you pick this back up if you have the time! If not, we can keep it on our list, but may not be terribly high priority at the moment.

PR #769, opened 2019-09-28
feat: add 'Open Web UI' button to the welcome page

### LINKED ISSUE #763: Add "Open Web UI" to Welcome page
After ipfs-companion is installed we display the welcome page.

When active node is online we say "you are all set" and point at the list of features, but an actionable button to open Web UI would be much more useful:

Thoughts?

colinfruit:
Demo of the proposal in issue #763. Personally, I think that this feature should be included because it provides a quick and obvious way to open the web ui.

closes #763

### DIFF
--- a/add-on/src/landing-pages/welcome/page.js
+++ b/add-on/src/landing-pages/welcome/page.js
@@ -86,7 +86,7 @@ const renderWelcome = (i18n, peerCount, onOpenWebUi) => {
       <p class="${copyClass}">${renderTranslatedSpans('page_landingWelcome_welcome_peers', [peerCount], 'class="aqua fw6"')}</p>
       <p class="${copyClass} mb4">${renderTranslatedLinks('page_landingWelcome_welcome_discover', ['https://github.com/ipfs-shipyard/ipfs-companion#features'], `target="_blank" class="${anchorClass}"`)}</p>
       <div class="mt4 f5 flex justify-center items-center">
-        <button class="mw8 pv3 ph4 b navy br2 bn bg-white hover-bg-white-90 pointer" onclick=${onOpenWebUi}>${i18n.getMessage('panel_openWebui')}</button>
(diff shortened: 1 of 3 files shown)


lidel:
Thank you @colinfruit! :heart: 

Screen Shot 2019-10-01 at 11 17 51

Before we merge this let's get a quick UX check from @ericronne: 
anything we should tweak in the style of the button?

(this screen is shown when user installs ipfs-companion for the first time and already has ipfs-desktop running in the background)

ericronne:
For accessibility's sake, i'd go with a higher-contrast button/text combo, eg …

Artboard

lidel:
@colinfruit mind swapping the CSS to match Eric's suggestion?
I know you tried to match the style buttons from Web UI but we need to fix contrast there as well, so it is okay to use custom one here for now :)

autonome:
UX recommendation re language - "Web UI" doesn't have meaning for IPFS
users, only for core developers.

Can we change to something descriptive like "View and manage IPFS"? Or
something like that?

ericronne:
That's a little long for a button label.
`Open Dashboard`?

colinfruit:
@lidel, I'm happy to change the CSS to match @ericronne 's suggestions. One question, if 'Open Web UI' isn't descriptive enough for an ordinary user, shouldn't it be changed in both the browser-action drop-down and the welcome screen button? Just asking because it will change the way I go about this. I will just leave the other language's translations as they are, because I don't speak any of the available languages fluently. Is there a proper process handling translations, or anyone I should notify? I prefer 'Open Dashboard' for both, myself.

ericronne:
Yes this is about more than a button label; it’s about what that browser-based thing that’s a lot like IPFS Desktop is called. It’s a branding issue. As such, I recommend that we keep “Open Web UI” as the button text for now – to unblock this PR – and continue the (_important_) naming chat in a new issue.

colinfruit:
OK, 'Open Web UI' it is.

PR #780, opened 2019-10-02
feat: default local gateway URLs to  'localhost'

### LINKED ISSUE #328: Mixed Content Warning when using `localhost` Gateway
This one is kind of _Rube Goldberg machine_, so I'll start with a bit of background information.

_[Mixed Content](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content) warning_ happens when a page loaded over HTTPS contains a resource (eg. image) that  is loaded from insecure HTTP.  Browser vendors recognized that loading content over unencrypted HTTP from the box browser is running on is  a special case that provides similar security guarantees to HTTPS over the internet, so they whitelisted `127.0.0.1` and it no longer trigger the warning.

IPFS Companion defaults to using `localhost` instead of `127.0.0.1` since  #291 was closed. To be more specific, if user enters gateway URL that includes `127.0.0.1` it will be replaced with hostname `localhost`. 

The goal was to improve UX in Location Bar (`localhost` looks _much_ better than raw IP), but it came with unexpected side effect regarding [Mixed Content](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content) whitelisting implemented in Firefox and Chrome.  It seems that browsers whitelisted `127.0.0.1` but not `localhost`, which still produces Mixed Content warning.

In our use case, the **Mixed Content Warning is triggered when there is an IPFS resource on HTTPS page and ipfs-companion's redirect loads it from Local Gateway that has `localhost` in its URL**.

----

_(I used [pls.watch](https://pls.watch/) as a PoC in examples below,  because it is a public HTTPS site that enables users to trigger Mixed Content Warning by loading external resources from HTTP URLs)_

#### Firefox: the green padlock

> 2017-12-09-125633_713x31_scrot

#### Firefox: grey padlock with warning

> 2017-12-09-125548_691x29_scrot

> 2017-12-09-125858_1147x1421_scrot

> 2017-12-09-125707_655x60_scrot

#### Chrome: no padlock, warning under  `(i)` icon

> 2017-12-09-125837_1145x1419_scrot

---- 

For now, we need to decide what is more important:  

* keeping human-readable `localhost` in Location Bar at a cost of signaling that use of local IPFS gateway is somehow "not secure"

or 

* keeping "the green padlock" at a cost of going back to `127.0.0.1` in location bar.

Personally, I think "keeping the green  padlock"  is more important and we should go back and default to  `127.0.0.1`. Rationale: we can work with browser vendors to whitelist `localhost` and switch back to it at a later time, when it does not cause Mixed Content Warning.  

I am usually biased towards security  over UX, so would love to hear what others think before doing any changes.

----

References:

- https://bugzilla.mozilla.org/show_bug.cgi?id=903966 (Firefox whitelists `127.0.0.1` since v55)
- https://chromium.googlesource.com/chromium/src.git/+/130ee686fa00b617bfc001ceb3bb49782da2cb4e (Chrome whitelis `127.0.0.1` since v53)

lidel:
Closes #328

## TODO

- [x] default local gateway to `localhost`
- [ ] remove warning on options page
- [ ] ensure webui loaded from gateway port works in a firefox/chromium/brave

## Known Issues

- webui loaded from gateway port is broken
  - go-ipfs returns 403 Forbidden if request is made to `localhost` API
  - normalizing API host to IP solves the issue under Firefox and Chromium
  - Brave [continues to show CORS error](https://user-images.githubusercontent.com/157609/66035880-46b2c500-e50c-11e9-8aa2-4d0e211ee497.png), needs additional work

### DIFF
--- a/add-on/src/lib/options.js
+++ b/add-on/src/lib/options.js
@@ -73,19 +73,22 @@ exports.storeMissingOptions = async (read, defaults, storage) => {
   return changes
 }
 
-function normalizeGatewayURL (url) {
+function normalizeGatewayURL (url, { localhost = true }) {
   if (typeof url === 'string') {
     url = new URL(url)
   }
-  // https://github.com/ipfs/ipfs-companion/issues/328
-  if (url.hostname.toLowerCase() === 'localhost') {
+  // localhost normalization (https://github.com/ipfs-shipyard/ipfs-companion/issues/328)
+  if (localhost && url.hostname === '127.0.0.1') {
+    url.hostname = 'localhost'
+  }
+  if (!localhost && url.hostname === 'localhost') {
     url.hostname = '127.0.0.1'
   }
   // Return string without trailing slash
   return url.toString().replace(/\/$/, '')
 }
 exports.normalizeGatewayURL = normalizeGatewayURL
-exports.safeURL = (url) => new URL(normalizeGatewayURL(url))
+exports.safeURL = (url, opts) => new URL(normalizeGatewayURL(url, opts || {}))
 
 // convert JS array to multiline textarea
 function hostArrayCleanup (array) {
--- a/add-on/src/lib/state.js
+++ b/add-on/src/lib/state.js
@@ -19,7 +19,7 @@ function initState (options) {
   delete state.publicGatewayUrl
   state.redirect = options.useCustomGateway
   delete state.useCustomGateway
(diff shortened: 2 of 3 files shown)


lidel:
Superseded by https://github.com/ipfs-shipyard/ipfs-companion/pull/853

Issue #716, opened 2019-04-22
Embedded JS-IPFS in Brave


lidel:
> This issue tracks remaining work around embedded js-ipfs running in ipfs-companion  in Brave with access to `chrome.sockets` APIs. For more context, see:
> - Option (B) from notes on bringing IPFS to Brave in 2019 
    (https://github.com/brave/brave-browser/issues/819#issuecomment-456039555))
> - Support chrome.sockets.* APIs (https://github.com/ipfs-shipyard/ipfs-companion/issues/664)
> - Old issues: https://github.com/ipfs/in-web-browsers/issues/64, https://github.com/ipfs-shipyard/ipfs-companion/issues/312

## High Level Goals

- Leveling up js-ipfs to match go-ipfs in features relevant to in web browser uses
- Brave users can use IPFS with Companion alone, without installing go-ipfs or ipfs-desktop
- IPFS nodes in Brave can discover each other on the same network and exchange data in p2p fashion

## Kanban Board

- https://github.com/ipfs-shipyard/ipfs-companion/projects/4

## How to opt-in

Stable release of IPFS Companion can be enabled via Brave Settings (`chrome://settings/extensions`):

> ![2019-09-20--14-43-17](https://user-images.githubusercontent.com/157609/65327738-0458c200-dbb5-11e9-95d5-5393591388f6.png)

By default, it works like regular ipfs-companion and  `External` node is used by default, assuming user has [go-ipfs](https://github.com/ipfs/go-ipfs#go-ipfs) or [IPFS Desktop](https://github.com/ipfs-shipyard/ipfs-desktop#ipfs-desktop) installed and running.

Brave users can opt-in and try `Embedded + chrome.sockets` (see [/docs/node-types](https://github.com/ipfs-shipyard/ipfs-companion/blob/master/docs/node-types.md) for more details)

> ![2019-09-20--14-44-37](https://user-images.githubusercontent.com/157609/65327907-66192c00-dbb5-11e9-962f-9b5cd122572f.png)

> ![2019-09-20--14-45-06](https://user-images.githubusercontent.com/157609/65327962-7fba7380-dbb5-11e9-954b-6a18136eef60.png)

# TODO

## :art:  UX

* [ ] UI for changing Gateway port
  - Right now its just raw JSON with js-ipfs config
* [x] Redirect to embedded Gateway
* [x] New embedded node type exists (`embedded:chromesockets`)
[...]

PR #782, opened 2019-10-04
feat: precached webui works in offline mode

### LINKED ISSUE #682: Ensure blessed Web UI is pinned
### Context 

We switched to loading Web UI from IPFS in https://github.com/ipfs-shipyard/ipfs-companion/pull/680.

Blessed CID is hardcoded in extension sources:

https://github.com/ipfs-shipyard/ipfs-companion/blob/1b81a4fac937fff36d8b88f2e48eb4ede6cfb58a/add-on/src/lib/state.js#L27

### Problem

@Mikaela noted interesting  problem:

> Should the webui get automatically pinned or something? I have gc enabled on go-ipfs
> I am wondering what happens if I get disconnected for two hours or so, will I lose the webui?

### Solution

ipfs-companion should: 
- keep current `webuiCid` pinned
  - check could happen every time remote API is connected to for the first time in browsing session 
- when `webuiCid` changes, unpin old version and pin new one
  - relevant API: `browser.runtime.onInstalled` with [`OnInstalledReason`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/OnInstalledReason)

### LINKED ISSUE #765: Brave: Web UI should be available instantly
> Parent: #716  :lion: 

We open WebUI from embedded gateway but it is not in local cache after the install, and takes more than 30s to do the initial load.

We should see if we can bundle compressed webui with extension itself  and add it to local repo after initial install.

## Prior Art

- see what ipfs-desktop does 
- make sure we don't bundle webui in Firefox builds
  - or make WebUI reproducible: https://github.com/ipfs-shipyard/ipfs-webui/issues/959

lidel:
This PR adds precache logic for Web UI which is executed when IPFS API client (or full node) starts.
The main purpose is to make Web UI load instantly in Brave  and work in offline environments.

closes #765, closes #682,  part of #716 :lion: 

## How does it work?

During the build .tar archives are placed in `add-on/dist/precache`.
This makes them available to fetch over `*-extension://{extension-id}/dist/precache/*` URLs.
Right now we have only one archive: a 22MB file with release version of Web UI.

When IPFS client starts, preache logic is executed:

1. read Web UI CID from Companion's config
2. check if local repo contains mentioned CID.
   - if it is present, finish
   - if missing, continue to the next step
3. precache asynchronously:
   - fetch TAR archive from `*-extension://{extension-id}/dist/precache/*.tar`
   - pass entire directory tree as unpacked streams to ipfs.add
   - confirm produced CID is matching the one from step (1)

## Checklist

- [x] add TAR to precache dir in Chromium bundle
- [x] async import of TAR archives
- [x] build script extracts webui CID from [state.js#L9](https://github.com/ipfs-shipyard/ipfs-companion/blob/b47f81b34f36ba8c1e5977670f63bff6f0386d60/add-on/src/lib/state.js#L9)
  (keeps a single place to update instead of two)

## Known Issues

- [x] `/api/v0/add` does not work with latest js-ipfs (and we can't use older version due to gc-lock bug breaking precache process)
  - fixed in https://github.com/ipfs-shipyard/ipfs-companion/pull/794


add-on/src/lib/ipfs-client/embedded-chromesockets/config.js
```diff
@@ -25,7 +25,6 @@ const chromeDefaultOpts = {
       Swarm: [
         // optional ws-star signaling provides a backup for non-LAN peer discovery
         // (this will be removed when autorelay and DHT are stable in js-ipfs)
-        '/dns4/ws-star1.par.dwebops.pub.com/tcp/443/wss/p2p-websocket-star',
```
lidel: I believe this was just a dev alias for `ws-star.discovery.libp2p.io`

autonome:
🎉🌈🍧⛄️

PR #783, opened 2019-10-06
feat: recover from failed HTTP requests to third party gateways

### LINKED ISSUE #640: Recover from dead Public Gateways
> Extracted from #639

By default Companion will redirect IPFS paths to local gateway, so even if remote HTTP server is down, content will load just fine over IPFS.

What if user does not have locally running IPFS node tho?

We are able to detect HTTP errors in [WebExtensions/API/webRequest/onErrorOccurred](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onErrorOccurred): 
> Fired when a request could not be processed due to an error: for example, a lack of Internet connectivity.
> 
> The error is passed to the listener as the `error` property of the `details` object.
> 
> Note that this event is not fired for HTTP errors (4XX or 5XX responses): these will go through the normal stages of a request, calling any event listeners, and setting `details.statusCode` to report the error.

A rough idea would be to do what [we already do for websites with DNSLink](https://github.com/ipfs-shipyard/ipfs-companion/blob/v2.6.2/add-on/src/lib/ipfs-request.js#L268-L299):
- detect the tab failed to load IPFS-like path
- open failed path at a public gateway in a new tab

Caveats: 
- we are unable to do a HTTP redirect from `onErrorOccurred` so we recover in a new tab, which means this will only work for the main request (URL in location bar), won't fix images embedded from a gateway that went offline.

Prior art:
- Internet Archive's Wayback addon makes it possible to recover from HTTP 404: https://addons.mozilla.org/en-US/firefox/addon/wayback-machine_new/

colinfruit:
This PR closes #640, enabling dead public gateways to be redirected to the default public gateway. Added as an experimental option, defaulting to `false`.

### DIFF
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -417,7 +417,7 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
       if (!redirect && recoverable) {
         const redirectUrl = ipfsPathValidator.resolveToPublicUrl(request.url, state.pubGwURLString)
         redirect = { redirectUrl }
-        log(`onErrorOccurred: attempting to recover using public gateway for ${request.url}`, redirect)
+        log(`onErrorOccurred: attempting to recover failed request for ${request.url}`, redirect)
       }
       // We can't redirect in onErrorOccurred, so if DNSLink is present
       // recover by opening IPNS version in a new tab
@@ -437,7 +437,7 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
         const redirectUrl = ipfsPathValidator.resolveToPublicUrl(request.url, state.pubGwURLString)
         const redirect = { redirectUrl }
         if (redirect) {
-          log(`onErrorOccurred: attempting to recover using public gateway for ${request.url}`, redirect)
+          log(`onCompleted: attempting to recover failed request for ${request.url}`, redirect)
           createTabWithURL(redirect, browser)
         }
       }
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -413,8 +413,8 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
       }
       // if error cannot be recovered via DNSLink
       // direct the request to the public gateway
-      const recoverableViaPubGw = isRecoverableViaPubGw(request, state, ipfsPathValidator)
-      if (!redirect && recoverableViaPubGw) {
+      const recoverable = isRecoverable(request, state, ipfsPathValidator)
+      if (!redirect && recoverable) {
         const redirectUrl = ipfsPathValidator.resolveToPublicUrl(request.url, state.pubGwURLString)
         redirect = { redirectUrl }
         log(`onErrorOccurred: attempting to recover using public gateway for ${request.url}`, redirect)
@@ -430,10 +430,10 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
     async onCompleted (request) {
       const state = getState()
 
-      const recoverableViaPubGw =
-        isRecoverableViaPubGw(request, state, ipfsPathValidator) &&
+      const recoverable =
+        isRecoverable(request, state, ipfsPathValidator) &&
         recoverableErrorCodes.has(request.statusCode)
-      if (recoverableViaPubGw) {
+      if (recoverable) {
         const redirectUrl = ipfsPathValidator.resolveToPublicUrl(request.url, state.pubGwURLString)
         const redirect = { redirectUrl }
         if (redirect) {
@@ -550,7 +550,7 @@ function findHeaderIndex (name, headers) {
 
 // utility functions for handling redirects
 // from onErrorOccurred and onCompleted
-function isRecoverableViaPubGw (request, state, ipfsPathValidator) {
+function isRecoverable (request, state, ipfsPathValidator) {
   return state.recoverFailedHttpRequests &&
     ipfsPathValidator.publicIpfsOrIpnsResource(request.url) &&
     !request.url.startsWith(state.pubGwURLString) &&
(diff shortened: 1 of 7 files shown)


add-on/src/lib/ipfs-request.js
```diff
@@ -380,29 +400,48 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
...
-            active: true,
-            openerTabId: currentTabId,
-            url: dnslinkRedirect.redirectUrl
-          })
+        redirect = dnslinkResolver.dnslinkRedirect(request.url, cachedDnslink)
+        log(`onErrorOccurred: attempting to recover using dnslink for ${request.url}`, redirect)
+      }
+      // if error cannot be recovered via DNSLink
+      // direct the request to the public gateway
+      const recoverableViaPubGw = isRecoverableViaPubGw(request, state, ipfsPathValidator)
```
colinfruit: Including the redirect functionality was necessary in both onCompleted and onErrorOccurred, because of the different types of gateway failures. Many requests result in failure, others return 4xx or 5xx error codes.

lidel:
Hi @colinfruit is it still wip, or ok to review?

colinfruit:
Hi @lidel, the basic structure is fine to review. I'm working on tests, finding a clean way to stub promises (browser.tabs.query/create), and reviewing the error codes that were added, to make sure I have all that we want, and none that we don't want. Glad to hear your comments!

colinfruit:
To elaborate a little more on the testing part of my work, I've been going through a couple of different libraries to see if any of them with promisify sinon-chrome. The only two that could theoretically work are then-chrome and chrome-promise. chrome-promise is no longer maintained, just bug fixes, and then-chrome is outdated and seemingly abandoned. The only other option afaik is stubbing the needed APIs manually. Has this been discussed before for this project, and is there a better way to accomplish this?

lidel:
@colinfruit I stumbled upon similar problem with testing WebExtension APIs before. On top of the problem with abandoned libraries, sometimes Chromium and Firefox have slight changes in some APIs, making a generic library not feasible. In redirect tests I ended up writing bare minimim input-output tests for `modifyRequest.*` methods with hardcoded input specific to Chromium and Firefox, to ensure we have regression tests for each vendor  ([example](https://github.com/ipfs-shipyard/ipfs-companion/blob/40b7ab75353dd5140c82d6884d23ceabbcd48220/test/functional/lib/ipfs-request-gateway-redirect.test.js#L117-L126)).

I think we should be fine with adding `test/functional/lib/ipfs-request-gateway-recover.test.js` with similar tests for `modifyRequest.onCompleted` and `modifyRequest.onErrorOccurred` (modelled after [ipfs-request-gateway-redirect.test.js](https://github.com/ipfs-shipyard/ipfs-companion/blob/v2.8.6.840/test/functional/lib/ipfs-request-gateway-redirect.test.js)). If there are no vendor-specific APIs in `onCompleted` and `onErrorOccurred`, a single set of tests for both Firefox and Chrome will be enough.

Let me know if you got any questions.

colinfruit:
@lidel, that sounds good, except for the fact that I will have to rework the way the onErrorOccurred originally was written. The current setup doesn't return anything, so in my current test setup I am checking that browser.create was called with the expected arguments, `it('should redirect broken non-default public gateway IPFS request to public gateway', async function () {
      const request = url2request('https://nondefaultipfs.io/ipfs/QmYbZgeWE7y8HXkH8zbb8J9ddHQvp8LTqm6isL791eo14h', 500)
      await requestHandler.onCompleted(request)
      assert.ok(browser.tabs.create.withArgs({ url: 'https://ipfs.io/ipfs/QmYbZgeWE7y8HXkH8zbb8J9ddHQvp8LTqm6isL791eo14h', active: true, openerTabId: 12345 }).calledOnce, 'tabs.create should be called with IPFS default public gateway URL')
    })` I was under the impression that it would be bad practice to add a return value solely for the purposes of testing, but I'm happy to rework them if you're OK with this?

colinfruit:
@lidel just added my tests as a working example so it's easier to see what I am doing, will remove them if your previous comment is what you want. :)

test/functional/lib/ipfs-request-gateway-recover.test.js
```diff
@@ -0,0 +1,163 @@
...
+    browser.flush()
+  })
+})
+
+describe('requestHandler.onErrorOccurred:', function () {
+  let state, dnslinkResolver, ipfsPathValidator, requestHandler, runtime
+
+  before(function () {
+    global.URL = URL
+    browser.tabs = { ...browser.tabs, query: sinon.stub().resolves([{ id: 20 }]) }
```
colinfruit: Otherwise this call will break this [https://github.com/ipfs-shipyard/ipfs-companion/pull/783/files#diff-79fc41a4f320d6665f9e976b50731ba3R561](call to query).

add-on/src/lib/ipfs-request.js
```diff
@@ -380,29 +400,48 @@ function createRequestModifier (getState, dnslinkResolver, ipfsPathValidator, ru
...
+
+    async onCompleted (request) {
+      const state = getState()
+
+      const recoverableViaPubGw =
+        isRecoverableViaPubGw(request, state, ipfsPathValidator) &&
+        recoverableErrorCodes.has(request.statusCode)
+      if (recoverableViaPubGw) {
+        const redirectUrl = ipfsPathValidator.resolveToPublicUrl(request.url, state.pubGwURLString)
+        const redirect = { redirectUrl }
```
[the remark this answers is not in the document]
colinfruit: I feel like this is the cleanest way to do this. Without it we would have to worry about the [https://github.com/ipfs-shipyard/ipfs-companion/pull/783/files#diff-79fc41a4f320d6665f9e976b50731ba3R411](dnslink redirect), and do something like `const redirectUrl = redirect ? redirect.redirectUrl : null`

colinfruit:
thanks for the review, @lidel, resolved most of your suggestions. It would be cool to recover broken sub-domains. Maybe a separate issue to track this would be good? I'm glad to take a look at it if I have some time.

PR #802, opened 2019-10-22
feat: recover dead sub-domain gateways

### LINKED ISSUE #800: Recover from dead Subdomain Gateways
> Continuation of https://github.com/ipfs-shipyard/ipfs-companion/issues/640, and https://github.com/ipfs-shipyard/ipfs-companion/pull/797, cc https://github.com/ipfs/ipfs/issues/337, https://github.com/ipfs/in-web-browsers/issues/89

## Motivation

We now have a framework for recovering from failed HTTP responses to [Path Gateways](https://docs.ipfs.io/guides/guides/addressing/#path-gateway) and [DNSLink websites](https://docs.ipfs.io/guides/guides/addressing/#dnslink-gateway). To make it feature-complete we should do the same for [Subdomain Gateways](https://docs.ipfs.io/guides/guides/addressing/#subdomain-gateway).

## Details

```
https://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs.broken-example.com/wiki/
```

[Above URL](https://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs.broken-example.com/wiki/) fails at DNS lookup stage because `broken-example.com` domain name does not exist. This produces `NS_ERROR_UNKNOWN_HOST` on Firefox and `net::ERR_NAME_NOT_RESOLVED` on Chromium, which are already supported since #797.

We want to recover from this by redirecting failed subdomain-based gateways to `dweb.link` (our canonical subdomain gateway):
    `https://{cid}.ipfs.dweb.link/path/`

In the future, when we have support for `{cid}.ipfs.localhost` (https://github.com/ipfs/in-web-browsers/issues/109), we will also redirect to local node, but for now redirect to `dweb.link` should be enough.

cc @colinfruit

colinfruit:
closes #800

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -292,7 +292,7 @@
     "description": "An option description on the Preferences screen (option_publicGatewayUrl_description)"
   },
   "option_publicSubdomainGatewayUrl_title": {
-    "message": "Default Subdomain Public Gateway",
+    "message": "Default Public Subdomain Gateway",
     "description": "An option title on the Preferences screen (option_publicSubdomainGatewayUrl_title)"
   },
   "option_publicSubdomainGatewayUrl_description": {
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -291,6 +291,14 @@
     "message": "Fallback URL used when Custom Gateway is not available and for copying shareable links",
     "description": "An option description on the Preferences screen (option_publicGatewayUrl_description)"
   },
+  "option_subdomainPublicGatewayUrl_title": {
+    "message": "Default Subdomain Public Gateway",
+    "description": "An option title on the Preferences screen (option_subdomainPublicGatewayUrl_title)"
+  },
+  "option_subdomainPublicGatewayUrl_description": {
+    "message": "Default subdomain public gateway for recovery of broken subdomain gateways",
+    "description": "An option description on the Preferences screen (option_subdomainPublicGatewayUrl_description)"
+  },
   "option_header_api": {
     "message": "API",
     "description": "A section header on the Preferences screen (option_header_api)"
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -291,13 +291,13 @@
     "message": "Fallback URL used when Custom Gateway is not available and for copying shareable links",
     "description": "An option description on the Preferences screen (option_publicGatewayUrl_description)"
   },
-  "option_subdomainPublicGatewayUrl_title": {
+  "option_publicSubdomainGatewayUrl_title": {
     "message": "Default Subdomain Public Gateway",
-    "description": "An option title on the Preferences screen (option_subdomainPublicGatewayUrl_title)"
+    "description": "An option title on the Preferences screen (option_publicSubdomainGatewayUrl_title)"
   },
-  "option_subdomainPublicGatewayUrl_description": {
+  "option_publicSubdomainGatewayUrl_description": {
     "message": "Default subdomain public gateway for recovery of broken subdomain gateways",
-    "description": "An option description on the Preferences screen (option_subdomainPublicGatewayUrl_description)"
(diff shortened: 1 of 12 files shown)


add-on/src/lib/ipfs-request.js
```diff
@@ -527,8 +536,8 @@ function isRecoverable (request, state, ipfsPathValidator) {
   return state.recoverFailedHttpRequests &&
     request.type === 'main_frame' &&
     (recoverableNetworkErrors.has(request.error) || recoverableHttpError(request.statusCode)) &&
-    ipfsPathValidator.publicIpfsOrIpnsResource(request.url) &&
-    !request.url.startsWith(state.pubGwURLString)
+    (ipfsPathValidator.publicIpfsOrIpnsResource(request.url) || ipfsPathValidator.ipfsOrIpnsSubdomain(request.url)) &&
+    !request.url.startsWith(state.pubGwURLString) && !request.url.includes(state.subdomainGwURL.hostname)
```
lidel: We already have `ipfsPathValidator.publicIpfsOrIpnsResource`, let's move those custom subdomain checks to similar `publicIpfsOrIpnsSubdomainResource`. Will be easier to maintain and reuse.

add-on/src/lib/ipfs-path.js
```diff
@@ -38,6 +38,17 @@ function pathAtHttpGateway (path, gatewayUrl) {
 }
 exports.pathAtHttpGateway = pathAtHttpGateway
 
+function redirectSubdomainGateway (url, subdomainGateway) {
+  if (typeof url === 'string') {
+    url = new URL(url)
+  }
+  const fqdn = url.hostname.split('.')
+  const cid = fqdn[0]
+  const protocol = fqdn[1]
```
lidel: I realized this won't produce correct results for gateways under more complex top level domains such as `cid.ipfs.something.co.uk`.

The fix is to switch to `is-ipfs`, probably something like:

```js
const { subdomainPattern } = require('is-ipfs')

const match = fqdn.match(subdomainPattern)
if (!match) throw new Error('no match for isIPFS.subdomainPattern')
const protocol = match[1]
const cid = match[2]
```

We should avoid `split('.')` and use `subdomainPattern`  here and in already existing `subdomainToIpfsPath` method.

add-on/src/lib/ipfs-request.js
colinfruit: I'd love to, and left a todo above the `ipfsOrIpnsSubdomainResource` in the original PR. Right now we don't have a `subdomainGwUrl` state variable that allows for a test similar to the one we have in `publicIpfsOrIpnsSubdomainResource`, short of hard-coding one. Glad to hear if there's another way.

lidel: Ah yes, we don't have localhost subdomains yet (https://github.com/ipfs/in-web-browsers/issues/109).
You are right, better to add those methods when that happens.

add-on/_locales/en/messages.json
```diff
@@ -291,6 +291,14 @@
     "message": "Fallback URL used when Custom Gateway is not available and for copying shareable links",
     "description": "An option description on the Preferences screen (option_publicGatewayUrl_description)"
   },
+  "option_publicSubdomainGatewayUrl_title": {
+    "message": "Default Subdomain Public Gateway",
```
lidel:
```suggestion
    "message": "Default Public Subdomain Gateway",
```

add-on/_locales/en/messages.json
```diff
@@ -291,6 +291,14 @@
...
+    "description": "An option title on the Preferences screen (option_publicSubdomainGatewayUrl_title)"
+  },
+  "option_publicSubdomainGatewayUrl_description": {
+    "message": "Default subdomain public gateway for recovery of broken subdomain gateways",
```
lidel:
```suggestion
    "message": "Default public subdomain gateway for recovery of broken subdomain gateways",
```

PR #809, opened 2019-10-30
feat: web ui 2.6.0

hacdias:
See https://github.com/ipfs-shipyard/ipfs-webui/releases/tag/v2.6.0.


lidel:
@hacdias  I am afraid we can't merge this specific release: v2.6.0 causes a crash in Brave.
Please resubmit when https://github.com/ipfs-shipyard/ipfs-webui/issues/1287 is fixed in next release.

PR #810, opened 2019-11-07
feat: import files to MFS

### LINKED ISSUE #415: Save all uploads to MFS (via ipfs.files API)
> Part of UX improvements for sharing local files (#342)

### Current behaviour

Right now upload via right-click context menu or browser action menu item adds raw data to IPFS and it is up to user to memorize or save CID. 

### Feature Description

Would be great to let user manage shared files via WebUI's Files screen.

Given discussion in https://github.com/ipfs/ipfs-gui/issues/10#issuecomment-373928862:

- when file is uploaded via browser extension it should be automatically added to  MFS, so that user can later go back to it or decide to "unpin" it by removing it from MFS. 
  - Maybe something like `/ipfs-companion-uploads/YYYY-MM-YY_HHMMSS`?
  - OR, **show a checkbox below** (checked by default) saying  "Add to My Files at `/some/path/`" where the default path from Preferences can be chaged inline with super fast autocomplete   

### References

- pin labels https://github.com/ipfs/go-ipfs/issues/4586
- Design: Pinning Content https://github.com/ipfs/ipfs-gui/issues/10#issuecomment-373928862
- Moving away from current pin system  https://github.com/ipfs/go-ipfs/issues/4763
- Context actions should add to MFS https://github.com/ipfs-shipyard/ipfs-companion/issues/635

### LINKED ISSUE #635: Context action to "Add to MFS"
> &lt;@whyrusleeping&gt; lidel: i would absolutely love a right click 'save to mfs' thing for ipfs companion
> i can haz? 

Cool idea! 
We've actually are been thinking about switching Upload and all "Add to IPFS" context actions to MFS.

Would appreciate some feedback:

- Should we still pin, or is presence in MFS is enough?
  (I'd say to not pin: MSF is enough, and if we remove file from MFS it can be GC'd)
- Where should we save them in MFS? I was bit worried about adding stuff to the root (`/`) as it may get noisy quite fast. Perhaps `/added-via-ipfs-companion/`? (looking for better name) 
-  Does it make sense to keep non-MFS versions in UI? My thinking is that we don't want duplicate entries, as they would clutter context menus.

colinfruit:
This PR changes the upload target from `ipfs add` to MFS. Currently, it uses the default directory `ipfs-companion-uploads` with the user option to change this on the upload page. On successful upload, it will open the Web UI, at the directory used for the upload. 

closes #415 and closes #635

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -39,10 +39,6 @@
     "message": "Share files via IPFS",
     "description": "A menu item in Browser Action pop-up (panel_quickUpload)"
   },
-  "quickUpload_options_openViaWebUI": {
-    "message": "Open imported files in Web UI",
-    "description": "Checkbox label on the share files page (quickUpload_options_openViaWebUI)"
-  },
   "panel_openWebui": {
     "message": "Open Web UI",
     "description": "A menu item in Browser Action pop-up (panel_openWebui)"
@@ -435,10 +431,6 @@
     "message": "MFS Import Directory",
     "description": "An option title on the Preferences screen (option_uploadDir_title)"
   },
-  "quickUpload_options_wrapWithDirectory": {
-    "message": "Wrap single files in a directory to preserve their filenames.",
-    "description": "Checkbox label on the share files page (quickUpload_options_wrapWithDirectory)"
-  },
   "option_uploadDir_description": {
     "message": "Customize the directory used for uploading files to MFS.",
     "description": "An option description on the Preferences screen (option_uploadDir_description)"
@@ -493,7 +485,15 @@
   },
   "quickUpload_options_uploadDir": {
     "message": "MFS directory to store imported files",
-    "description": "Textbox label on the share files page (quickUpload_options_wrapWithDirectory)"
+    "description": "Textbox label on the share files page (quickUpload_options_uploadDir)"
+  },
+  "quickUpload_options_wrapWithDirectory": {
+    "message": "Wrap single files in a directory to preserve their filenames.",
+    "description": "Checkbox label on the share files page (quickUpload_options_wrapWithDirectory)"
+  },
+  "quickUpload_options_openViaWebUI": {
+    "message": "Open imported files in Web UI",
+    "description": "Checkbox label on the share files page (quickUpload_options_openViaWebUI)"
   },
   "page_proxyAcl_title": {
     "message": "Manage Permissions",
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -215,6 +215,10 @@
     "message": "IPFS Node",
     "description": "A section header on the Preferences screen (option_header_nodeType)"
   },
+  "option_header_fileImport": {
+    "message": "File Import",
+    "description": "A section header on the Preferences screen (option_header_fileImport)"
+  },
   "option_ipfsNodeType_title": {
     "message": "IPFS Node Type",
     "description": "An option title on the Preferences screen (option_ipfsNodeType_title)"
@@ -432,7 +436,7 @@
     "description": "An option title on the Preferences screen (option_importDir_title)"
   },
   "option_importDir_description": {
-    "message": "Customize the directory used for uploading files to MFS.",
+    "message": "Customize the directory used for imported files.",
     "description": "An option description on the Preferences screen (option_importDir_description)"
   },
   "option_resetAllOptions_title": {
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -127,17 +127,13 @@
     "message": "This Page",
     "description": "An item in right-click context menu (contextMenu_parentPage)"
   },
-  "contextMenu_AddToIpfsKeepFilename": {
-    "message": "Add to IPFS (Keep Filename)",
-    "description": "An item in right-click context menu (contextMenu_AddToIpfsKeepFilename)"
+  "contextMenu_importToIpfs": {
+    "message": "Import to IPFS",
+    "description": "An item in right-click context menu (contextMenu_importToIpfs)"
   },
-  "contextMenu_AddToIpfsRawCid": {
-    "message": "Add to IPFS",
-    "description": "An item in right-click context menu (contextMenu_AddToIpfsRawCid)"
-  },
-  "contextMenu_AddToIpfsSelection": {
-    "message": "Add Selected Text to IPFS",
-    "description": "An item in right-click context menu (contextMenu_AddToIpfsSelection)"
+  "contextMenu_importToIpfsSelection": {
+    "message": "Import Selected Text to IPFS",
+    "description": "An item in right-click context menu (contextMenu_importToIpfsSelection)"
   },
   "notify_addonIssueTitle": {
     "message": "IPFS Add-on Issue",
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -476,7 +476,7 @@
     "description": "Button on the share files page (quickUpload_options_show)"
   },
   "quickUpload_options_uploadDir": {
-    "message": "MFS directory to store uploaded files",
+    "message": "MFS directory to store imported files",
     "description": "Textbox label on the share files page (quickUpload_options_wrapWithDirectory)"
   },
   "page_proxyAcl_title": {
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -416,11 +416,11 @@
     "description": "Link text for managing permissions"
   },
   "option_preloadAtPublicGateway_title": {
-    "message": "Preload Uploads",
+    "message": "Preload Imports",
     "description": "An option title on the Preferences screen (option_preloadAtPublicGateway_title)"
   },
   "option_preloadAtPublicGateway_description": {
-    "message": "Enables automatic preload of uploaded assets via asynchronous HTTP HEAD request to a Public Gateway",
+    "message": "Enables automatic preload of imported assets via asynchronous HTTP HEAD request to a Public Gateway",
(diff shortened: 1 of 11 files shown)


add-on/src/popup/quick-upload.js
```diff
@@ -61,34 +59,28 @@ async function processFiles (state, emitter, files) {
...
-    const wrapFlag = (state.wrapWithDirectory || streams.length > 1)
     const options = {
-      progress: (len) => progressHandler(len, totalSize, state, emitter),
-      wrapWithDirectory: wrapFlag,
-      pin: state.pinUpload
+      create: true,
+      parents: true
     }
-    let result
+    state.progress = `Uploading ${streams.length} files...`
```
lidel: Note to self: it looks like we are losing import progress report, but it never worked correctly over HTTP API (https://github.com/ipfs/js-ipfs-http-client/issues/842), so its ok to drop it for now.

add-on/src/lib/ipfs-companion.js
```diff
@@ -787,6 +793,10 @@ module.exports = async function init () {
       return uploadResultHandler
     },
 
+    get openWebUiAtDirectory () {
+      return openWebUiAtDirectory
+    },
```
lidel: Ugh, another realization: we can't open webui if embedded js-ipfs is used in non-Brave browser.
We need to check if `state.nodeType === 'embedded'` and open resource at public gateway instead.

lidel: Perhaps we could make this even more useful:
What if we add toggle "Open imported files in WebUI"?
Enabled by default, but when disabled it would open at Gateway (restoring old behavior).

add-on/src/popup/quick-upload.js
```diff
@@ -61,34 +59,28 @@ async function processFiles (state, emitter, files) {
...
-      // This is just an additional safety check, as in past combination
-      // of specific go-ipfs/js-ipfs-http-client versions
-      // produced silent errors in form of partial responses:
-      // https://github.com/ipfs-shipyard/ipfs-companion/issues/480
-      const partialResponse = result.length !== streams.length + (options.wrapWithDirectory ? 1 : 0)
-      if (partialResponse) {
-        throw new Error('Result of ipfs.add call is missing entries. This may be due to a bug in HTTP API similar to https://github.com/ipfs/go-ipfs/issues/5168')
-      }
-      await ipfsCompanion.uploadResultHandler({ result, openRootInNewTab: true })
+      const files = streams.map(stream => (ipfsCompanion.ipfs.files.write(`${uploadDir}${stream.path}`, stream.content, options)))
```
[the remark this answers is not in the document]
colinfruit: Thanks!

add-on/src/lib/ipfs-companion.js
colinfruit: Sounds good. It would be great in subsequent issue to find a way of handling opening the WebUI when using embedded nodes. Has this been looked at before?

colinfruit:
Thanks, @lidel!

> Please change labels and variable names from "upload" to "import"

Makes sense. I changed all the user-facing labels this PR introduces from "upload" to "import". I did not change the variable names, because it feels inconsistent with the existing code. References to `upload` are everywhere, even the file name is `quick-upload.js`. If you want to make an issue, I'll be glad to remove `upload` references throughout the code base.

> Would it be possible to expose the name of default directory for uploads as an editable setting on Preferences screen?

Sure. Also, I removed it from the upload screen, because it seemed redundant. On the other hand, it provides useful functionality there as well. What do you think?

> I am bit worried what happens to /ipfs-companion-imports/ when user uploads 10th, 100th file. 

I'm looking into the best way to handle user-generated date symbols (`%Y-%m-%d`), and the best way to demonstrate this functionality to the user. What if they removed them, and then forgot what the symbols were, or weren't sure what they meant to begin with? Maybe a result field on the Preferences screen (and upload screen if we return the functionality) that show what directory will be used? Or an instruction in the description?

lidel:
Thank you @colinfruit! I am travelling this week, but will review this as soon I have some uninterrupted time.

add-on/_locales/en/messages.json
```diff
@@ -427,6 +427,14 @@
     "message": "Customize which namespaces are logged to Browser Console. Changing this value will trigger extension restart.",
     "description": "An option description for the log level (option_logNamespaces_description)"
   },
+  "option_uploadDir_title": {
+    "message": "MFS Import Directory",
```
lidel: We try to avoid "MFS" in user interfaces as it is a technical term and regular users don't know what it means.
```suggestion
    "message": "File Import Directory",
```

add-on/_locales/en/messages.json
```diff
@@ -472,16 +480,20 @@
     "description": "Status label on the share files page (quickUpload_state_buffering)"
   },
   "quickUpload_options_show": {
-    "message": "upload options",
+    "message": "import options",
     "description": "Button on the share files page (quickUpload_options_show)"
   },
+  "quickUpload_options_uploadDir": {
+    "message": "MFS directory to store imported files",
```
lidel:
```suggestion
    "message": "Path to store imported files",
```

add-on/_locales/en/messages.json
```diff
@@ -472,16 +480,20 @@
...
+  },
   "quickUpload_options_wrapWithDirectory": {
     "message": "Wrap single files in a directory to preserve their filenames.",
     "description": "Checkbox label on the share files page (quickUpload_options_wrapWithDirectory)"
   },
-  "quickUpload_options_pinUpload": {
-    "message": "Pin files so they are retained when performing garbage collection on your IPFS repo.",
-    "description": "Checkbox label on the share files page (quickUpload_options_pinUpload)"
+  "quickUpload_options_openViaWebUI": {
+    "message": "Open imported files in Web UI",
```
lidel: "imported files" is already in `quickUpload_options_uploadDir`, so we can make this one shorter or need to rephrase it
```suggestion
    "message": "Open in Web UI",
```

add-on/src/options/forms/experiments-form.js
```diff
@@ -153,6 +154,26 @@ function experimentsForm ({
...
+                <p><a href="https://docs.ipfs.io/guides/concepts/mfs/" target="_blank">
+                  ${browser.i18n.getMessage('option_legend_readMore')}
+                </a></p>
+              </dd>
+            </dl>
+          </label>
+          <input
+            id="uploadDir"
+            type="text"
+            pattern="^\/(.*)"
```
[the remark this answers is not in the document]
colinfruit: I added this before import happens in `quick-upload.js`, to account for the potential case of someone forgetting the trailing slash on the import page. It handles both cases, and prevents the user from having to guess that a trailing slash is necessary. https://github.com/ipfs-shipyard/ipfs-companion/pull/810/files#diff-9984375a8301f7832fa0da0a5aafd803R75

colinfruit:
> Let's add "File Import" section to Options page and move related settings there

add-on/src/popup/quick-upload.js
```diff
@@ -81,20 +87,36 @@ async function processFiles (state, emitter, files) {
...
+    state.progress = 'Completed'
     emitter.emit('render')
-    console.log('Upload result', result)
+    console.log(`Successfully imported ${streams.length} files`)
+
+    // open web UI at proper directory
+    // unless and embedded node is in use (no access to web UI)
+    // in which case, open resource.
+    if (state.ipfsNodeType === 'embedded' || !state.openViaWebUI) {
+      await ipfsCompanion.uploadResultHandler({ result, openRootInNewTab: true })
```
colinfruit: Noticed this is not working as expected for me tonight. Maybe it's some weirdness with my connection; I'll take a look at this tomorrow.

colinfruit: Same issue this morning, importing multiple files and opening them in a context menu hangs. However, it is reproducible on current master, so I won't deal with it in this PR.

add-on/src/popup/quick-upload.js
```diff
@@ -134,6 +134,8 @@ function file2buffer (file) {
 
 function formatImportDirectory (path) {
   path = path.replace(/\/$|$/, '/')
+  path = path.replace(/(\/)\/+/g, '$1')
```
colinfruit: I added this just because it makes a nicer user experience. Files will appear in the Web UI with multiple slashes, despite being stored without them if multiple slashes are not removed on format.

lidel: Thanks! Do you mind adding unit tests for `formatImportDirectory`, so we don't break this during future refactors?

For now it can be `test/functional/lib/quick-upload.test.js` (similar to existing `test/functional/lib/ipfs-path.test.js`)

add-on/src/popup/quick-upload.js
```diff
@@ -61,17 +71,20 @@ async function processFiles (state, emitter, files) {
...
-    progressHandler(0, totalSize, state, emitter)
+    const streams = files2streams(files)
     emitter.emit('render')
-    const wrapFlag = (state.wrapWithDirectory || streams.length > 1)
+    const wrapFlag = streams.length > 1
     const options = {
-      progress: (len) => progressHandler(len, totalSize, state, emitter),
-      wrapWithDirectory: wrapFlag,
-      pin: state.pinUpload
+      wrapWithDirectory: wrapFlag
```
lidel: The `ipfs.add` command  adds a low level pin by default.
We use MFS instead of low level pins, so we need to disable  it here:


```suggestion
      wrapWithDirectory: wrapFlag,
      pin: false // we use MFS for implicit pinning instead
```

colinfruit:
@lidel, fine to merge this. Thanks!

lidel:
@colinfruit any way to contact you privately? mind emailing me at `lidel<at>lidel.org`?

PR #811, opened 2019-11-07
refactor: the async await extravaganza

### LINKED ISSUE #843: Migration to JS API with Async Await and Async Iterables
👋 hey, a new js-ipfs release is coming soon \o/

It's a big release with breaking API changes, please check the [release issue](https://github.com/ipfs/js-ipfs/issues/2656) for all the info you need. There's also a [migration guide](https://gist.github.com/alanshaw/04b2ddc35a6fff25c040c011ac6acf26) to help you to migrate your code to the new API and a [blog post](https://blog.ipfs.io/2020-02-01-async-await-refactor/) explaining the changes and motivations behind them.

Unfortunately, the tests for your application _failed_ when they were run with the latest release candidate (`0.41.0-rc.0`). Do not worry, this was very likely considering the magnitude of the changes! Please consult the release info, it should have everything you need to be able to upgrade, and if you're still having problems then please reply to this issue and we'll get you sorted ASAP!

Aside from that, have fun with this release, we're really excited about this one and the possibilities it brings to our users! ❤️

lidel:
This PR:

- switches to latest async iterator versions of js-ipfs and js-ipfs-http-client
  - Context why this is a big deal:  https://blog.ipfs.io/2020-02-01-async-await-refactor/
  - Closes #843 
- simplifies code related to libp2p in brave (switching to convention from https://github.com/ipfs/js-ipfs/pull/2591)

## TODO

- [x] switch to new way of customizing libp2p bundle in brave
- [x] switch companion codebase to new API
  - [x] quick import in Firefox
  - [x] quick import in Chromium
    - broken by additional serialization that was not present in Firefox
      refs: https://bugs.chromium.org/p/chromium/issues/detail?id=112163, potential fix via trampoline from https://stackoverflow.com/a/21101597/11518426, https://stackoverflow.com/a/24303394/11518426
    - There is a discrepancy between how objects are passed between a page and background page within the same extension in Firefox and Chromium.
    - It seems that `ReadableStream` produced by `file.stream()` is able to correctly pass the boundary between quick import page and the IPFS API object running in the background.
    -  Did some tests to see how big is the difference, to decide if its worth special-casing http client when external node is used, results below:
      - 4GB import test Firefox
        - `file.stream()` + bg-page http client → ~47s
        - `file.stream()` + same-page http client → ~47s
        - `file` (Blob) +  same-page http client → ~25s :green_heart: 
        - `file.stream()` + bg-page js-ipfs  → ~1m40s
      - 4GB import test Chromium
        - `file.stream()` + bg-page  http client → :broken_heart: always errors after ~15s
        - `file.stream()` +  same-page http client → :broken_heart: always errors after ~15s
        - `file` (Blob) +  same-page http client → ~25s :green_heart: 
        - `file.stream()` + bg-page js-ipfs  → ~1m30s
      - **tl;dr** `file` (Blob) +  same-page http client provide best performance for external node, while  `file.stream()` + bg-page js-ipfs make things work okay-ish when embedded js-ipfs is used (we don't expect people to import 4GB with embedded node so good for now)
  - [x] fix progress reporting
	- for now only works with embedded node, but as soon its fixed in http client it will start working there too:
      > 

  - [x] precache
  - [x] context actions
  - [x] import via context action -- seee https://github.com/ipfs/js-ipfs/issues/3195 first
  - [x] embedded js-ipfs in Chrome
  - [x] embedded js-ipfs in Brave
- [x] switch Brave to webrtc (and stardust?) 
  - [ ] ensure discovery works
  - [ ] ensure node starts if signaling server is down
  - [ ] ensure mdns bug from https://github.com/libp2p/js-libp2p-mdns/issues/95 is fixed
<del>- [ ] switch window.ipfs to new API (or fill an issue to do it in separate PR)</del> (window.ipfs has been disabled for a while)
  - honestly not worth it, its an experiment, should not block this


add-on/src/lib/ipfs-client/embedded-chromesockets/config.js
```diff
@@ -113,7 +124,36 @@ async function buildConfig (opts, log) {
...
+
+      libp2pOptions.modules.peerDiscovery = [
+        wrtcstar.discovery,
+        wsstar.discovery,
+        MulticastDNS,
+        Bootstrap
+      ]
+
+      log('initializing libp2p with libp2pOptions', libp2pOptions)
+      return new Libp2p(libp2pOptions)
```
lidel: If we ignore necessary ws-star weirdness and inline code, this is basically 5 lines instead of 200  :+1:

lidel:
@alanshaw i don't think https://github.com/ipfs/js-ipfs/pull/2591 landed in v0.40.0, it will land in next release, right?

alanshaw:
Yep!

lidel:
Need to switch to latest js-ipfs, so I am repurposing this PR for general async iterators refactor, context #843

lidel:
Alright, this is as ready as it will be.

There are no functional changes, just refactor that accounts for new APIs
 + small bugfixes that I was able to applie while doing this.

Merging to unblock #921 and https://github.com/ipfs-shipyard/ipfs-companion/issues/888

PR #827, opened 2019-12-09
feat: preload visited DNSLink URLs to local node

lidel:
> For the context see #826 and discussion in https://github.com/ipfs/docs/issues/405

With https://github.com/ipfs-shipyard/ipfs-companion/pull/826 the data of visited DNSLink websites was no longer getting into the cache of local IPFS node by default.

This PR fixes the data distribution problem with a background task queue that ensures every visited DNSLink asset is preloaded to the cache of local IPFS node, contributing to the swarm. If original HTTP server is down, IPFS Companion will recover (#640) in a new tab.

This preload of DNSLink data is enabled by default, but can be disabled with a toggle on Preferences screen:

>

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -287,6 +287,14 @@
     "message": "If global redirect is enabled, this will include DNSLink websites and redirect them to respective /ipns/{fqdn} paths at Custom Gateway",
     "description": "An option description on the Preferences screen (option_dnslinkRedirect_description)"
   },
+  "option_dnslinkDataPreload_title": {
+    "message": "Preload visited pages",
+    "description": "An option title on the Preferences screen (option_dnslinkDataPreload_title)"
+  },
+  "option_dnslinkDataPreload_description": {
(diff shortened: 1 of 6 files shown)


cwaring:
This looks great, a couple of questions:

1) Will companion only cache visited pages or in the case of a root CID does it recursively cache the entire website? (considering bandwidth and space required with the latter option)
2) What happens if a PWA loads from the local browser cache and displays the website while HTTP is down, will companion still attempt to recover from IPFS in a new tab that scenario?

lidel:
Ad 1: only visited pages (requested URLs)
Ad 2: If it loads from cache, it will work fine. Companion recovers by opening URL in a new tab only if the main request (from location bar) ends with network error or HTTP error.

cwaring:
With both those scenarios that makes this implementation ideal 👍

lidel:
Cool, let's merge this. 
This feature will be useful to users who decide to disable DNSLink redirect via Preferences.
(I will flip the default setting in separate PR, keeping this one on topic).

PR #829, opened 2019-12-10
refactor: replace upload with import terminology

### LINKED ISSUE #817: Language: replace "upload" with "import"
## Current state

References to "upload" are everywhere

## Proposed change 

Rename labels and variable names from "upload" to "import" 
  (eg. "import to IPFS", `/ipfs-companion-imports/` directory on MFS, etc).

## Motivation

"upload to IPFS" brings some confustion, users reuse mental model of the old web suggesting that data is pushed to a third party server, while in truth data is just _imported_ to local node and _optionally preloaded_ to external one.

"import" is a safe alternative: Filecoin uses `import` term for similar purpose, and `ipfs add` may be renamed to `ipfs import` in the future (see https://github.com/ipfs/interface-js-ipfs-core/pull/552).

colinfruit:
Following up on the changes made in #810. This simple change replaces references to `upload` with `import`. 

closes #817


docs/privacy-policy.md
```diff
@@ -29,7 +29,7 @@ None. We don’t collect your Personal Data, period.
 We do not collect, rent, store or sell your Personal Data to anyone. However
 because IPFS Companion is a web extension that provides access to the real-time,
 peer-to-peer IPFS Network (which is a public platform for which anyone may join
-and participate) the data that you add or upload to the IPFS Network using IPFS
```
colinfruit: Wanted to call this out because `upload` in the context of the privacy policy may have implications I don't know about, and didn't want it to be merged silently.

lidel: I really appreciate this, thank you!

Two asks:
- In this case it makes more sense to remove `add` and just keep `import` (it covers more ground)
- Also, there is a date labeled `Last Update` at the top of the file, please bump it to current date (helps when we copy & paste it to addon stores)

lidel:
Thanks!

PR #847, opened 2020-02-14
fix: default to https when recovering .eth 

### LINKED ISSUE #841: Redirecting .ETH/ -> .ETH.LINK should use HTTPS instead of HTTP
**Platform:**
Windows
Microsoft Edge / Chrome,

**Issue**
Navigating to `almonit.eth/` redirects to `http://almonit.eth.link/`

**Expected result**
It should redirect to `https://almonit.eth.link/`

Defaulting to HTTPS makes sense, as .eth.link supports HTPS, and Edge/chrome will start warning that `http` sites are unsafe.

lidel:
> Closes #841 – please see it for full context

This PR ensures that navigating to `almonit.eth/` or `http://almonit.eth/`  uses HTTPS URL `https://almonit.eth.link/`


lidel:
Note to self:  we will be able to recover to `http://almonit.eth.ipns.localhost:8080` after #853 lands.

nachomazzara:
I'm interested in this too. Any update? 

Will work for any https://**.eth.link site?

lidel:
@nachomazzara this will work for all sites, however subdomains will cause HTTPS certificate error, as noted in https://github.com/ipfs-shipyard/ipfs-companion/issues/841#issuecomment-586487470

nachomazzara:
> @nachomazzara this will work for all sites, however subdomains will cause HTTPS certificate error, as noted in [#841 (comment)](https://github.com/ipfs-shipyard/ipfs-companion/issues/841#issuecomment-586487470)

Where the fix should come from?

lidel:
- short answer: it is up to the operator of HTTP gateway such as https://eth.link to set up a proper certificate for each subdomain.
  - longer answer: only a single level of wildcards is supported in PKI, so additional orchestration is needed to request wildcard certificates for subdomains while they are requested for the first time. gateway operator needs to taken care of this + negotiate with a Certificate Authority if this behavior is allowed (some CA explicitly forbid "mining certificates on the fly" in their ToS).

PR #848, opened 2020-02-14
docs(README): user counts and latest browser logos

lidel:
@autonome @momack2 fyi I've found a way to display user counts in realtime, right in the README:

> 

I feel it is ok to do it, but want to double check with you.


momack2:
This is awesome!! Great work! Would *love* to have this for IPFS Desktop as well (but that's probably less doable)

Issue #667, opened 2019-01-23
Redirect of CID-in-subdomain and DNSLink websites


lidel:
> _Parent issues: https://github.com/ipfs/ipfs/issues/337, https://github.com/ipfs/in-web-browsers/issues/89, https://github.com/ipfs-shipyard/ipfs-companion/issues/527_

## Problem

There is an open problem of handling CID in Subdomain. CID-in-subdomain gives a website Origin-based isolation if loaded from public gateway, but we continue to have a single Origin for all websites if it is redirected and  loaded from local gateway (`/ipfs/<cid>`).

Current status:

- IPFS Companion extension DOES NOT redirect from `cidv1b32.ipfs.foo.tld`  to `http://127.0.0.1:8080/ipfs/cidv1b32` because that would decrease security guarantees ([example](http://bafybeifksbbtpipf2vwbyhxybkvvpnuus7efpz7nlr2s7pw2u3vwifax74.ipfs.dweb.link/en/))

- However, we DO redirect DNSLinked websites eg. https://docs.ipfs.io to http://127.0.0.1:8080/ipns/docs.ipfs.io/ because at the time nobody relied on Origin-based guarantees on IPFS websites.

I feel we should come up with a solution that is addressing both use cases, as having different behavior is confusing.

## Short Term Fix

In the short term we have two options: either redirect or not.

**( A ) :anger:  Redirect** to `http://127.0.0.1:8080/ip(f|n)s/..` even IF it would decrease Origin-based guarantees
  - This assumes that website owner set it up knowing about the lack of Origin-based isolation at the local gateway (detecting Origin change and disabling sensitive features). I am not convinced this is a realistic assumption, and would like to go away from it for DNSLink websites.
  - Potential mitigation would be to add redirect opt-out via additional DNS TXT record, but it overcomplicates DNSLink even further
 
**( B )** :angel:    (@lidel's plan)  **DO NOT redirect** to `http://127.0.0.1:8080/ip(f|n)s/..` by default, but give opt-in option to open at local gateway
  - [display IPFS page actions](https://user-images.githubusercontent.com/157609/51607505-1ba3ba80-1f15-11e9-92df-ddf642723267.png) and add an entry there to either:
[...]

PR #853, opened 2020-03-22
feat: *.localhost subdomain gateway support with http proxy

### LINKED ISSUE #328: Mixed Content Warning when using `localhost` Gateway

### LINKED ISSUE #736: Use Web UI version from IPFS Node

lidel:
This PR adds support for subdomain gateways introduced in go-ipfs v0.5.0-dev (https://github.com/ipfs/go-ipfs/pull/6096), specifically, one running at `*.localhost` subdomains.

### Um.. what? 

**TL;DR** we will change the URL present in the address bar when user loads content from the local node (ipfs-desktop or standalone go-ipfs):

```
http://127.0.0.1:8080/ipns/en.wikipedia-on-ipfs.org/wiki/
```

to:

```
http://en.wikipedia-on-ipfs.org.ipns.localhost:8080/ipns/wiki/
```

#### Why?

- **localhost subdomain gateway provides unique Origin per content root**
  - this is extremely important, see
    - historical discussion about URI and Origin
      -  https://github.com/ipfs/in-web-browsers/issues/6
      - https://github.com/ipfs/in-web-browsers/issues?q=label%3AOrigin
    - in-depth analysis of how to unify the way we do redirects without compromising website isolation  https://github.com/ipfs-shipyard/ipfs-companion/issues/667
- better UX on DNSLink websites
- DNSLink websites get proper Origin isolation per content root (https://github.com/ipfs-shipyard/ipfs-companion/issues/667)
- we can enable redirect of subdomain gateways  (https://github.com/ipfs/in-web-browsers/issues/89, https://github.com/ipfs-shipyard/ipfs-companion/issues/667)

### How?

> Disclaimer: below is a draft of what may end up a blogpost after go-ipfs 0.5 ships. WIP. 

They key challenge was to ensure *.localhost DNS names resolve to 127.0.0.1 on all platforms, specifically Linux distribution which ship with very basic DNS resolver that returns "not found" when trying to resolve localhost subdomains. 

Browsers are moving in the direction of hardcoding `localhost` names to loopback interface, but we are not there yet. We need to solve this problem with tools at hand.

We do that by setting up HTTP Gateway port of local go-ipfs to act as HTTP Proxy. This removes DNS lookup step from the browser, and go-ipfs ships with implicit support for subdomain gateway when request comes with "Host: <cid>.ipfs.localhost:8080" or
similar. 

We register HTTP proxy using Firefox and Chromium-specific APIs, but the  end result is the same. When enables, default gateway uses 'localhost'  hostname (subdomain gateway) instead of '127.0.0.1' (path gateway)  and every path-pased request gets redirected to subdomain by go-ipfs itself, which decreases complexity on browser extension side.

#### TODO

This is work in progress:

- [x] support enabling/disabling HTTP proxy
  - [x] Firefox
  - [x] Chromium
- [x] make `localhost` the default
- [x] update old configs from `127.0.0.1` to `localhost`
- [x] fix context actions on *.localhost
  - [x] release `is-ipfs` with  `isIPFS.dnslinkSubdomain(url)`: https://github.com/ipfs/is-ipfs/pull/32
- [x] redirect from `127.0.0.1` to `localhost`  when proxy is enabled (unify user experience)
- [x] enable redirect of subdomain gateways (like`*.dweb.link`) to `*.locahost`
- [x] fix toggle per site on `<fqdn>.ipns.localhost`
- [x] fix _Copy Public Gateway URL_ on `<fqdn>.ipns.localhost`
- [x] ship a beta (before merging this PR)
- [x] confirm default settings work fine with go-ipfs 0.4.23
- [x] confirm default settings work fine with go-ipfs 0.5.0-dev
- [x] confirm default settings work fine in Brave
- [x] fix / add tests
  - [x] https://ipfs.io/ipfs/QmbcBPAwCDxRMB1Qe7CRQmxdrTSkxKwM9y6rZw2FjGtbsb/%3Ffilename=test.jpg can be opened from local gw
  - [x] do not redirect non-main-frame requests to local subdomain, so embedding images works in firefox
  - [x] do not proxy all `localhost`  ports, only the one of current gateway
  - [x] state.enabled = false should disable proxy

### caveats 

- [match pattern](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns) used for registering proxy handler in Firefox uses `localhost` name without port. This is by design, match patterns do not seem to support ports and are executed on all of them. We need to check port inside of handler, and return `direct` proxy route if it does not match
- proxy situation is temporary and will change over time
  - Chromium does not need proxy, they hardcode `localhost` and `*.localhost` to loopback IPs
  - Firefox needs proxy until https://bugzilla.mozilla.org/show_bug.cgi?id=1220810 lands

### Related

- Migration to CIDv1 (default base32) https://github.com/ipfs/ipfs/issues/337
- CID as a subdomain https://github.com/ipfs/in-web-browsers/issues/89
- Redirect of CID-in-subdomain and DNSLink websites: https://github.com/ipfs-shipyard/ipfs-companion/issues/667
- Subdomain Gateway https://github.com/ipfs/go-ipfs/issues/6498
  - Act as HTTP PROXY for `http://<cidv1b32>.ipfs.localhost`: https://github.com/ipfs/go-ipfs/issues/5982
  - PR: subdomain gateway via http proxy in go-ipfs: https://github.com/ipfs/go-ipfs/pull/6096
- Closes #328
- Closes #736


lidel:
Surfacing some notes on additional fixes that had to be applies to keep subresource redirects working on HTTPS websites:

> Firefox 74 does not mark 	`*.localhost subdomains`	 as Secure Context yet
> (https://bugzilla.mozilla.org/show_bug.cgi?id=1220810#c23) so we can't redirect
> there when we have IPFS resource embedded on HTTPS page (eg.  image loaded from
> a public gateway) because that would cause mixed-content warning and
> subresource would fail to load.  Given the fact that `localhost/ipfs/*` provided
> by go-ipfs 0.5+ returns a redirect to `*.ipfs.localhost` subdomain we need to
> check requests for subresources, and manually replace `localhost` hostname with
> `127.0.0.1` (IP is hardcoded as Secure Context in Firefox). The need for this
> workaround can be revisited when Firefox closes mentioned bug.
> 
> Chromium 80 seems to force HTTPS in the final URL (after all redirects) so
> `https://*.localhost` fails. This needs additional research (could be a bug in
> Chromium). For now we reuse the same workaround as Firefox.
> 
> To unify use of `127.0.0.1` and `localhost` in address bar (eg. when user opens an
> image in a new tab etc) when Subdomain Proxy is enabled we normalize address
> bar requests made to the local gateway and replace raw IP with 'localhost'
> hostname to take advantage of subdomain redirect provided by go-ipfs >= 0.5

----

In other news:

- Added a bunch of tests, all green now. 
- In the meantime I noticed Chrome does not require proxy mode because they already hardcode `localhost` to raw IP. I am considering removing `proxy` permission from Chromium package as it may trigger yet another manual review at Chrome Web Store. 
  - Will keep the code commented out in case we need to enable it in the future for some reason.

add-on/_locales/en/messages.json
```diff
@@ -279,6 +279,14 @@
...
     "description": "An option description on the Preferences screen (option_useCustomGateway_description)"
   },
+  "option_useSubdomains_title": {
+    "message": "Use Subdomains",
+    "description": "An option title on the Preferences screen (option_useSubdomains_title)"
+  },
+  "option_useSubdomains_description": {
+    "message": "Isolate content roots from each other by loading them from subdomains at *.localhost and creating a unique Origin for each CID, IPNS or DNSLink record. Requires a local go-ipfs 0.5.0 or later.",
+    "description": "An option description on the Preferences screen (option_useSubdomains_description)"
+  },
```
lidel, line 289: On _Preferences_ screen the above labels look like this:



@jessicaschilling @autonome I suspect this could be phrased in a more human way, but this is the best I could do for next Beta release.  Suggestions welcome before we ship to Stable.

This is a description of a toggle that lets people to switch to old path-based gateway on `127.0.0.1` if they choose to do so for some reason. (disabling it stops HTTP proxy and swaps Custom Gateway URL to use raw IP)

lidel:
Shipped this in a new Beta: [v2.10.0.902](https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.10.0.902).  Will play with it during the weekend.

I do not expect big changes apart from tweaking labels mentioned in previous comment.

test/functional/lib/ipfs-request-workarounds.test.js
```diff
@@ -146,65 +148,23 @@ describe('modifyRequest processing', function () {
...
-      const response = Object.assign({}, request)
-      delete response.requestHeaders
-      response.responseHeaders = []
-      const responseHeaders = modifyRequest.onHeadersReceived(response).responseHeaders
-      const corsHeader = { name: 'Access-Control-Allow-Origin', value: state.gwURL.origin }
-      const acahHeader = { name: 'Access-Control-Allow-Headers', value: acrhHeader.value } // expect value restored from preflight
-      expect(responseHeaders).to.deep.include(corsHeader)
-      expect(responseHeaders).to.deep.include(acahHeader)
-
+      expect(modifyRequest.onBeforeSendHeaders(request).requestHeaders).to.include(nullOriginHeader)
```
lidel, line 167: I've added test to ensure Companion does not remove `Origin: null`

test/functional/lib/ipfs-request-workarounds.test.js
```diff
@@ -146,65 +148,23 @@ describe('modifyRequest processing', function () {
   })
 
   describe('a request to <apiURL>/api/v0/ with Origin=null', function () {
-    it('should remove Origin header ', async function () {
-      // set vendor-specific Origin for WebExtension context
+    it('should keep the "Origin: null" header ', async function () {
+      // Presence of Origin header is important as it protects API from XSS via sandboxed iframe
+      // NOTE: Chromium <72 was setting this header in requests sent by browser extension,
+      // but they fixed it since then.
```
lidel, line 154: Removed unused code related to loading webui from gateway port without setting CORS and added test to ensure Companion does not remove `Origin: null` 

cc https://github.com/ipfs/go-ipfs-cmds/pull/190

lidel:
After feedback from two beta releases I feel this is stable enough to merge to master and ship beta+stable. 

We need this not only for subdomains, but  also to ensure POST-only API (https://github.com/ipfs/go-ipfs/pull/7097) scheduled for go-ipfs 0.5 does break DNSLink lookup and ipfs-webui.

lidel:
Released in Beta: https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.11.0.904
Scheduled to ship to stable as v2.11.0

PR #858, opened 2020-04-02
docs: update SECURITY.md

lidel:
`SECURITY.md` was ancient:

- updated it to make more sense in Firefox+Chromium context
- added `Reporting security issues` section
  - <del>for now I've put my email there, but perhaps we should</del> 
    used sec process from https://github.com/ipfs/ipfs#a-word-on-security 

cc @autonome @jessicaschilling

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,19 +1,23 @@
-# Security Considerations
+# Security notes
 
-The Firefox platform has no mechanisms to restrict the privileges of add-ons
-and add-on code is fully trusted by Firefox.
+## Reporting security issues
 
-Security-conscious users should confirm that downloaded XPI package does match source code BEFORE installation.
+If you found a security issue, please email `lidel [-at-] protocol.ai` ([gpg](https://keybase.io/lidel))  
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -1,19 +1,23 @@
...
+
+## Local build and source verification
+
+Security-conscious users can confirm that downloaded package does match source code.
 
 Required steps:
 
-1. Download XPI package in version that is to be verified
+1. Download package version that is to be verified
 2. Checkout sources of the [same tag](https://github.com/ipfs/ipfs-companion/tags)
```
jessicaschilling, line 15:
```suggestion
2. Check out sources of the [same tag](https://github.com/ipfs/ipfs-companion/tags)
```

lidel:
Thanks! I am ok with being the entry point, but that comes with risky bus factor, so agree we should avoid it.

I replaced my email with the one from
https://github.com/ipfs/ipfs/blob/master/README.md#a-word-on-security

I'll make a smoke-test of `security@ipfs.io` just to confirm process is working before merging this PR.

PR #862, opened 2020-04-05
style: rename menu items for copying and sharing

### LINKED ISSUE #857: Rename "Copy Public Gateway URL"
### TL;DR

When user wants to share link, they usually copy URL from address bar.
This brings UX issue when resource was loaded from local IPFS node running on user's machine.

### Problem

Sharing `http://` link to the canonical gateway and then upgrading it to IPFS if User Agent supports it would be ideal, but we don't have API for manipulating the URL copied from address bar and probably won't see one due to the potential abuse such API could bring.

### Existing solution 

The best we can do right now is to provide menu item for copying a shareable link to the IPFS resource at a public gateway.

The goal is to make it easier to share links to IPFS resources, however I don't believe the currently used label is descriptive enough:
https://github.com/ipfs-shipyard/ipfs-companion/blob/7575d09ee68d0d22f5da70711a267e8f8281fca5/add-on/_locales/en/messages.json#L91

>  
>
>  

### How to improve it?

Multiple folks seemed to be surprised this exists.  We will redesign Companions UI in the future, but perhaps we could make a small  change  sooner?

I imagine changing label + adding some icon could help in discovery.

Would something like `Copy link for sharing` be better?

Other suggestions are welcome. Please comment below.

lidel:
Closes #857

- `Copy Public Gateway URL` → `Copy shareable link`
- `Copy IPFS Path` → `Copy IPFS content path`

>

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -72,7 +72,7 @@
     "description": "A menu item tooltip in Browser Action pop-up (panel_activeTabSiteIntegrationsToggleTooltip)"
   },
   "panel_pinCurrentIpfsAddress": {
-    "message": "Pin IPFS Resource",
+    "message": "Pin IPFS resource",
     "description": "A menu item in Browser Action pop-up (panel_pinCurrentIpfsAddress)"
   },
   "panel_pinCurrentIpfsAddressTooltip": {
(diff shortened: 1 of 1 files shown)


jessicaschilling:
Would “Copy IPFS content path” be more explicit? There’s room for it.

lidel:
@jessicaschilling updated: 

>

jessicaschilling:
Nice! Oh: “Resource” in the pinning menu item should be lowercase for consistency. Otherwise LGTM 😊

Issue #861, opened 2020-04-04
import to IPFS is always disabled


bluelovers:
**Describe the bug**

import to IPFS is always disabled

![image](https://user-images.githubusercontent.com/167966/78463829-5bb71d80-7714-11ea-9253-050cd65b6a6f.png)

**To Reproduce**

**Expected behavior**

allow import

**Screenshots**

![image](https://user-images.githubusercontent.com/167966/78463703-b8193d80-7712-11ea-8606-704829f22104.png)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] win 10
 - Browser [e.g. chrome, safari] chrome

![image](https://user-images.githubusercontent.com/167966/78463729-f44c9e00-7712-11ea-8227-9400d0c26cef.png)

**Smartphone (please complete the following information):**

**Additional context**

lidel:
@bluelovers do you have IPFS running? what the version of go-ipfs?

For import to work, you need to have extension enabled and a local IPFS node running:

![2020-04-05--19-18-43](https://user-images.githubusercontent.com/157609/78505223-483a9f80-7772-11ea-8b70-71bfad91f83c.png)

bluelovers:
go-ipfs v0.4.23

![image](https://user-images.githubusercontent.com/167966/78505486-51456400-77a6-11ea-918a-1858cc61ce62.png)

![image](https://user-images.githubusercontent.com/167966/78505492-5efae980-77a6-11ea-941c-80260b780d51.png)

use js-ipfs and update server conf in ipfs-companion

`import to IPFS` is also disabled

lidel:
I see you have `0` peers – that is why it is disabled:

https://github.com/ipfs-shipyard/ipfs-companion/blob/df96a055157a8e715d141ea5bb1d03beb3b58acc/add-on/src/lib/context-menus.js#L166-L168

I believe this is a bug similar to https://github.com/ipfs-shipyard/ipfs-companion/issues/790. 
We should allow imports when there are no peers, as long API is online.

lidel:
I believe the fix is to update mentioned line  to:

```diff
- const ifApi = getState().peerCount > 0 
+ const ifApi = getState().peerCount >= 0 
```

@bluelovers is this something you would be interested in contributing as a PR?

bluelovers:
hmm..., how can u have that many peers ?

PR #863, opened 2020-04-05
fix: import to ipfs is disabled

bluelovers:
https://github.com/ipfs-shipyard/ipfs-companion/issues/861

code by @lidel


jessicaschilling:
@lidel, sending this your way 😊

PR #871, opened 2020-04-24
feat: make Preferences link to docs site

jessicaschilling:
- Moves `privacy-policy.md` and `examples` directory to top level; fixes links to privacy policy that result from the move
- Removes `docs` directory; all the remaining content has been moved to docs beta per https://github.com/ipfs/ipfs-docs-v2/pull/182
- Updates all "Read more" links in Companion prefs to point to the appropriate pages on docs-beta.ipfs.io
   - NOTE: These URLs need to be changed once the docs legacy site is deprecated! But, the [deprecation plan](https://github.com/ipfs/docs/issues/475) does include searching all IPFS codebases for instances of "docs-beta" and fixing. Plus redirects.

Closes #595 BUT shouldn't be merged until https://github.com/ipfs/ipfs-docs-v2/pull/182 is merged and docs are in place on the docs site.

**One additional note:** The "Load websites from custom gateway" option still points to https://github.com/ipfs-shipyard/ipfs-companion/issues/667 for its "Learn more" link. @lidel, are you up for writing a brief explainer to put in the docs once that issue is closed (presumably 0.5 release)?

PR #879, opened 2020-05-08
feat: open Preferences in a new tab

### LINKED ISSUE #502: Open preferences in new tab for vendor consistency
_(This may be a prerequisite for #491)_

## Background

Preferences screen looks different in every browser:

- Firefox opens `about:addons` page with vendor-specific settings  and [embeds Preferences inline on the bottom of the page](https://user-images.githubusercontent.com/157609/41788493-010a931e-764c-11e8-9651-0938a9103127.png)
- Chrome opens it in [a bizarre overlay that is too small for advanced UI](https://user-images.githubusercontent.com/157609/41789646-554ffc9e-7650-11e8-83b6-6ea5fa68d5c0.png)
- Firefox for Android opens it in a new tab

:carousel_horse: 

## Change

Always open _Preferences_ in a new tab: 
- a single UI for all vendors, no need for Chrome-specific styles
- bigger canvas enable us to design intuitive interfaces for advanced features as multiple backends (#491)
- we can finally switch _Preferences_ screen to use [ipfs-css](https://github.com/ipfs-shipyard/ipfs-css)

lidel:
> Closes #502

This PR opens _Preferences_ in a new tab, which makes it look the same in every browser.

### TODO

- [x] update manifest
- [x] remove vendor-specific quirks and hacks
- [x] improve styling of `add-on/src/options/options.css|html`
- [x] test styling at various responsive widths for both Chrome and Firefox

--- 

@jessicaschilling as you suggested in https://github.com/ipfs-shipyard/ipfs-companion/issues/502#issuecomment-622204584  I've made the backend changes, it works, but needs style work to make it look decent

Shortcuts to build & preview:
- `yarn dev-build firefox`
- `yarn dev-build chromium`


jessicaschilling:
@lidel Thanks! Do you mind reviewing the updated links to documentation so we can get those merged in before I start on this?

lidel:
@jessicaschilling I've rebased this PR on top of #871 – should be ok  to work on now :+1:

jessicaschilling:
@lidel - FYI, it's getting there in the most recent commit I pushed, but still want to do some more tweaking. It's looking a lot like Web UI prefs now.

lidel:
Nice! It is already much better!

Small nit: seems color used for `.link` changed, and that impacts other screens (eg. [welcome page](https://user-images.githubusercontent.com/157609/81815704-989fdc80-952a-11ea-981a-105de9e1e30d.png)).

jessicaschilling:
@lidel - thanks for the reminder on the welcome screen. There's a few things I'd like to tweak there anyway.

lidel:
## On preferences

I think its already much better than it was and we should merge, but the background color of inputs is bit aggressive: 

> 

Perhaps we can use something more subtle? [old version](https://user-images.githubusercontent.com/157609/81970564-0548d300-9620-11ea-853f-d42147e34269.png) used white background and border, here we probably could also use white and color accent on the bottom, perhaps?

### On welcome screen

Hm... now those colorful links distract from the content / feel noisy:

> 

Could we move in the other direction, and make links have the same color as text, but make the underline in this color? 

For the reference, current `master` with underline in the same color as text:

>

jessicaschilling:
@lidel - made some tweaks as we discussed on Slack. I think this might be ready to go. Had a look around in both FF and Chrome.

I'd like to do something more interesting with the header - like bring the heartbeat cube back in if Companion is active - but that's something for later.

Thanks!

PR #883, opened 2020-05-18
feat: responsive prefs page

### LINKED ISSUE #880: Improve Companion prefs page responsive behavior
As of now, the Companion preferences page works on narrow-width/mobile screens, but it could use some optimization:
- Scale down header text, scale up logo to match each other in importance
- Stack toggles/dropdowns/entry fields vertically against descriptions, rather than side-by-side

This would also be a good time to introduce any visual tweaks we may want to make to the header just to make it more delightful:
- Maybe bring the "heartbeat" cube into this page's header, depending on if the global toggle is on?

jessicaschilling:
- Clones header behavior from IPLD Explorer
- Stacks toggles/dropdowns/entry fields vertically against labels at mobile widths (Tachyons class suffix `-ns`)

Refrained from animating the cube when global toggle on in order to be consistent with other IPFS headers across the ecosystem. (Ah well.)

Closes #880


jessicaschilling:
@lidel -- improvements to the Companion prefs page, if you don't mind taking a look. Hopefully not too late to get into the next release? Thanks!

PR #885, opened 2020-05-20
fix: include CSS assets in the install package

lidel:
Turns out `web-ext` ignores everything under `node_modules` no matter how deep in the project tree it is:

https://github.com/mozilla/web-ext/blob/065188db35a4f63c2d17ff744cc1de2601a62222/src/util/file-filter.js#L40-L46

That caused `/dist/bundles/node_modules/ipfs-css/fonts` to be excluded from the production package :-(

The fix here is to put all assets under hardcoded `/dist/bundles/assets/` which is not excluded.

cc @jessicaschilling – ignore if fonts look weird, this will fix them

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,14 +61,15 @@
     "ci:build": "./ci/update-manifest.sh && npx yarn@1.22.4 build",
     "ci:test": "npx yarn@1.22.4 test",
     "ci:lint": "npx yarn@1.22.4 lint",
-    "beta-build": "docker build -t ipfs-companion-beta-build --build-arg USER_ID=$(id -u ${USER}) --build-arg GROUP_ID=$(id -g ${USER}) . && mkdir -p build && docker run --rm -it --net=host -e RELEASE_CHANNEL=beta -v $(pwd)/build:/home/node/app/build ipfs-companion-beta-build yarn ci:build",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -61,14 +61,15 @@
...
+    "beta-build": "docker rmi -f ipfs-companion-beta-build && docker build -t ipfs-companion-beta-build --build-arg USER_ID=$(id -u ${USER}) --build-arg GROUP_ID=$(id -g ${USER}) . && mkdir -p build && docker run --rm -it --net=host -e RELEASE_CHANNEL=beta -v $(pwd)/build:/home/node/app/build ipfs-companion-beta-build yarn ci:build",
+    "release-build": "docker rmi -f ipfs-companion-release-build && docker build -t ipfs-companion-release-build --build-arg USER_ID=$(id -u ${USER}) --build-arg GROUP_ID=$(id -g ${USER}) . && mkdir -p build && docker run --rm -it --net=host -e RELEASE_CHANNEL=stable -v $(pwd)/build:/home/node/app/build ipfs-companion-release-build yarn ci:build",
     "dev-build": "npx yarn@1.22.4 && npx yarn@1.22.4 build",
     "yarn-build": "npm run dev-build"
   },
   "private": true,
   "preferGlobal": false,
   "resolutions": {
+    "v8": "npm:empty-module@0.0.2",
```
lidel, line 72: This suppresses an error during the build related to metrics in js-ipfs embedded in Brave – not used, but was annoying.

Issue #896, opened 2020-06-18
Revamp Companion "sales" and overview materials


lidel:
Chrome Web Store is the main user aquisition channel for this browser extension:

- https://chrome.google.com/webstore/detail/ipfs-companion/nibjojkomfdiaoajekhjakgkdhaomnch

Current listing is pretty old. It consists of `Overview` text and screenshots.
There are also additional "promo tile" graphics displayed on the store listings (note resolutions):

> ![2020-06-18--14-19-04](https://user-images.githubusercontent.com/157609/85019417-1d57c900-b16f-11ea-83be-52b499c9395d.png)

----

@jessicaschilling as you noted in [IPFS GUI tools: June 2020 survey analysis](https://docs.google.com/document/d/1V5sDSxMqhhplpcB8u8CffiGWHUvw-t4p_sn5vigdR90), a big chunk of users does not feel like they understand what to expect when they install this extension. Perhaps rewriting `Overview` and producing better screenshots (macOS?) could take us 80% where we need to be?

jessicaschilling:
First pass at Chrome Store text below — once we arrive at something more final, can amend/extrapolate for other channels.

Also want to create an ipfs.io link to the privacy policy so we're not asking folks to copy/paste a long GitHub link.

@lidel -- initial thoughts?

**TAGLINE (note this appears in big type at the bottom of the Chrome Store page):**
Harness the power of IPFS in your browser

**BODY:**

IPFS Companion harnesses the power of your locally running IPFS node (either through the IPFS Desktop app or the command-line daemon) directly inside your favorite browser, enabling support for ipfs:// paths, automatic IPFS gateway loading of websites and file paths, easy IPFS file import and sharing, and more.

IPFS is a peer-to-peer hypermedia protocol designed to make the web faster, safer, more resilient, and more open. It enables the creation and dissemination of completely distributed sites and applications that don’t rely on centralized hosting and stay true to the original vision of an open, flat web. Visit https://ipfs.io to learn more.

FEATURES
- Automatically use IPFS to retrieve sites, files, and other resources already stored on IPFS (including redirecting DNSLink hostnames to IPNS names via your choice of gateway)
- Support IPFS paths (ipfs://) in your browser, and redirect them to the gateway of your choice
- Share files from your browser by importing them to your local IPFS node via right-click or drag-and-drop, including the option to preload files at a public gateway
- See how many peers you’re connected to at a glance from your browser’s menu bar
[...]

PR #898, opened 2020-06-19
style: update manifest_extensionDescription

jessicaschilling:
This PR changes `manifest_extensionDescription` in order to update taglines in Chrome and Firefox extension stores.

Old: "Browser extension that simplifies access to IPFS resources"
New: "Harness the power of IPFS in your browser"

Ref https://github.com/ipfs-shipyard/ipfs-companion/issues/896


jessicaschilling:
@lidel -- turns out the tagline in the extension stores comes from the manifest. If we can get this included in a coming release, that would be great. 😊

Issue #902, opened 2020-06-25
Missing favicon


lidel:
Various screens are lacking favicon:

- Welcome screen
- Preferences
- Share files via IPFS (quick-upload)

We probably could reuse one from https://github.com/ipfs/dir-index-html/pull/36 – 
@jessicaschilling  thoughts?

jessicaschilling:
@lidel I'm seeing favicons on all those pages. Confused.

![image](https://user-images.githubusercontent.com/1507828/85794986-c1ec8480-b6f4-11ea-9c2a-d463e3e83fb4.png)

jessicaschilling:
Update: Favicons display in Chrome, but not Firefox.

lidel:
I guess when there is no explicit favicon defined in HTML loaded from browser extension Chrome fallbacks to extension's icon, and Firefox does not. So adding explicit definition should fix firefox.

PR #906, opened 2020-06-30
fix: add favicon to welcome, options, quick-import pages

jessicaschilling:
Adds favicon as inline data to `<head>` of the following pages, so Firefox has something to use (Chrome seems to be falling back to the extension's icon in absence of explicit favicon):
- Welcome (dist/landing-pages/welcome/index.html)
- Options (dist/options/options.html)
- Quick import (dist/popup/quick-import.html)

Closes https://github.com/ipfs-shipyard/ipfs-companion/issues/902.


jessicaschilling:
@lidel - this may be the littlest PR you see all week.

Issue #806, opened 2019-10-30
Should Companion recover from missing local gateway?


lidel:
Recovery from dead public gateways was introduced in #640.

What about scenarios when _local_ gateway goes offline?

Right now, opening http://127.0.0.1:8080/ipns/docs.ipfs.io/ ends up with an error:

> ![2019-10-30-114745_905x465_scrot](https://user-images.githubusercontent.com/157609/67852642-13fcec00-fb0d-11e9-9515-f91077fda21f.png)

Should we recover from this and redirect to https://ipfs.io/ipns/docs.ipfs.io/ ?

jessicaschilling:
@lidel I can certainly stub out a custom error page, but how tough would implementation of this be? I don't want to expose a rabbit hole.

lidel:
@jessicaschilling in this case it should be relatively easy to wire it up :crossed_fingers: , in broad strokes:

we would add code in [ipfs-request.js / onErrorOccured handler](https://github.com/ipfs-shipyard/ipfs-companion/blob/a219d2b3556404fc9c6cff6d5c8a6e0f072f957e/add-on/src/lib/ipfs-request.js#L345) to check if failed request was for custom gateway defined in _Preferences_ or a DNSLink website, and if so, open a static page, passing the failed URL after `#`, so the error page would read original URL via `window.location.hash`

If you stub the page and open a PR with it, I can block some time to wire it up.

jessicaschilling:
https://github.com/ipfs-shipyard/ipfs-companion/pull/920 will close this once merged.

PR #920, opened 2020-09-10
[WIP] feat: offline error page

jessicaschilling:
### About
This PR adds a more friendly and useful error page when either user's custom gateway is offline, or user tries to load a DNSLink-enabled site from an unavailable remote host.

Closes https://github.com/ipfs-shipyard/ipfs-companion/issues/806.

#### To do:
- [ ] Move text into i18n keys once agreed upon
- [ ] Wire up buttons, error-check condition, and display for `window.location.hash`

image
image

### DIFF
--- a/add-on/src/landing-pages/offline/page.js
+++ b/add-on/src/landing-pages/offline/page.js
@@ -1,0 +1,53 @@
+'use strict'
+
+const html = require('choo/html')
+const logo = require('../../popup/logo')
+const { renderTranslatedLinks, renderTranslatedSpans } = require('../../utils/i18n')
+
+// Assets
+const libp2pLogo = '../../../images/libp2p.svg'
+const multiformatsLogo = '../../../images/multiformats.svg'
+const ipldLogo = '../../../images/ipld.svg'
+
+// Colors
+const colorIpfsLogo = '#57cbd0'
+const colorWhite = '#ffffff'
+const colorYellow = '#f39021'
+
+function createOfflinePage (i18n) {
+  return function offlinePage (state, emit) {
+    const { isIpfsOnline, peerCount } = state
+    const openWebUi = (page) => () => emit('openWebUi', page)
+
+    // Set translated title
+    document.title = i18n.getMessage('page_landingOffline_title')
+
+    return html`
+    <div class="mw5 mw8-ns ph5-ns center pv5">
+      <div class="fl w-30-ns w-100 tl">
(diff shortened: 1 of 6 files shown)


jessicaschilling:
@lidel - No need to look at code for this yet, just wanted to confirm this direction with you before moving the text into i18n keys. Note that the explanatory text is intended to change depending on error context, so a wired-up version of this would only display one text-plus-button(s) combo (either "Your custom gateway ..." or "You tried to load ..."). Do you mind a quick look? Thanks 😊

add-on/src/landing-pages/offline/page.js
```diff
@@ -0,0 +1,53 @@
...
+        <h3 class="f4 fw4 mt0 monospace charcoal-muted" style="word-break:break-all;">window.location.hash123123123123123123123123123</h3>
+        <p class="f4 fw4">
+          Your custom local gateway might be offline. You can update your local gateway in the preferences, or try loading this resource from a public gateway instead.
+        </p>
+        <div class="tc tr-ns">
+          <button class="f5 pv3 ph4 mh2 mb3 b white br2 bn bg-charcoal-muted hover-bg-charcoal pointer">Update Preferences</button>
+          <button class="f5 pv3 ph4 mh2 mb3 b white br2 bn bg-teal-muted hover-bg-teal pointer">Try Public Gateway</button>
+        </div>
+        <p class="f4 fw4">
+          You tried to load a DNSLink-enabled site from a remote host that's currently unavailable. Try loading it from the original host instead.
```
lidel, line 42: I believe this will be less ambigous:

```suggestion
          You tried to load a DNSLink-enabled site from a custom local gateway that's currently unavailable. Try loading it from the original host instead.
```

I think there is an additional scenario here is when redirect to local gateway is disabled and the original server is down. In that case user wants to recover in reverse direction:

```suggestion
          You tried to load a DNSLink-enabled site from a host that's currently unavailable. Try loading it from the local/public gateway instead.
```

lidel (commented):
Looks great, thank you :pray: 

SidenotE: this screen will be useful not only for recovering when local gateway is down. For example,  if DNSLink is present, if original server is down, Companion is able to recover by loading website from local gateway (example below).

add-on/src/landing-pages/offline/page.js
jessicaschilling: Sounds good - resolved in https://github.com/ipfs-shipyard/ipfs-companion/pull/920/commits/f9b8933c54e1e724d78fe48f6d4a38ae28339a65.

jessicaschilling:
@lidel -- I think this is ready for you to wire up at your leisure. CI is failing due to missing manifest.json but looks like that's intentionally excluded? Let me know if there's anything else I can do.

PR #929, opened 2020-10-15
feat: remember manual opt-ins and opt-outs per site

### LINKED ISSUE #921: Opt-out proto.shool from the DNSLink redirect by default
As noted in https://github.com/ProtoSchool/protoschool.github.io/issues/548 there is a need for disabling DNSLink-redirect-by-default so the users of ProtoSchool do not lose progress information store in `window.localStorage`, which is tied to the Origin  at `https://proto.school` (with option to opt-in via existing "Enable" toggle in browser action)

We have opt-out list already, so the quick fix is to add one-time migration that injects `proto.school` to user's list. To do it in a generic way we should include `app.fleek.co` and perhaps `slate.host` too. 

ps. In the long term, we should do this in a generic way to help in addressing future requests like this without the need for making a new release of IPFS Companion – perhaps loading a JSON with 'opt-out list' hosted in this very repo? Ideally we would  have GithubAction that takes care of publishing and load it from IPNS.

lidel:
This PR improves the way we handle manual opt-out performed per site by the user by storing both opt-out and opt-in decisions made by the user.

### Explicit opt-in and opt-out lists in _Preferences_

- _Preferences_ screen now has separate list for opt-outs and opt-ins:
  > 
- In the past we only had opt-out list named `noIntegrationsHostnames`. 
  - Items from the old list are automatically migrated to the new one named `disabledOn`.
- If for some reason item is present on both lists, opt-out takes a priority.

### Menu toggle adds/moves hostname between lists

- The above lists can be modified by hand on _Preferences_ screen, 
  however the expectation is that most of the time users will not touch them. 
  - Instead, user will be toggling integrations for current website via browser action menu:
    > 

#### Short term (this PR)

- In this PR we are adding `proto.school` to opt-out list if it was not already present on either of lists.
  - This way we are able to tweak implicit default per website while respecting  preexisting user choices. 
  - Also, in the future we may run opt-in metrics  that compare lengths of both lists to gain better insight into user's behavior.
- We also include `app.fleek.co` – https://github.com/ipfs-shipyard/ipfs-companion/pull/929#pullrequestreview-509501401  (does not work from local gateway atm due to stripe.com limitation)

#### Long term (separate PR)

We will be looking into supporting opt-out based on HTTP header, so any IPFS-powered website can change the implicit behaviour to opt-out. Tracked in https://github.com/ipfs-shipyard/ipfs-companion/issues/930

Closes #921 cc @terichadbourne @jessicaschilling

### DIFF
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -344,7 +344,7 @@
     "description": "An option title on the Preferences screen (option_disabledOn_title)"
   },
   "option_disabledOn_description": {
-    "message": "Sites in this list (one hostname per line) will have IPFS integrations disabled.",
+    "message": "Sites in this list (one URI or hostname per line) will always load with IPFS integrations disabled. Toggling \"Enable on ...\" from your browser's action menu while visiting a site will also add or remove it from this list.",
     "description": "An option description on the Preferences screen (option_disabledOn_description)"
   },
   "option_enabledOn_title": {
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -352,7 +352,7 @@
     "description": "An option title on the Preferences screen (option_enabledOn_title)"
   },
   "option_enabledOn_description": {
-    "message": "Sites in this list (one hostname per line) will have IPFS integrations enabled.",
+    "message": "Sites in this list (one URI or hostname per line) will always load with IPFS integrations enabled. Toggling \"Enable on ...\" from your browser's action menu while visiting a site will also add or remove it from this list.",
     "description": "An option description on the Preferences screen (option_enabledOn_description)"
   },
   "option_publicGatewayUrl_title": {
--- a/add-on/_locales/en/messages.json
+++ b/add-on/_locales/en/messages.json
@@ -339,13 +339,21 @@
     "message": "Do not use if your IPFS node does not support *.ipfs.localhost. Redirecting to a path-based gateway breaks the origin-based security isolation of DNSLink websites, so make sure you understand the related risks.",
     "description": "A warning on the Preferences screen, displayed when URL does not belong to Secure Context (option_customGatewayUrl_warning)"
   },
-  "option_noIntegrationsHostnames_title": {
-    "message": "Site Opt-Out List",
-    "description": "An option title on the Preferences screen (option_noIntegrationsHostnames_title)"
+  "option_disabledOn_title": {
+    "message": "Manual Opt-Out List",
+    "description": "An option title on the Preferences screen (option_disabledOn_title)"
   },
-  "option_noIntegrationsHostnames_description": {
-    "message": "Sites in this list (one hostname per line) will have all IPFS integrations disabled.",
-    "description": "An option description on the Preferences screen (option_noRedirectHostnames_description)"
+  "option_disabledOn_description": {
+    "message": "Sites in this list (one hostname per line) will have IPFS integrations disabled.",
+    "description": "An option description on the Preferences screen (option_disabledOn_description)"
+  },
+  "option_enabledOn_title": {
+    "message": "Manual Opt-In List",
+    "description": "An option title on the Preferences screen (option_enabledOn_title)"
+  },
+  "option_enabledOn_description": {
+    "message": "Sites in this list (one hostname per line) will have IPFS integrations enabled.",
+    "description": "An option description on the Preferences screen (option_enabledOn_description)"
   },
   "option_publicGatewayUrl_title": {
     "message": "Default Public Gateway",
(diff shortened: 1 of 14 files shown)


add-on/src/lib/options.js
```diff
@@ -184,4 +191,26 @@ exports.migrateOptions = async (storage) => {
...
+      await storage.set({ disabledOn: disabledOn.concat(noIntegrationsHostnames) })
+      await storage.remove('noIntegrationsHostnames')
+    }
+  }
+
+  { // ~v2.15.x: opt-out some hostnames if user does not have excplicit rule already
+    const { enabledOn, disabledOn } = await storage.get(['enabledOn', 'disabledOn'])
+    for (const fqdn of [
+      'proto.school', //  https://github.com/ipfs-shipyard/ipfs-companion/issues/921
+      'app.fleek.co' // TODO: confirm if ok
```
jsonsivar, line 208: Janison from Fleek here - this looks good.  Thanks for this.  cc @studna

add-on/_locales/en/messages.json
```diff
@@ -339,13 +339,21 @@
     "message": "Do not use if your IPFS node does not support *.ipfs.localhost. Redirecting to a path-based gateway breaks the origin-based security isolation of DNSLink websites, so make sure you understand the related risks.",
     "description": "A warning on the Preferences screen, displayed when URL does not belong to Secure Context (option_customGatewayUrl_warning)"
   },
-  "option_noIntegrationsHostnames_title": {
-    "message": "Site Opt-Out List",
-    "description": "An option title on the Preferences screen (option_noIntegrationsHostnames_title)"
+  "option_disabledOn_title": {
+    "message": "Manual Opt-Out List",
```
jessicaschilling, line 343:
```suggestion
    "message": "Always Disable IPFS Integrations",
```

add-on/_locales/en/messages.json
```diff
@@ -339,13 +339,21 @@
...
   },
-  "option_noIntegrationsHostnames_description": {
-    "message": "Sites in this list (one hostname per line) will have all IPFS integrations disabled.",
-    "description": "An option description on the Preferences screen (option_noRedirectHostnames_description)"
+  "option_disabledOn_description": {
+    "message": "Sites in this list (one hostname per line) will have IPFS integrations disabled.",
+    "description": "An option description on the Preferences screen (option_disabledOn_description)"
+  },
+  "option_enabledOn_title": {
+    "message": "Manual Opt-In List",
```
jessicaschilling, line 351:
```suggestion
    "message": "Always Enable IPFS Integrations",
```

add-on/_locales/en/messages.json
```diff
@@ -339,13 +339,21 @@
...
-    "description": "An option title on the Preferences screen (option_noIntegrationsHostnames_title)"
+  "option_disabledOn_title": {
+    "message": "Manual Opt-Out List",
+    "description": "An option title on the Preferences screen (option_disabledOn_title)"
   },
-  "option_noIntegrationsHostnames_description": {
-    "message": "Sites in this list (one hostname per line) will have all IPFS integrations disabled.",
-    "description": "An option description on the Preferences screen (option_noRedirectHostnames_description)"
+  "option_disabledOn_description": {
+    "message": "Sites in this list (one hostname per line) will have IPFS integrations disabled.",
```
jessicaschilling, line 347:
```suggestion
    "message": "Sites in this list (one URI or hostname per line) will always load with IPFS integrations disabled. Toggling \"Enable on ...\" from your browser's action menu while visiting a site will also add or remove it from this list.",
```

add-on/_locales/en/messages.json
```diff
@@ -339,13 +339,21 @@
...
+    "description": "An option description on the Preferences screen (option_disabledOn_description)"
+  },
+  "option_enabledOn_title": {
+    "message": "Manual Opt-In List",
+    "description": "An option title on the Preferences screen (option_enabledOn_title)"
+  },
+  "option_enabledOn_description": {
+    "message": "Sites in this list (one hostname per line) will have IPFS integrations enabled.",
```
jessicaschilling, line 355:
```suggestion
    "message": "Sites in this list (one URI or hostname per line) will always load with IPFS integrations enabled. Toggling \"Enable on ...\" from your browser's action menu while visiting a site will also add or remove it from this list.",
```

jessicaschilling (changes requested):
Suggested some prefs text tweaks for clarity ...

lidel:
@jessicaschilling merged your  suggestions, but removed "URI" from preference label [`853099b` (#929)](https://github.com/ipfs-shipyard/ipfs-companion/pull/929/commits/853099b15a2ea100e0ad85b62b83271658b6ff48) – we only support hostnames.  Would that be acceptable?

jessicaschilling (approved):
LGTM, thank you!

lidel:
I'm going to ship this to Beta soon, so we can dogfood it more, 
hopefully shipping to Stable next week.

Issue #933, opened 2020-10-20
Don't pop up new tabs on updates


kallisti5:
Yeah... I was mid-typing something **important** and ipfs-companion popped up a new window for an update release notes.
Please *DON'T* do this anymore.

I don't mind seeing release notes in the ipfs-companion extension (maybe red circle notification or something over the icon), but spamming users with pop-ups is a sure-fire way to get a bunch of uninstalls.

kallisti5:
At minimum... this should be "opt-in" and **not** "opt-out"

![disabled](https://user-images.githubusercontent.com/100110/96603791-ff83a780-12b9-11eb-996f-87dc66388ae2.png)

lidel:
Ouch, thank you for opening this issue and sorry for inconvinience :( 
We will definitely improve the default behaviour and remove surface for disruptions like the one you experienced.

@jessicaschilling thoughts on how we can make this less disruptive?

Perhaps we could add some icon to the header section in the popup menu?

Prior art from 1password:

> ![2020-10-20--17-31-36](https://user-images.githubusercontent.com/157609/96609102-72147780-12fa-11eb-99fd-058c8b6b4901.png)

when clicked:

> ![2020-10-20--17-31-58](https://user-images.githubusercontent.com/157609/96609132-79d41c00-12fa-11eb-943b-bd9947bda60e.png)

kallisti5:
@lidel that seems like a more reasonable design :-)
Generally the add-ons that spam with pop-ups are the ones you *really* don't want to install lol.

jessicaschilling:
@kallisti5 @lidel -- this sounds good to me. I'd still leave the toggle in the prefs though, just default to off ... some users may genuinely want the active reminder 😊

@lidel, I can stub this out once my IRL emergency dies down a bit, if you don't get to it first. Will probably still need a little help with the conditional on-first-load part though -- so either I can build the UX and you can wire it up, or you could drop in the code for some really barebones UI. LMK what you'd rather.

lidel:
@jessicaschilling I can wire things up. no rush, I won't have bandwidth for this any time soon, so we can tackle this next+ week.

jessicaschilling:
@lidel - https://github.com/ipfs-shipyard/ipfs-companion/pull/935 should be a useful start. Let me know if I can do anything else - thanks!

kallisti5:
I can confirm 2.16.0 came out and the upgrade notification is a lot more reasonable.
[...]

Issue #662, opened 2019-01-21
Add an indicator to show that resources have been loaded over IPFS


da2x:
1. Install IPFS Companion extension.
2. Visit a website that loads some resources over IPFS (like https://www.ctrl.blog).

**Expected:** There should be an indicator to show that some resources on the current page have been loaded over IPFS. The page is loaded over HTTPS but some static assets are served from an DNSLink enabled subdomain.

**Actual:** There is no visual indicator.

**Notes:** I'd normally suggest fixing up the pageAction icon, but browser support for pageAction isn't all that great compared to browserAction. Maybe just change the color of the browserAction icon? Attaching [a green version](https://github.com/ipfs-shipyard/ipfs-companion/files/2777197/ipfs-logo-active.svg.zip.zip) for this purpose.

lidel:
Subresource requests are easy to detect: [`ResourceType`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType) !==  `main_frame`, and parent page is in `request`'s `documentUrl`/`originUrl` (Firefox) or `initiator` (Chromium)

I agree, UX will be the tricky to get right, especially on Chromium browsers where it is not possible to show `pageAction` and `browserAction` at the same time.

What if we  stop displaying peer count on the badge, and instead show the number of redirected resources on current page? (if any)

jessicaschilling:
Revisiting this one with the idea that we can probably accomplish this without removing the "number of peers" indicator. The below sketch:
- Replaces the filled "off" state icon with a stroke version for greater contrast between "off" and "on"
- Offers a few versions of an "on, but not loading resources via IPFS" state (aka "partially on") for discussion
- Reserves the use of the full-color IPFS cube for an "on, and loading resources via IPFS" state

![image](https://user-images.githubusercontent.com/1507828/87477915-2dd05780-c5e6-11ea-9669-ac0a809bea04.png)

@lidel, any thoughts?

lidel:
I like the stroke version for "off" and aqua stroke for "on, but current tab had no resources loaded from IPFS", but at the same time worry that this aqua stroke would be shown for the most of the time as people browse regular internet without IPFS resources, effectively replacing the cube as default companion icon.

[...]

PR #935, opened 2020-10-21
feat: icon in browser action menu upon version update

jessicaschilling:
Closes https://github.com/ipfs-shipyard/ipfs-companion/issues/933.

## To do
- [x] Add icon to browser action menu; icon links to latest release notes in new tab
- [x] Make that icon only appear upon first load of new version
- [x] Default "Show Release Notes on Update" in prefs page to off

**Note:** There was discussion in https://github.com/ipfs-shipyard/ipfs-companion/issues/933 about changing the browser bar icon itself in some way to indicate the new version. Suggesting we leave this off (at least for now) - we have significant cube icon mods in https://github.com/ipfs-shipyard/ipfs-companion/issues/662 and I think the cube icon may be "full up" on its ability to convey different states once that issue is implemented.

## Screenshot
image


jessicaschilling:
Hi @lidel -- is this enough to get started?
Thanks!

lidel:
Thanks! I'll try to wire this up today :crossed_fingers:

lidel (approved):
LGTM! Thank you @jessicaschilling 

- I've wired it up 
  - Turned out to be more involved than expected, mostly because choo does not support async render functions: I had to calculate show/hide flag in background page and then send it to the popup. 
- Added migration from opt-out to opt-in
- If someone opts-in to opening release notes in a new tab, the redundant notification via icon won't be displayed.

Mind double checking if works as expected?

jessicaschilling:
@lidel -- confirmed everything works as expected in both Chrome and Firefox on Mac. Also saw the same FF visual glitch, and adjusted icons' `viewBox`es to compensate. Mind taking one final look on your end?

lidel:
Look good in Firefox now, thanks!

Issue #790, opened 2019-10-13
Enable pinning when the ipfs daemon is in offline mode


rehno-lindeque:
Currently ipfs companion is completely disabled when ipfs is run in offline mode. I.e.

```
ipfs daemon --offline
```

However, it should still be possible to pin content when not connected to the network. That is

* `This Page > Add to IPFS (Keep Filename)`
* `This Page > Add to IPFS`
* `Tools > Open Web UI`

Should all be available. Also possibly 

* `Tools > Share files via IPFS`

For this last menu option, the wording "Share" may introduce some confusion. On the other hand, presumably, someone that is savvy enough to run with `--offline` probably already understands that sharing will only take effect once they put the daemon into online mode.

lidel:
Thank you for filling this. It compliments #782  nicely, enabling users to add/browse content in truly offline mode.  

The fix is to enable Web UI item  in browser action menu when peer count is 0.
I believe all is needed to enable offline mode is to remove (or replace) `isIpfsOnline`  with `isApiAvailable` in two places:

https://github.com/ipfs-shipyard/ipfs-companion/blob/8512691b59020e1403d6157e53f6b18acde91a81/add-on/src/popup/browser-action/tools.js#L18

https://github.com/ipfs-shipyard/ipfs-companion/blob/8512691b59020e1403d6157e53f6b18acde91a81/add-on/src/popup/browser-action/context-actions.js#L30

@rehno-lindeque  is this something you would be interested in contributing via a PR?
I'd be happy to provide more context if needed.

rehno-lindeque:
I'd love to be able to contribute to ipfs related projects but unfortunately I can't justify it for the foreseeable future. Thank you for the suggestion though.

BrianBelhumeur:
👋🏻 I'll take this one

lidel:
@BrianBelhumeur sweet! 
I believe the code did not change much so entry points described in https://github.com/ipfs-shipyard/ipfs-companion/issues/790#issuecomment-542858298 should be still valid.

lidel:
@BrianBelhumeur hi, are you  still interested in this, or can someone else pick this up?

BrianBelhumeur:
[...]

PR #936, opened 2020-10-23
Fix/offline mode

BrianBelhumeur:
This is a fix for https://github.com/ipfs-shipyard/ipfs-companion/issues/790

When the IPFS daemon is running in offline mode, pinning should still be enabled in the extension.

This fix functions by replacing `isIpfsOnline` with `isApiAvailable` to allow pinning actions in the Tools menu and in-page context menu.

### DIFF
--- a/add-on/src/lib/context-menus.js
+++ b/add-on/src/lib/context-menus.js
@@ -163,7 +163,7 @@ function createContextMenus (getState, runtime, ipfsPathValidator, { onAddFromCo
             ipfsContext = ipfsPathValidator.isIpfsPageActionsContext(currentTab.url)
           }
         }
-        const ifApi = getState().peerCount >= 0
+        const ifApi = getState().peerCount > -1
         for (const item of apiMenuItems) {
           await browser.contextMenus.update(item, { enabled: ifApi })
         }
--- a/add-on/src/popup/browser-action/context-actions.js
+++ b/add-on/src/popup/browser-action/context-actions.js
@@ -41,7 +41,7 @@ function contextActions ({
   onUnPin
 }) {
   const activeCidResolver = active && isIpfsOnline && isApiAvailable && currentTabCid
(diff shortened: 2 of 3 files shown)


lidel:
Thank you @BrianBelhumeur! Your comments from https://github.com/ipfs-shipyard/ipfs-companion/issues/790#issuecomment-715046792 were really useful and prompted me to refactor a few things (added them to this PR):

-  a7793c7 added a way to differentiate between "api is down" and "there are no peers but api is up" states and switched codebase to work better in offline mode
- 53ca68c detects offline mode and tries to precache webui into the store by prefetching it from public gateway, which partially unblocks import operations (they open in webui, and that final step fails right now)
  - remaining fix needs to happen in ipfs-webui, which right now does not work correctly in `--offline` mode: https://github.com/ipfs-shipyard/ipfs-webui/issues/853#issuecomment-715470040
  - let's park this PR for now and see if we can fix it upstream easily
    - if not, then we may change the code responsible for import, and open imported file from gateway instead via webui

BrianBelhumeur:
@lidel Thanks for all the help! I looked through the commits you provided... makes sense. Looks like the online/offline issue affects several parts of the code.

Happy to park until there's a resolution to https://github.com/ipfs-shipyard/ipfs-webui/issues/853

boomshroom:
There appear to be quite a few conflicts to getting this merged. I'm not terribly familiar with Javascript, so how bad are these conflicts that we're looking at?

I'll just quickly mention that I'll often run IPFS with --offline purely out of paranoia and instead use it as a personal archive, so I'm fairly excited to see this issue resolved.

lidel:
IIRC this is blocked by https://github.com/ipfs/ipfs-webui/issues/853#issuecomment-715470040
This PR is pretty stale, so when that webui issue is resolved we will most likely create a new PR here.

PR #940, opened 2020-10-30
docs: updated developer notes

jessicaschilling:
Updates `README.md` and `DEVELOPER_NOTES.md` to the following:
- Advises using `npm run dev-build` rather than `npm install`
- Calls out `npm run dev-build chromium` and `npm run dev-build firefox:nightly` as all-in-one shortcuts
- Updates outdated instructions for manual loading extension in FF and Chromium (and removes old animated gifs as they're outdated and redundant; can update to new gifs if we feel it's necessary)

### DIFF
--- a/DEVELOPER-NOTES.md
+++ b/DEVELOPER-NOTES.md
@@ -69,7 +69,7 @@ npm run dev-build chromium        # all-in-one
 You can also build it to manually install:
 
 ```bash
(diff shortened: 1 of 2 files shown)


jessicaschilling:
@lidel -- for your eyes. I think this takes care of all the items we'd discussed, but please amend as desired!

lidel (approved):
:+1:  thank you!

PR #945, opened 2020-11-24
fix: IP support on opt-in and opt-out lists

### LINKED ISSUE #944: Can't add an IP address to "Always Disable IPFS Integrations" list.
I have 2 nodes one of which is hosted on a server I can access over VPN.
Attempting to access the remote gateway results in redirection to 127.0.0.1.
"Always Disable IPFS Integrations" list does not accept IP addresses - after clicking off the field the IP address disappears.

Tested in Firefox and Chromium, the behavior is identical.

Sierra410:
fixes #944

### DIFF
--- a/add-on/src/lib/options.js
+++ b/add-on/src/lib/options.js
@@ -118,11 +118,23 @@ function guiURLString (url, opts) {
 exports.safeURL = safeURL
 exports.guiURLString = guiURLString
 
+// ensure value is a valid URL.hostname (FQDN || ipv4 || ipv6 WITH brackets)
+exports.isHostname = x => isFQDN(x) || isIP.v4(x) || (!isIP.v6(x) && isIP.v6(x.replace(/[[\]]+/g, '')))
+
 // convert JS array to multiline textarea
 function hostArrayCleanup (array) {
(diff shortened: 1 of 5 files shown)


jessicaschilling:
@lidel - might you have time for a review? Thanks!

lidel (approved):
Thank you @Sierra410! 

ps. I noticed potential problems with IPv6 and inlined custom ports, so I've added additional normalization/tests for those scenarios.

Sierra410:
```js
x => isFQDN(x) || isIP.v4(x) || (!isIP.v6(x) && isIP.v6(x.replace(/[[\]]+/g, '')))
```
This will report malformed IPv6s like ``]::1[`` or ``[:[:[1[`` as valid.

```js
function isHostname(x) {
	if (isFQDN(x) || isIP.v4(x))
		return true
	
	const match = x.match(/^\[(.*)\]$/)
	
	if (match == null)
		return false
	
	return isIP.v6(match[1])
}
```

lidel:
@Sierra410 good catch, mind submitting above in a new PR? :pray:  thanks!

PR #946, opened 2020-11-27
fix: isHostname reports malformed IPv6 addresses as valid

Sierra410:
Previous version of the function was reporting addresses like ``]::1[`` or even ``]:]:]1]`` as valid.


jessicaschilling:
@lidel - this is a follow-up from https://github.com/ipfs-shipyard/ipfs-companion/pull/945 - mind having a look? Thanks!

lidel (changes requested):
@Sierra410 good catch! mind adding tests in https://github.com/ipfs-shipyard/ipfs-companion/blob/master/test/functional/lib/options.test.js#L64 with those false-positives?

Sierra410:
Like this?

lidel (approved):
:+1:

PR #959, opened 2021-01-15
Change DNS resolution endpoint to allow ipns recursive resolution

Ktl-XV:
This fixes #877. I couldn't add any tests because, from what i saw, readDnslinkFromTxtRecord always get mocked. If this is a problem, and someone can point me to the right direction I could add tests.

PR #960, opened 2021-01-17
fix: redirect to native URI in Brave

lidel:
This is a cosmetic fix that ensures IPFS resources are redirected to
`ipfs://` or `ipns://` URI and that URI is displayed in the browser UI
in the address bar.

Due to the way Brave implemented URI cloak for local HTTP-based gateway
provided by go-ipfs managed by the browser itself the regular redirect
done via webRequest API did not produce native URI in address bar.

However, `tabs.update` does the trick, so we use that for root requests.
We can remove this in the future, if webRequest behavior is fixed.

cc @autonome (mitigates issue we both discovered independently)

### DIFF
--- a/test/functional/lib/ipfs-path.test.js
+++ b/test/functional/lib/ipfs-path.test.js
@@ -3,7 +3,7 @@ const { stub } = require('sinon')
 const { describe, it, beforeEach, afterEach } = require('mocha')
 const { expect } = require('chai')
 const { URL } = require('url')
-const { ipfsContentPath, createIpfsPathValidator, sameGateway } = require('../../../add-on/src/lib/ipfs-path')
+const { ipfsUri, ipfsContentPath, createIpfsPathValidator, sameGateway } = require('../../../add-on/src/lib/ipfs-path')
 const { initState } = require('../../../add-on/src/lib/state')
 const createDnslinkResolver = require('../../../add-on/src/lib/dnslink')
 const { optionDefaults } = require('../../../add-on/src/lib/options')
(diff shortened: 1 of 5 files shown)


lidel:
I'm fast-tracking this to Beta channel, so we can ship it together with #959 before BRave 1.19 hits Stable this week :dash: 

ln-infinity

test/functional/lib/ipfs-path.test.js
```diff
@@ -96,6 +97,54 @@ describe('ipfs-path.js', function () {
...
+      expect(ipfsUri(url)).to.equal('ipfs://bafybeicgmdpvw4duutrmdxl4a7gc52sxyuk7nz5gby77afwdteh3jc5bqa/wiki/Mars.html?argTest#hashTest')
+    })
+    it('should return null if there is no valid path for input URL', function () {
+      const url = 'https://foo.io/invalid/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest'
+      expect(ipfsUri(url)).to.equal(null)
+    })
+    it('should return null if there is no valid path for input path', function () {
+      const path = '/invalid/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR'
+      expect(ipfsUri(path)).to.equal(null)
+    })
```
autonome, line 145: lots of positive tests, should add negative tests too

autonome (approved):
couple minor comments, can be fixed as followups.

PR #966, opened 2021-01-20
feat: brave node indicator

### LINKED ISSUE #965: Explore visual representation in menu when running Brave node
Right now we've got a pretty clear indicator when the user is running a js-ipfs node. Should we do something similar for when a user is running a Brave-internal node?

image

Suggestion: like this, but with a less pixellated lion. Could also explicitly note `0.7.0/Brave` or similar.
image

lidel:
This adds brave badge to indicate that companion is backed by the node
managed by the brave browser: https://brave.com/ipfs-support/

SVG thanks to @jessicaschilling 
Closes  #965

## Preview (on/off)

|  |  |
| ---- | ---- |


jessicaschilling (approved):
Nice!

PR #979, opened 2021-01-30
chore: remove beta channel

### LINKED ISSUE #964: Sunsetting the beta channel
#  TLDR: switch to the Stable channel ASAP

|  [Firefox](https://www.mozilla.org/firefox/new/) / [Firefox for Android](https://play.google.com/store/apps/details?id=org.mozilla.firefox) |  [Chrome](https://www.google.com/chrome/) /  [Brave](https://brave.com/) /  [Opera](https://www.opera.com/)  /  [Edge](https://www.microsoftedgeinsider.com/)
|------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| [Install From AMO](https://addons.mozilla.org/firefox/addon/ipfs-companion/) | [Install from Chrome Store](https://chrome.google.com/webstore/detail/ipfs-companion/nibjojkomfdiaoajekhjakgkdhaomnch)  |

# Rationale

v2.17.1 illustrated problems with Chrome Web Store when publishing bugfix releases. 
Below is a plan to improve current state of things.

## Current setup

We have a dedicated Beta channel (which has separate ID) and we published a fix release there first.  
After a day, we decided the fix is important enough that we also submitted it to the Stable channel. 

Stable channel got reviewed within 1-2h, while Beta was still pending review, which defeats its purpose as a platform for fast-tracked testing. Based on our observations during 2020, the number of users impacts the priority in the review queue (~40k stable vs ~200 beta). With the user count on the Stable channel, reviews do not seem to ne an issue, and we should leverage that.

## Proposed change

Deprecate Beta channel.
Chrome Web Store supports staged rollouts, where published can define % of users to get autoupdate.

This enables us to have a single publishing channel (current Stable) and initially publish to 1% of userbase (~400 atm), then do all the QA we need. It also enables us to bump it to 10%, 25% 50% 75% each following day and monitor github/forums, increasing our confidence and making big changes much less stressful.

### Added value

- Collapsing a lot of complexity and red tape
- Single channel could be automated with less effort (https://github.com/ipfs-shipyard/ipfs-companion/issues/778)
- Improved UX when enthusiasts want to test new release (no need to disable Stable and switch to Beta, no need to manually migrate settings)

### TODOs

- [ ] we should stop announcing it in release notes, readmes and docs
- [ ] we should add code that detects user is running Beta version and urging them to switch to stable channel
  - could be big red message in the main menu, in the place of https://github.com/ipfs-shipyard/ipfs-companion/pull/935
  - I strongly believe the version should remain functional, but there should be permanent call for action urgin user to switch to the stable channel. 
  - we might be able to signal when stable/beta is installed #beta and disable beta when stable is present (if there is no API, we may leverage `ipfs block get|rm`)
-  [ ] keep Beta channel unlisted "just in case" (errors/mistakes happen, extension ID for Beta is safelisted in Brave so it may be handy to temporarily switch to different ID if we want to quickly fix some upstream issue with Chrome Web Store)

lidel:
Closes #964 

- [x] detect when running as beta and add icon indicator that opens explainer in #964:
  > Peek 2021-01-30 03-03

- [x] remove "beta channel"  from docs
- [x] keep scripts around in case we need to build a new beta with more prominent messaging in the future

### DIFF
--- a/DEVELOPER-NOTES.md
+++ b/DEVELOPER-NOTES.md
@@ -117,7 +117,6 @@ Each `npm` task can run separately, but most of the time, `dev-build`, `test`, a
 - `npm run firefox:nightly`: Run as temporary add-on in Firefox Nightly (uses one in `./firefox/`, see `get-firefox-nightly` below)
 - `npm run chromium`: Run as temporary add-on in Chromium
 - `npm run get-firefox-nightly`: Fetch latest Firefox nightly build to `./firefox/`
-- `npm run firefox:beta:add -- --update-link "https://host/path/to/file.xpi" file.xpi`: Add a manifest entry for new self-hosted beta for Firefox
 - `npm run build`: Build the add-on (copy external libraries, create `.zip` bundles for Chrome and Firefox)
 - `npm run watch`: Rebuild JS/CSS on file changes (run regular `build` first to ensure everything else is in place)
 - `npm run bundle:chromium`: Overwrite manifest and package a generic, Chromium-compatible version
@@ -134,7 +133,6 @@ Each `npm` task can run separately, but most of the time, `dev-build`, `test`, a
 Release build shortcuts:
 
 - `npm run dev-build`: All-in-one: fast dependency install, build with yarn (updates `yarn.lock` if needed)
-- `npm run beta-build`: Reproducible beta build in docker with frozen `yarn.lock`
 - `npm run release-build`: Reproducible release build in docker with frozen `yarn.lock`
 
 ## Other tips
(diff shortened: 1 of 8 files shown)


jessicaschilling (approved):
LGTM ... also 
- confirmed no beta language exists in add-on store pages
- opened https://github.com/ipfs/ipfs-docs/pull/624 to remove mention in docs.ipfs.io

Ordinarily I'd avoid keeping the explanation (what's launched from the triangle) in GitHub but considering the size and nature of this audience, seems OK.

PR #990, opened 2021-03-22
remove find-badwords.js

notslang:
badwords.json has been removed from the linter. see: https://github.com/mozilla/addons-linter/issues/2591


lidel (approved):
Thanks!

PR #997, opened 2021-05-07
feat: replace pinning with import to mfs

### LINKED ISSUE #742: Changing language around Pinning
> #### Changing language around some features 
> 
> We are at the stage where we need to **rework language** around some features provided by IPFS Companion. Current labels describe what happens at the technical level and don't tell much about the purpose unless user is already familiar with low level details of our stack.

## Motivation

Pinning keeps the data you care about from being purged when garbage collection happens. 

It is a low level concept that does not have  good UX when translated to GUI.
User is unable to assign label or manage pins in efficient manner: it is just a simple on/off toggle presented in [browser action menu](https://github.com/ipfs-shipyard/ipfs-companion/issues/689#issuecomment-470123068) for an active browser tab:

> 2019-07-26-115906_599x33_scrot:

We plan to move "pinning"  from using `ipfs pin` API to copying content to be "pinned" into MFS (which acts as implicit pin and lets user to manage "pinned" content), however if we continue using "pin" metaphor after that change happens, we will introduce a lot of confusion, especially when Web UI decides to show resources pinned with `ipfs pin` API at some point. 

When pinning via IPFS Companion switches to MFS we should stop calling `ipfs pin` and remove "pin" from UI related to this feature. 

This means we need a different name for this.

## Changes to explore

Replace "pin" with other term and stick to it in our UI

In discussion about DNSLink websites @olizilla suggested something like **"make this [..] available offline and re-host it"** (https://github.com/ipfs-shipyard/ipfs-companion/issues/710#issue-431551526). 

I imagine we could have a menu option that said "Persist and co-host this", "Keep co-hosting this"

### LINKED ISSUE #888: Pinning service integration: Augment "Pin IPFS Resource" functionality
_Note: This issue is part of a larger [pinning service integration epic](https://github.com/ipfs/ipfs-gui/issues/91) undertaken spring/summer 2020._

### The need

Augment functionality of existing "Pin IPFS Resource" toggle to support the use of third-party pinning services:

- [ ] Keep UI of existing "Pin IPFS Resource" toggle as-is (don't break UX for people who are used to "pinning via Companion")
- [ ] When a resource is not pinned, and the toggle is clicked, in addition to creating low-level pin (current behavior), it also imports to MFS and opens imported resource in the same UI as "Quick Import" (e.g. takes you to appropriate page in Files screen)

### Context

Advantages of this approach are:
- works the same as importing non-IPFS data, reducing cognitive overhead
- they can immediately manage imported resource via UI on webui's Files screen
- works fine even if user pins entire wikipedia (lazy-loading, only pages they actually visited are kept around)

Original note from PRD: 

> (TBD) Feature in IPFS Companion that allows a user to pin data to their chosen pinning service
> - This is likely unnecessary with auto-upload rules for pinned files

> User flow:
> - Users can use Desktop or WebUI to connect their local node to an IPFS pinning service
> - User right-clicks a file or image and selects “pin to IPFS”
> - If they have an auto-upload rule configured, that file is automatically added to their associated pinning service

lidel:
This PR changes the way IPFS content is kept around.

- Instead of using low level pins we copy the item to MFS, unifying experience to match recent changes in [ipfs-webui v2.12](https://github.com/ipfs/ipfs-webui/releases/tag/v2.12.0) and [ipfs-desktop v0.15.0](https://github.com/ipfs/ipfs-desktop/releases/tag/v0.15.0).
- This also removed page-action, because it was Firefox-specific feature and made maintenance and testing more difficult (now we have same UX in all browsers).

Closes #742
Closes #888
Closes https://github.com/ipfs/ipfs-gui/issues/91


add-on/_locales/en/messages.json
```diff
@@ -83,13 +83,13 @@
...
-    "description": "A menu item in Browser Action pop-up (panel_pinCurrentIpfsAddress)"
+  "panel_importCurrentIpfsAddress": {
+    "message": "Import to Files at My Node",
+    "description": "A menu item in Browser Action pop-up (panel_importCurrentIpfsAddress)"
   },
-  "panel_pinCurrentIpfsAddressTooltip": {
-    "message": "Pin this page's IPFS resources to your node to have a local copy that's available offline and never thrown away.",
-    "description": "A menu item tooltip in Browser Action pop-up (panel_pinCurrentIpfsAddressTooltip)"
+  "panel_importCurrentIpfsAddressTooltip": {
+    "message": "Import this tab's IPFS resource to your Files to have a persistent local copy that's available offline and never thrown away.",
```
lidel, line 91: @jessicaschilling this is the only change in UI, lmk if the label and tooltip explainer could be improved :pray: 

> import-to-Files-2021-05-07--21-50-40

jessicaschilling: @lidel Maybe "on My Node" rather than "at"?

add-on/_locales/en/messages.json
```diff
@@ -83,13 +83,13 @@
     "message": "Enable/disable all IPFS integrations on $1.",
     "description": "A menu item tooltip in Browser Action pop-up (panel_activeTabSiteIntegrationsToggleTooltip)"
   },
-  "panel_pinCurrentIpfsAddress": {
-    "message": "Pin IPFS Resource",
-    "description": "A menu item in Browser Action pop-up (panel_pinCurrentIpfsAddress)"
+  "panel_importCurrentIpfsAddress": {
+    "message": "Import to Files at My Node",
```
jessicaschilling, line 87:
```suggestion
    "message": "Import to Files on My Node",
```

add-on/_locales/en/messages.json
jessicaschilling, line 91:
```suggestion
    "message": "Import this tab's IPFS resources to your Files to have a persistent local copy that's available offline and never thrown away.",
```

jessicaschilling (approved):
Two small text suggestions. Thanks!

lidel:
Thanks! Final look:

> 


This will confuse existing users who are used to seeing low level pin controls there, but the new flow is way more user-friendly  and enables users to do remote pinning with UI from webui v2.12: https://github.com/ipfs/ipfs-gui/issues/91

PR #1044, opened 2022-01-13
feat: MV3 boilerplate

meandavejustice:
This is the starting point for Manifest v3 extension.

What currently works?
* simple redirects from `ipfs.io` and `dweb` links.
* dnslink urls are fetched and added to the DeclarativeNetRequest api ruleset for redirection.
* added workflow to start testing out CI
* add typescript
* add prettier
* add eslint
* building with esbuild for ⚡ fast builds
* adds boilerplate for popup

All of this is in `/mv3`

cc/ @lidel


mv3/package.json
```diff
@@ -0,0 +1,60 @@
...
+      "browser": true,
+      "webextensions": true
+    }
+  },
+  "browserslist": {
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
```
lidel, line 44: No need to support everything, let's narrow this down to runtimes where people use Companion:

```suggestion
    "production": [
      "last 1 chrome version",
      "last 1 firefox version",
      "Firefox ESR"
    ],
```

mv3/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
```
lidel, line 48: Will add Safari when they support all the APIs we need, so far I don't believe it's the case. 
```suggestion
      "last 1 chrome version",
      "last 1 firefox version"
```

mv3/addon/background.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (!url) return true;
+  return url.includes("chrome://");
+}
+
+async function resolveDNSAndAddRule(domain) {
+  const cid = await client.dns(domain);
+  if (!cid) return;
+
+  const id = Math.floor(Math.random() * 29999);
+  // @ts-ignore
```
lidel, line 51: `@ts-ignore` like this triggers github action check which prints same warning 3 times below (can we remove this noise?)

lidel (changes requested):
Thanks!  Small asks before we merge (inline)

meandavejustice:
@lidel Comments addressed 👍

meandavejustice:
@lidel Could I get yours 👀 on this again? Thanks!

mv3/addon/rules.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "condition": { "urlFilter": "||ipfs.io", "resourceTypes": ["main_frame"] }
+  },
+  {
+    "id": 2,
+    "priority": 1,
+    "action": {
+      "type": "redirect",
+      "redirect": {
+        "transform": { "scheme": "http", "host": "localhost", "port": "8080" }
```
lidel, line 19: :thought_balloon: given that port will be something that can be changed by the user, we will most likely end up adding these as dynamic rules during startup.

.github/workflows/ci-mv3.yml
```diff
@@ -0,0 +1,71 @@
...
+          restore-keys: |
+            ${{ runner.os }}-${{ hashFiles('package*json', 'yarn.lock') }}
+            ${{ runner.os }}-
+
+      - name: Restore node_modules
+        id: yarn-cache
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-${{ hashFiles('package*json', 'yarn.lock') }}
```
lidel, line 49: :warning: please  remove `yarn.lock` and this cache – `npm ci` uses `package-lock.json` and ignores `yarn.lock` anyway (which is  good, we want to try using `package-lock.json` and remove `yarn` dependency)

lidel (changes requested):
Thank you @meandavejustice, and apologies for late review.

Dynamic rules are pretty powerful, and I think we should be able to live with them as long we add proper cache around their creation. 


Left some feedback below:
- :warning: important things we should fix
  - lmk if you prefer to do them in this PR or merge as-is and  track them in separate issues
- :thought_balloon: lower priority thoughts on style and nice-to-haves

mv3/addon/background.ts
```diff
@@ -0,0 +1,92 @@
...
+function isNewTabURL(url) {
+  if (!url) return true;
+  return url.includes("chrome://");
+}
+
+async function resolveDNSAndAddRule(domain) {
+  const contentPath = await client.resolve(`/ipns/${domain}`);
+  if (!contentPath) return;
+
+  const id = Math.floor(Math.random() * 29999);
```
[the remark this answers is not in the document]
meandavejustice: I could not find an answer for this in `declarativeNetRequest` docs. There is a possibility that they are namespaced or scoped to the `id` declared under `rule_resources` key in the manifest.json. But just to be safe I'm going to add a prefix to the id of the extension name :)

mv3/addon/background.ts
```diff
@@ -0,0 +1,92 @@
...
+  browser.tabs.onUpdated.addListener(retryOnTabUpdate);
+}
+
+function isNewTabURL(url) {
+  if (!url) return true;
+  return url.includes("chrome://");
+}
+
+async function resolveDNSAndAddRule(domain) {
+  const contentPath = await client.resolve(`/ipns/${domain}`);
```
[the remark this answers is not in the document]
meandavejustice: I realized this is an out of date file. I will push the updates shortly with the updated branch I've been working in 🤦. Luckily for me there was some things you caught that I wasn't doing in the new branch! so it was NOT a wasted effort :)

meandavejustice:
@lidel Thanks again for the review, I apologize for the branching mixup. I've got it working now correctly in this branch and addressed most of your comments. (I have two of them added to my todo list for the week). I'm going to go ahead and close this pr and open up a fresh one later on in the week. I need to add an updated list of features, and screenshots and clean up the git history.

Issue #1028, opened 2021-09-25
Save "Path to store imported files" setting on Import page


AndrewBelt:
**Is your feature request related to a problem? Please describe.**
On the Import page, clicking "import options" shows the desired import path that can be changed by the user. Unfortunately any changes to this setting will not be used the next time you go to the Import page.

**Describe the solution you'd like**
When focus of the text field is lost, it should save to the extension's storage so that the next time the Import page is opened, the updated folder path is used instead of the default `/ipfs-companion-imports/%Y-%M-%D_%h%m%s/`.

**Describe alternatives you've considered**

**Additional context**
A changed import path does not save to settings:
![2021-09-25T03:41:32-04:00](https://user-images.githubusercontent.com/338179/134763555-ac90454a-f5f6-4248-9856-331e0b37cdda.png)

lidel:
Sounds useful, happy to review PR if anyone wants to pick this up.

Carson12345:
I would pick this up!

PR #1054, opened 2022-02-19
refactor: necromancy to restore MV2 build

lidel:
This PR:
-  removes a lot of unused functionality (windows.ipfs proxy logic) which has been deprecated and disabled for years, which in turn removed a bunch of deps
- removes yarn (Node we now has `npm ci` and `package-lock.json` which should be good enough)

## TODO

- [x] remove yarn and all the hackery around transitive dependency overrides
- [x] bump all dependencies
- [x] switch to ipfs-core for fewer deps
- [x] remove unused deps
- [x] Firefox manifest fix to pass latest webext lint
- [x] remove remaining window.ipfs code that was pulling dead dependencies
  (https://github.com/ipfs/in-web-browsers/issues/172)
- [x] ensure https://github.com/ipfs/ipfs-companion/pull/1070 is ported (if still relevant)
- [x] fix tests (they fail due to ESM error)
  - :point_right:  TBD, but  we may need to refactor things to ESM and just call it a day anyway
    - we removed `jest` from ipfs-webui e2e tests and [replaced it with `playwright`](https://github.com/ipfs/ipfs-webui/blob/main/test/e2e/playwright.config.js) 
- [x] docker builds (bump to latest nodejs)
- [x] browserAction icon (it is always "inactive" now) – fixed in 4e1704372f5973abbfaae6063a401dbdb595d846
- [x] something about is-ipfs  (update?)
- [ ] <del>quick import with js-ipfs works</del> was broken in main  – can be fixed in separate PR  (or removed)
- [ ] confirm import of an image via context menu works (addFromURL)
- [x] confirm webui preload logic works (should check if CID is in datastore and preload only if it is) – closes #1059 closes #878
- [x] cleanup: remove any left-overs from window.ipfs logic – done in https://github.com/ipfs/ipfs-companion/pull/1054/commits/830ae1f017255f9218c7173a2d21c69aa1b98dd5, see https://github.com/ipfs/ipfs-companion/issues/589#issuecomment-712356448
- [x] confirm Brave onboarding works 
- [x] update client so `swarm.peers` don't fail on webtransport – a96cf2049e13c34f7a4eeab2cc93bd436d6426b6
- [x] fix the way `debug` is initialized, so  namespaces are applied before companion is initialized – a96cf2049e13c34f7a4eeab2cc93bd436d6426b6
- [x] work and require latest Node LTS

PR #1053, opened 2022-02-18
fix: truncate no. of peers to unit of "k" when > 999

Carson12345:
#1017 

**About**
The issue opener found out that when number of peers is 1000 or above, the last digit might be partially invisible and cause confusion due to the badge's max width. 

The core team replied to the issue suggesting to truncate it in the format of "1k", "2k". So I implemented the approach and made this PR.

**Original UI**

128388115-9fcf5a09-6e82-4234-9d7d-b9225213ccad

**Committed UI Change**

- [X] truncate no. of peers to unit of "k" when > 999 (e.g. the count is 2352 in this case)

Screenshot 2022-02-18 at 11 44 58 AM

Show full number when < 999

Screenshot 2022-02-18 at 11 46 59 AM

This is my first PR, please let me know if I could do it better, looking forward to help with other tasks of this repo and IPFS!

PR #1063, opened 2022-02-26
feat: save "Path to store imported files" preference from import page

Carson12345:
#1028 

**About** 
When a user specified a custom path to store imported file while importing with the quick import popup, the path will be saved to preference and will be automatically applied next time.

Screenshot 2022-02-26 at 5 59 32 PM

Screenshot 2022-02-26 at 6 00 15 PM

**Remarks**
The issue opener suggests to save path to preference as long as the textbox lost focus.

I have concern that, sometimes the path might be incomplete / not valid when user lost focus of the textbox.

So for the code I commit, I save the path to preference after a successful import to make sure the path is completed already and is valid

Please let me know if you would like me to do any other approaches.

### DIFF
--- a/add-on/src/popup/quick-import.js
+++ b/add-on/src/popup/quick-import.js
@@ -61,6 +61,11 @@ function quickImportStore (state, emitter) {
 
   emitter.on('fileInputChange', event => processFiles(state, emitter, event.target.files))
 
+  // update companion preference
+  emitter.on('optionChange', ({ key, value }) => {
+    browser.storage.local.set({ [key]: value })
+  })
+
   // drag & drop anywhere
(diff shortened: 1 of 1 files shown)


lidel (approved):
LGTM, but needs #1054  if we ever want to ship this.

Carson12345:
> LGTM, but needs #1054 if we ever want to ship this for MV2.

Thanks and looking forward to it!
Will also see if any issue that I can help with!

P.S. My another pull request 
#1053

SgtPooki (approved):

lidel (approved):
Let's land this one too (not sure when we will be able to ship a new release, but till be included).

PR #1094, opened 2022-09-01
fix(ci): windows build

whizzzkid:
Looks like builds are broken on `windows-latest`: https://github.com/ipfs/ipfs-companion/actions/workflows/ci.yml

The reported error: `gyp ERR! stack Error: Could not find any Visual Studio installation to use`

I could not test this as I don't have perms to run CI nor do I have access to a windows machine atm, but there seems like two ways to fix this:
- We can either move to node-v16 which ships with build tools.
- Or add this dependency based on https://github.com/felixrieseberg/windows-build-tools

the github `actions/setup-node@v1` is also 2 major versions behind, so should be upgraded as well, but I'm unsure of the breaking API changes between the two.

🤞🏽 this should fix the broken builds in theory.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 14.x
+          node-version: 14
 
       - name: Restore .cache
         uses: actions/cache@v2
@@ -38,6 +38,10 @@ jobs:
         with:
           path: node_modules
           key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install Microsoft Build Tools
+        if: runner.os == 'Windows'
+        run: npm install --global windows-build-tools
 
       - name: Install dependencies
         if: steps.yarn-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -39,6 +39,10 @@ jobs:
           path: node_modules
           key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
 
+      - name: Install Microsoft Build Tools
+        if: runner.os == 'Windows'
+        run: npm install --global windows-build-tools
```
whizzzkid, line 44:
```suggestion
        run: npm install --global --production windows-build-tools --vs2015
```

whizzzkid:
@lidel looks like the build is stuck https://github.com/ipfs/ipfs-companion/runs/8139100657?check_suite_focus=true

Turns out this is an issue: https://github.com/felixrieseberg/windows-build-tools/issues/208

I added the suggested fix in the thread, sorry I couldn't test it locally without a windows machine. But that didn't kill the existing build.

whizzzkid:
@lidel @SgtPooki I just realized I can run GH Actions on my own fork as well. so now this passes: https://github.com/whizzzkid/ipfs-companion/runs/8143445527?check_suite_focus=true

lidel (approved):
Thank you @whizzzkid  :heart:

PR #1101, opened 2022-10-22
refactor: working necromancy tests

whizzzkid:
Enhances #1054 

In this PR:
- I am replacing `require` with `import`/`export`
- Upgrading even more packages.
- Fixing `webExt` missing config
- Replacing `nyc` with `c8`


package.json
```diff
@@ -37,7 +38,7 @@
     "watch:js": "run-p watch:js:*",
     "watch:js:webpack": "webpack --watch --mode development --devtool inline-source-map  --config ./webpack.config.js",
     "test": "run-s test:*",
-    "test:functional": " nyc --reporter=lcov --reporter=text mocha --timeout 15000 --require ignore-styles \"test/functional/**/*.test.js\"",
+    "test:functional": "c8 mocha --timeout 15000 --require ignore-styles \"test/functional/**/*.test.js\"",
```
whizzzkid, line 41: `c8` is a drop-in replacement for `nyc`. `nyc` fails with ESM on Node > 16.12. Src:
- https://github.com/istanbuljs/nyc/issues/659
- https://github.com/istanbuljs/nyc/issues/1287
- https://stackoverflow.com/a/68095739

package.json
```diff
@@ -62,28 +63,42 @@
...
+    "all": true,
+    "reporter": [
+      "lcov",
+      "text",
+      "html"
+    ],
+    "include": [
+      "add-on/src/**"
+    ]
+  },
```
whizzzkid, line 76: moving configs like so

package.json
```diff
@@ -62,28 +63,42 @@
   },
   "private": true,
   "preferGlobal": false,
+  "c8": {
+    "all": true,
+    "reporter": [
+      "lcov",
+      "text",
+      "html"
```
whizzzkid, line 71:
```suggestion
      "text"
```

test/functional/lib/ipfs-companion.test.js
```diff
@@ -1,31 +1,36 @@
...
+global.browser = browser
+global.chrome = browser
+global.navigator = {
+  clipboard: {
+    writeText: () => {}
+  }
+}
+// We need to do this because global is not mapped otherwise, we need to stub browser and chrome runtime
+// so that the webextension-polyfill does not complain about the test runner not being a browser instance.
+const init = async () => (await import('../../../add-on/src/lib/ipfs-companion.js')).default()
```
whizzzkid, line 17: this was really the problem I had to solve because how node's module resolution works. stubbing `global.browser` was not possible when `import` for a module is called.

However doing it this way allows us to pass in stubbed references.

test/functional/lib/ipfs-companion.test.js
```diff
@@ -1,31 +1,36 @@
...
+import browser from 'sinon-chrome'
+import { URL } from 'url'
+import { AbortController } from 'abort-controller'
+import { optionDefaults } from '../../../add-on/src/lib/options.js'
+browser.runtime.id = 'testid'
+global.browser = browser
+global.chrome = browser
+global.navigator = {
+  clipboard: {
+    writeText: () => {}
```
whizzzkid, line 12: this was something extra I had to add to prevent dependency from failing.

whizzzkid (commented):
explainers

test/functional/lib/ipfs-companion.test.js
```diff
@@ -1,33 +1,37 @@
...
+global.AbortController = AbortController
+global.chrome = browser
+global.navigator = {
+  clipboard: {
+    writeText: () => {}
+  }
+}
+// We need to do this because global is not mapped otherwise, we need to stub browser and chrome runtime
+// so that the webextension-polyfill does not complain about the test runner not being a browser instance.
+const init = async () => (await import('../../../add-on/src/lib/ipfs-companion.js')).default()
```
whizzzkid, line 18: this was really the problem I had to solve because how node's module resolution works. stubbing `global.browser` was not possible when `import` for a module is called.

However doing it this way allows us to pass in stubbed references.

package.json
```diff
@@ -37,7 +38,7 @@
     "watch:js": "run-p watch:js:*",
     "watch:js:webpack": "webpack --watch --mode development --devtool inline-source-map  --config ./webpack.config.js",
     "test": "run-s test:*",
-    "test:functional": " nyc --reporter=lcov --reporter=text mocha --timeout 15000 --require ignore-styles \"test/functional/**/*.test.js\"",
+    "test:functional": "c8 mocha --timeout 5000 --require ignore-styles \"test/functional/**/*.test.js\"",
```
whizzzkid, line 41: `c8` is a drop-in replacement for `nyc`. `nyc` fails with ESM on Node > 16.12. Src:
- https://github.com/istanbuljs/nyc/issues/659
- https://github.com/istanbuljs/nyc/issues/1287
- https://stackoverflow.com/a/68095739

package.json
```diff
@@ -62,28 +63,41 @@
...
+  "c8": {
+    "all": true,
+    "reporter": [
+      "lcov",
+      "text"
+    ],
+    "include": [
+      "add-on/src/**"
+    ]
+  },
```
whizzzkid, line 75: moving configs like so instead of flags

add-on/src/lib/precache.js
```diff
@@ -1,21 +1,21 @@
...
+// import  CID  from 'cids'
 
-// const Tar = require('it-tar')
-// const pipe = require('it-pipe')
-// const all = require('it-all')
-// const concat = require('it-concat')
+// import  Tar  from 'it-tar'
+// import  pipe  from 'it-pipe'
+// import  all  from 'it-all'
+// import  concat  from 'it-concat'
```
whizzzkid, line 8: do we need these? this got replaced as part of the regex match.

add-on/src/pages/proxy-access-dialog/store.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-function createProxyAccessDialogStore (i18n, runtime) {
+export default function createProxyAccessDialogStore (_i18n, runtime) {
```
whizzzkid, line 3: just to make the editor happy, we are not using this.

add-on/src/lib/options.js
```diff
@@ -1,9 +1,9 @@
 'use strict'
 
-const isIP = require('is-ip')
-const isFQDN = require('is-fqdn')
+import { isIPv4, isIPv6 } from 'is-ip'
```
whizzzkid, line 3: isIP@4 has this breaking change.

whizzzkid (commented):
more explainers.

whizzzkid:
@lidel bringing this up again for your consideration! ❤️

add-on/src/lib/precache.js
meandavejustice: Looks like they are all for "downloading CID from public gateway". It's commented out code so I would say it's good to be removed. It's in git history if you ever want to reference it. But @lidel may have more insight

test/functional/lib/dnslink.test.js
```diff
@@ -1,11 +1,11 @@
...
-const { URL } = require('url')
-const sinon = require('sinon')
-const createDnslinkResolver = require('../../../add-on/src/lib/dnslink')
-const { initState } = require('../../../add-on/src/lib/state')
-const { optionDefaults } = require('../../../add-on/src/lib/options')
+import { describe, it, before, beforeEach, after } from 'mocha'
+import { expect } from 'chai'
+import { URL } from 'url'
+import sinon from 'sinon'
+import createDnslinkResolver from '../../../add-on/src/lib/dnslink.js'
```
meandavejustice, line 6: NIT: `.js` shouldn't be necessary for these imports.

package.json
```diff
@@ -62,28 +63,41 @@
...
     "ignore-styles": "5.0.1",
     "json": "11.0.0",
     "mem-storage-area": "1.0.3",
     "mini-css-extract-plugin": "2.5.3",
-    "mocha": "9.2.0",
+    "mocha": "9.2.2",
     "npm-run-all": "4.1.5",
-    "nyc": "15.1.0",
-    "raw-loader": "4.0.2",
+    "path": "^0.12.7",
```
meandavejustice, line 99: NIT: I usually pin my dependencies (don't use `^`), npm packages are known to frequently push breaking changes and not correctly follow semver. https://bytearcher.com/articles/semver-explained-why-theres-a-caret-in-my-package-json/

I prefer to use a [dependency manager](https://github.com/dependabot)

meandavejustice (approved):
👍 Left a couple comments since I was here

add-on/src/lib/precache.js
lidel: Yes, this is legacy code that needs to be replaced  – filled https://github.com/ipfs/ipfs-companion/issues/1102 so it is not blocking necromancy PRs ;)

lidel (commented):
I'll try to block some time this week to review this properly (there are some manual tests around Brave onboarding which I need to go over + document in `./docs`).

For now, dropped quick comments above.

add-on/src/lib/constants.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 /* eslint-env browser, webextensions */
 
-exports.welcomePage = '/dist/landing-pages/welcome/index.html'
-exports.optionsPage = '/dist/options/options.html'
-exports.tickMs = 250 // no CPU spike, but still responsive enough
+export const welcomePage = '/dist/landing-pages/welcome/index.html'
+export const optionsPage = '/dist/options/options.html'
+export const tickMs = 250 // no CPU spike, but still responsive enough
```
SgtPooki, line 6: I would love a JSdoc comment on this to explain its intention. I can guess, but it's better if we don't force others to.

add-on/src/lib/dnslink.js
```diff
@@ -1,25 +1,26 @@
...
 /* eslint-env browser */
 
-const debug = require('debug')
+import Pqueue from 'p-queue'
+
+import debug from 'debug'
+import IsIpfs from 'is-ipfs'
+import LRU from 'lru-cache'
+import { offlinePeerCount } from './state.js'
+import { ipfsContentPath, sameGateway, pathAtHttpGateway } from './ipfs-path.js'
```
SgtPooki, line 10: nit: spacing - I personally prefer organizing imports in four (potentially 5 groups depending on how large your 1st party import sections are): 
1. stdlib imports; 
2. 3p imports; 
3. 2p imports (internal to company but not to team)
4. 1p imports (internal to team)
5. local/relative imports (internal to project)

eslint-plugin-imports allows you to automate this

add-on/src/lib/ipfs-client/external.js
```diff
@@ -1,20 +1,20 @@
...
   const clientConfig = opts.apiURLString
   // https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-http-client#importing-the-module-and-usage
   const api = await create(clientConfig)
   return api
 }
 
-exports.destroy = async function (browser) {
+export async function destroy (browser) {
   log('destroy')
 }
```
SgtPooki, line 20: remove / delete?

add-on/src/popup/browser-action/gateway-status.js
```diff
@@ -17,12 +17,12 @@ function statusEntry ({ label, labelLegend, value, check, itemClass = '', valueC
...
-module.exports = function gatewayStatus ({
-  ipfsApiUrl,
+export default function gatewayStatus ({
   gatewayAddress,
   gatewayVersion,
-  swarmPeers,
-  ipfsNodeType
+  ipfsApiUrl,
+  ipfsNodeType,
+  swarmPeers
```
SgtPooki, line 25: if you're alphabetizing things like this manually, stop and wait for eslint automation. I prefer organized keys and imports as well, but they will quickly get out of date without a 'fixable' eslint rule.

package.json
SgtPooki: seems like a blocker needing addressed prior to merge.

test/functional/lib/dnslink.test.js
SgtPooki: we should standardize on esm imports and the `.js` extension is part of the filename. We should be consistent even if there is tooling that does "magic" for us. Open to hearing your thoughts @meandavejustice

test/functional/lib/ipfs-companion.test.js
SgtPooki: which dependency? Can we add a comment to the code to call it out? Do we need that dependency?

test/functional/lib/ipfs-request-dnslink.test.js
```diff
@@ -27,6 +27,7 @@ describe('modifyRequest processing of DNSLinks', function () {
   before(function () {
     global.URL = URL
     global.browser = browser
+    browser.runtime.id = 'testid'
```
SgtPooki, line 30: We're using the same id for `browser.runtime.id` in multiple tests. What is `browser.runtime.id` used for, and would it aid in test debugging to provide a different id?

Can we set the id to `this.testName` or similar?

test/helpers/acl.js
```diff
@@ -1,5 +1,3 @@
-// Converts an ACL expressed as a plain old object into an ACL Map
-const objToAcl = (obj) => new Map(pairs(obj).map(e => [e[0], new Map(pairs(e[1]))]))
 const pairs = (obj) => Object.keys(obj).map(k => [k, obj[k]])
-
-module.exports.objToAcl = objToAcl
+// Converts an ACL expressed as a plain old object into an ACL Map
+export const objToAcl = (obj) => new Map(pairs(obj).map(e => [e[0], new Map(pairs(e[1]))]))
```
SgtPooki, line 3: note: these functions are extremely convoluted. It would be nice to clear this up and make it more maintainable (not for this PR)

Is there a reason we only need these in test/helpers? Can we mock whatever logic is in src? Not creating a tracking item for this because maybe these are dumb irrelevant questions

SgtPooki (changes requested):

test/helpers/acl.js
lidel: I believe everything related to ACK / ipfs-proxy related to deprecated window.ipfs can be removed.

lidel (approved):
Thank you @whizzzkid.
In effort to unblock general Companion work I'm merging, let's continue in #1054  and/or separate PRs.

PR #1106, opened 2022-11-24
fix(ci): hanging tests

lidel:
`npm test`  takes ~30s. If it hangs for longer than a minute, it is clearly a bug.
This PR  ensures we get failure sooner than after 360m (default on github)

## Why?

CI hang problem appeared around #1101 –  i think it is c8   being very, very slow  – note that tests finish correctly, but node process hangs and never exists.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,9 @@
   "preferGlobal": false,
   "c8": {
     "all": true,
+    "src": [
+      "add-on/src/"
+    ],
     "reporter": [
       "text"
     ],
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -65,6 +65,9 @@
   "preferGlobal": false,
   "c8": {
     "all": true,
+    "src": [
+      "add-on/src/"
+    ],
```
lidel, line 70: This seems to fix CI hangs,  i think c8 was processing `./` which includes `node_modules` or :shrug:

lidel:
It seems se still have flaky tests: 
- https://github.com/ipfs/ipfs-companion/actions/runs/3541084119/jobs/5944912207#step:11:529

So slow c8 (fixed by this PR?) was not the underlying cause.

PR #1108, opened 2022-11-24
fix: allowing CSP eval to not throw errors

whizzzkid:
Allowing `unsafe-eval` for firefox


lidel:
`unsafe-eval` won't be accepted by Mozilla  – see https://github.com/ipfs/ipfs-companion/issues/269

Issue #1114, opened 2022-12-14
Add metrics tracking notice to IPFS Companion


juliaxbow:
Background:
Following efforts to track and obtain necessary metrics around IPFS usage, we must notify companion users about how data is being collected and used as well as easily allow for users to opt out of non-necessary metrics

**Proposed solution**
- One way to do this is to have a push notification in the companion extension. Chrome and Firefox have requirements for how these are formatted; see screen recording for proposal
- Pros: One-time alert that defaults to tracking user metrics; user would click on the alert or gear icon to go to preferences to update settings in preferences window. A push notification makes the change in how we collect metrics super visible.
- Cons: Users may have push notifications disabled through their browser. Push notifications are also generally found to be annoying

https://user-images.githubusercontent.com/58958327/207476491-33002444-5389-4c9b-a565-34cf820b9864.mov

**Alternative Solutions**

In all options, the actual toggles for cookie opt-ins/out live in preferences. The alternatives below describe methods of notifying the user. 

Description: Notification alert on gear icon for initial notice of cookie tracking. Clicking on this will bring the user to the user preferences, where they will see a section that allows them to update their cookie preferences. The notification will disappear after first click. 
![Gear Notification](https://user-images.githubusercontent.com/58958327/207475937-35cdf4b7-458a-44d4-b958-2af5d06e2ad5.png)

Description: Cookie link in extension window. Always include link at bottom of the extension window that will link to user preferences and allow the user to change cookie preferences. 
![IPFS Cookie](https://user-images.githubusercontent.com/58958327/207475934-19f9b07f-de84-4108-91e8-4dd3e3c834e8.png)

**Additional context**
- See https://github.com/ipfs/ipfs-gui/issues/105 for discussion around necessary metrics; language not yet formalized

lidel:
EU people are way pass the cookie-popup fatigue. Whatever we do should minimize the annoyance.

- "Notification alert on gear icon" sounds good
- Alternative idea is to add notice about "Opt-out telemetry"  to the welcome screen that is presented after initial install.  Either at the bottom, or as an additional section on the right (above or below  videos): 
  > ![2022-12-18_23-53](https://user-images.githubusercontent.com/157609/208323612-d9b719ee-d00c-462e-83c3-e3c9bc6180da.png)
 
 
 If you want to re-open the welcome screen, click on the cube in the left top corner of the Companion's browser action menu.

SgtPooki:
[...]

PR #1117, opened 2022-12-16
feat: add telemetry to companion

SgtPooki:
## Things implemented:

- [x] View tracking of companion pages
- [x] Session tracking (powering down the extension ends your session and stops tracking sessions)
- [x] consent management: disabling telemetry (opting-out) prevents the above two things from happening, but we do send an event indicating that you've opted-out

## Things to do
- [x] #1114
- [x] Fix CI failures
- [x] Need to discuss with @BigLep and @tinytb w.r.t. their discussion on 2023-01-05 whether #1114 needs to be a pre-req for merging this

image


SgtPooki:
@juliaxbow I would love your opinion on the formatting of the telemetry section

juliaxbow:
**Descriptions**
- **Telemetry section** In the description, would say "To assist in our efforts, please consent to the collection of these metrics by toggling "on" in the corresponding section. You can change this setting at any time by navigating back to the Companion Preferences page."  It's a little unusual to offer consent via a toggle but I think it's ok?
- Is the toggle default to on? If so, we should change the wording even further. Something like "IPFS Companion is collecting minimal telemetry data to improve and prioritize our work. You may opt out of the collection of these metrics by toggling "off" in the corresponding section below. You can change this setting ......"
- **Metrics Section** the metric collection description, I would change wording to "Sends ‘session’ and ‘view’ events only when ipfs-companion is enabled and your browser is not idle." Current wording raises the question of who is sending this information where? 
- The section header "minimal metrics: ..." makes it seem like there are additional, more intrusive metrics somewhere. Until we have other metrics we're tracking I think we can just call it "Session & Companion views." The Telemetry description already points to the fact that these are minimal. 

**Other**
- "Read More" links on the page have a link-out icon, so would add that in. (in the screenshot below i have the same link blue because I think all link-out icons should look like links but keep consistent with the rest of the page for now ;) )
- The toggle is centered but it doesn't look like it is. Until we have more sections & metrics that we're tracking, I would move it up to be top-aligned with the first sentence in the description. I know this is inconsistent with all other toggles on the page but to me it's more clear what the user is agreeing to. This is subjective, though, so feel free to ignore.

Incl screenshot to show the _suggested_ toggle alignment
Screen Shot 2022-12-16 at 8 07 57 PM

add-on/_locales/en/messages.json
```diff
@@ -726,5 +726,53 @@
...
+  "option_telemetryGroupMinimal_title": {
+    "message": "Minimal metrics: session & companion views",
+    "description": "A title for the 'minimal' grouping of metrics we collect (option_telemetryGroupMinimal_title)"
+  },
+  "option_telemetryGroupMinimal_description": {
+    "message": "We send `session` and `view` events only if ipfs-companion is enabled and your browser is not idle.",
+    "description": "A description for the 'minimal' grouping of metrics we collect (option_telemetryGroupMinimal_description)"
+  },
+  "option_telemetryGroupMinimal_session_description": {
+    "message": "A `session` event - every 60 seconds.",
```
lidel, line 747: Would this be better?
```suggestion
    "message": "A `session` event  for every minute the extension was active.",
```

add-on/_locales/en/messages.json
```diff
@@ -726,5 +726,53 @@
...
+    "description": "An explanation of the `session` telemetry we collect (option_telemetryGroupMinimal_session_description)"
+  },
+  "option_telemetryGroupMinimal_view_description": {
+    "message": "A `view` event - only for ipfs-companion pages.",
```
lidel, line 751: Users have no idea browser extensions are regular webpages behind the scenes.
Maybe more explicit:

```suggestion
    "message": "A `view` event for every time a part of extension's user interface is visible.",
```

add-on/src/lib/options.js
```diff
@@ -31,7 +33,11 @@ export const optionDefaults = Object.freeze({
   importDir: '/ipfs-companion-imports/%Y-%M-%D_%h%m%s/',
   useLatestWebUI: false,
   dismissedUpdate: null,
-  openViaWebUI: true
+  openViaWebUI: true,
+  telemetryGroupMinimal: true,
+  telemetryGroupMarketing: false,
+  telemetryGroupPerformance: false,
+  telemetryGroupTracking: false
```
lidel, line 40: - rename `telemetryGroupMinimal` to `telemetryFeatures`
- remove `telemetryGroupMarketing` and `telemetryGroupTracking` (we are not doing these, afaik don't plan to do it, so no reason to trigger anyone with this language)

add-on/src/lib/telemetry.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const enabledConsentGroups = Object.keys(obj).filter(key => obj[key] === true)
+  console.log('enabledConsentGroups: ', enabledConsentGroups)
+  return enabledConsentGroups
+}
+function logConsent () {
+  console.log('checkConsent(\'minimal\'): ', getMetricsProviderInstance().checkConsent('minimal'))
+  console.log('checkConsent(\'marketing\'): ', getMetricsProviderInstance().checkConsent('marketing'))
+  console.log('checkConsent(\'performance\'): ', getMetricsProviderInstance().checkConsent('performance'))
+  console.log('checkConsent(\'tracking\'): ', getMetricsProviderInstance().checkConsent('tracking'))
```
lidel, line 34: Remove all `console.log` calls, use scoped `debug` logger instead – see `src/lib/on-installed.js`, we want something similar:

```js
import debug from 'debug'
const log = debug('ipfs-companion:telemetry')
log(`checkConsent(foo): ${bar}`)
```
```

**Review by lidel — CHANGES_REQUESTED** [review]
Looks sensible, as long we address below:
1. We should not be sending requests to Countly every minute a user uses a browser. 
   (I assume that is what is happening? I was unable to build this and test locally)
   - This causes many issues, for example drains laptop battery by waking up WIFI when no internet conneciton would be needed otherwise. 
   - As discussed, figure a way to batch things so we report them only once an hour or once a day.  
1. Regular people don't know hat 'session' and 'view' means. 
   - See my suggestions below on humanizing the language :)

File: add-on/src/lib/notifier.js [hunk 5]
```diff
@@ -21,8 +21,8 @@ export default function createNotifier (getState) {
         return await browser.notifications.create({
           type: 'basic',
           iconUrl: browser.runtime.getURL('icons/ipfs-logo-on.svg'),
-          title: title,
-          message: message
+          title,
+          message
```
SgtPooki, line 25: Lint fixes

add-on/src/lib/options.js
```diff
@@ -1,8 +1,11 @@
 'use strict'
 
-import { isIPv4, isIPv6 } from 'is-ip'
 import isFQDN from 'is-fqdn'
+import { isIPv4, isIPv6 } from 'is-ip'
```
SgtPooki, line 4: Vscode ordered imports on save. Willing to remove

add-on/src/options/page.js
```diff
@@ -102,6 +103,13 @@ export default function optionsPage (state, emit) {
     logNamespaces: state.options.logNamespaces,
     onOptionChange
   })}
+  ${telemetryForm({
+    telemetryGroupMinimal: state.options.telemetryGroupMinimal,
+    telemetryGroupMarketing: state.options.telemetryGroupMarketing,
+    telemetryGroupPerformance: state.options.telemetryGroupPerformance,
+    telemetryGroupTracking: state.options.telemetryGroupTracking,
+    onOptionChange
+  })}
```
SgtPooki, line 112: @SgtPooki to update

add-on/src/types.d.ts
```diff
@@ -0,0 +1,52 @@
+
```
SgtPooki, line 1: Typescript types supported for linting. No work done for building validation (except from linter) nor exporting types (shouldn't be needed).

Errors show in vscode but don’t fail build yet

docs/telemetry/COLLECTED_DATA.md
```diff
@@ -0,0 +1,40 @@
+# Collected telemetry data
```
SgtPooki, line 1: @whizzzkid @lidel first demo of what COLLECTED_DATA.md should look like as defined from ipfs-shipyard/ignite-metrics#35.  Lmk what you think

SgtPooki (commented):
Self review

SgtPooki:
> We should not be sending requests to Countly every minute a user uses a browser.
> (I assume that is what is happening? I was unable to build this and test locally)

I just tested again, and since the “autoTrack” is set to false, countly does not send session_duration events anymore (those are sent by countly if you call `.track_sessions()`). See my comment about state.active and starting/ending sessions.

Also, i feel like we should so some implementation of observability on “idle” state. Any thoughts on changing the current use of “runOfNotIdle” to a boolean on the state object?  We could have a single interval that checks browser.idle.queryState and updates state.isIdle. Then check that variable prior to actions.

We could also add “runIfNotIdle” as a function to the state object so it’s re-usable. 

The main thing i want to stay away from is a pattern of multiple intervals scattered throughout the codebase, so lmk if you two have any thoughts here please!

SgtPooki:
@lidel @whizzzkid note that I pushed up a fix for the CI failures using the version of ignite-metrics at https://github.com/ipfs-shipyard/ignite-metrics/pull/60. will update package.json once I can use the version on npm again.

SgtPooki:
FYI @lidel @whizzzkid I had some changes locally that weren't pushed up... getting those up now

SgtPooki:
Agreed with @whizzzkid. Metrics are only sent (currently) when interacting with the extension (enabling/disabling, viewing different pages), so battery use should not be a concern.

Issue #778, opened 2019-10-02
Release Automation


lidel:
Things we could automate with CI or [Github Actions](https://github.com/features/actions):

- create `beta` and `release` branches 
- merge from `master` to the branch triggers build, signs & publishes beta release  to AMO and Chrome Web Store
  - caveat: firefox Beta is self-hosted and requires publishing `.xpi` + updating [this manifest](https://github.com/ipfs-shipyard/ipfs-companion/blob/master/ci/firefox/update.json)

Ref. https://github.com/kewisch/action-web-ext/

lidel:
We recently moved to Github Actions for running tests (#928), this would be the next step.
I believe these days we could match the level of relesse automation we have in ipfs-desktop:

1. publishing release builds and signs artifacts
2. artifacts are attached to the release
3. artifacts are published to extension web stores (Mozilla's AMO and Google's Chrome Web Store)

whizzzkid:
@lidel can we also make the following issues part of this?
- #839
- #1005 

Once we can unblock releases and have automation around the release processes we can around #666

meandavejustice:
I just added this for [🔗pin-tweet-to-ipfs](https://github.com/meandavejustice/pin-tweet-to-ipfs), [🔗Here is the file](https://github.com/meandavejustice/pin-tweet-to-ipfs/blob/main/.github/workflows/extension.yml) to take a look at with the actions. 

Here are some links to the actual actions used:
* [🔗build and sign for chrome](https://github.com/cardinalby/webext-buildtools-chrome-webstore-builder)
* [🔗publish for chrome](https://github.com/cardinalby/webext-buildtools-chrome-webstore-builder)
* [🔗publish for edge](https://github.com/wdzeng/edge-addon)

[🔗Here is a blog post](https://dev.to/cardinalby/publish-on-chrome-web-store-59oi) with all of the steps for each webstore.

lidel:
- Automating releases would be wonderful!
- Publishing to Edge is fine, as long it does not require any manual labor.  Would not include it in MVP tho, focus on existing stores (Firefox and Chrome).

PR #1122, opened 2022-12-20
feat: Adding Release Automation

whizzzkid:
Relates to: #778 

In this PR (Solves part 1 of #778):

- Once a PR is merged, a release PR is now automatically created.
- This release PR stays collecting new versions till we are ready to merge.
- On release PR merge, a new tag gets released.
- On a new tag push, we then create a `draft` release with:
  - New version tag.
  - Release Notes.
  - Assets downloaded from the previous step.

QoL improvements:
- The assets are now built and added to the draft release, the releaser only needs to download these assets and upload to respective web-stores.
- Future step would be to sign these assets and upload as part of CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,10 @@ jobs:
 
     steps:
       - name: Check out Git repository
-        uses: actions/checkout@v1
+        uses: actions/checkout@v3.3.0
 
       - name: Install Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3.6.0
         with:
           node-version: 18
 
@@ -39,7 +39,7 @@ jobs:
           npm prefix -g | % {npm config set node_gyp "$_\node_modules\node-gyp\bin\node-gyp.js"}
 
       - name: Restore .cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3.2.4
         id: cache
         with:
           path: ${{ github.workspace }}/.cache
@@ -50,7 +50,7 @@ jobs:
 
       - name: Restore node_modules
         id: yarn-cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3.2.4
         with:
           path: node_modules
           key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
@@ -84,7 +84,7 @@ jobs:
       # - action artifacts can be downloaded for 90 days, then are removed by github
       # - binaries in PRs from forks won't be signed
       - name: Attach produced packages to Github Action
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3.1.2
         with:
           name: built-on-${{ matrix.os }}
           path: build/ipfs_companion*.*
@@ -98,7 +98,7 @@ jobs:
     needs: [test]
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     steps:
-      - uses: google-github-actions/release-please-action@v3
+      - uses: google-github-actions/release-please-action@v3.7.3
         with:
           release-type: node
           package-name: ipfs-companion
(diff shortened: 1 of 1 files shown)


SgtPooki (approved):
this all looks legit to me and looks like a significant improvement

.github/workflows/ci.yml
```diff
@@ -92,3 +92,38 @@ jobs:
...
+          path: ~/downloads/
+
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: ${{ steps.download.outputs.download-path }}/*.zip
+          fail_on_unmatched_files: true
+          generate_release_notes: true
+          draft: true
+          append_body: true
```
lidel, line 128: `built-on-ubuntu-latest.zip`is nice,  but having to unpack it every time to get firefox or chrome specific version and then re-upload it to release is a bit tedious.

Mind adding a step that unzips it, so CI attach three `ipfs_companion*.zip` to the release, and person making the release does not need to unzip anything?

lidel (changes requested):
This is a very nice quality of life improvement!  Thank you @whizzzkid :heart: 

Before we merge, mind addressing two issues below?

.github/workflows/ci.yml
```diff
@@ -110,18 +110,42 @@ jobs:
     needs: [test]
     if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
     steps:
-      - uses: actions/checkout@v2
+      - name: Check out Git repository
+        uses: actions/checkout@v1
```
SgtPooki, line 114: why downgrading to checkout@v1?

.github/workflows/ci.yml
```diff
@@ -110,18 +110,42 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-${{ hashFiles('package*json', 'package-lock.json', '*config.js') }}
+            ${{ runner.os }}-
+
+      - name: Restore node_modules
+        id: yarn-cache
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
```
SgtPooki, line 136: just FYI: aegir actually has a great github action that allows us to cache node-modules. we should be using that most places to ensure we're consistent.

SgtPooki (approved):
approving the updates, but a note that I only saw the fix for the release-build comment lidel had. I didn't catch the three separate zips issue he was talking about so leaving that to @lidel to confirm

.github/workflows/ci.yml
```diff
@@ -92,3 +92,62 @@ jobs:
...
+
+      - name: Install dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: npm run ci:install
+
+      - name: Build Stable Release
+        run: npm run ci:build:stable
+
+      - name: Release
+        uses: softprops/action-gh-release@v1
```
lidel, line 146: :warning:  @whizzzkid  any reason why we are not reusing `google-github-actions/release-please-action` here with `command` set to `github-release`? 

Not a blocker, but `softprops/action-gh-release`  being a personal account of an individual has a higher risk of being compromised, and that could inject code that would ship to our users.

If we have to use it, please pin it to a specific revision before merging this PR:

```suggestion
        uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844
```

For future reference, remember to add to the safelist at https://github.com/ipfs/ipfs-companion/settings/actions:
> 2023-01-31-231838_819x604_scrot

lidel (approved):
@whizzzkid thank you!

Small concern around `softprops/action-gh-release` described below, but if you switch to  specific revision, it is  ok to merge.

.github/workflows/ci.yml
galargh: Other actions from the `actions` namespace also seem to be outdated. I'd recommend updating them all.

Issue #1120, opened 2022-12-19
Migrate from ipfs-http-client -> kubo-rpc-client


whizzzkid:
**Problem**
Based on @lidel's comment in #1118 companion needs to move away from ipfs-http-client and instead implement kubo-rpc-client which is more resilient because: https://github.com/ipfs/js-kubo-rpc-client/issues/103

**Additional context**
- #1118

PR #1124, opened 2022-12-22
feat: Moving from ipfs-http-client -> kubo-rpc-client

whizzzkid:
Closes: #1120 

In this PR:

- the build seems to work with a drop-in replacements
- removed other unused packages
- since we bundle this, we need this as a dev desp.


hacdias (approved):
I think this should be fine, as the API calls have not been changed.

PR #1136, opened 2023-01-23
fix: :bug: waiting for test to finish for longer

whizzzkid:
https://github.com/ipfs/ipfs-companion/actions/runs/3985067134


SgtPooki (approved):
approval, but also comment with concern

Issue #1138, opened 2023-01-27
Telemetry: Add Companion Version as a Segment


whizzzkid:
After #1117, companion now submits telemetry, but telemetry is missing the companion version for attribution.

For the first version it should not be a problem, but this should be added to the next version.

PR #1142, opened 2023-01-31
feat(telemetry): :sparkles: Adding companion version to segments

whizzzkid:
Closes: #1138 

In this PR:
- Reporting `ipfs-companion` version as a segment.

### DIFF
--- a/add-on/src/lib/ipfs-companion.js
+++ b/add-on/src/lib/ipfs-companion.js
@@ -179,7 +179,8 @@ export default async function init () {
 
   function onTelemetryMessage (request, sender) {
     if (request.trackView) {
-      return trackView(request.trackView)
+      const { version } = browser.runtime.getManifest()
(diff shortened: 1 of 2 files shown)


lidel (approved):
lgtm, but would like to get :+1: and thoughts on below questions from @SgtPooki  who has more holistic view of telemetry.

My questions (none of them blocking)
1. (could be separate PR) would it be useful to also read `arch` and `os` via [browser.runtime.getPlatformInfo](https://developer.chrome.com/docs/extensions/reference/runtime/#method-getPlatformInfo)?
1. (just cosmetics) Is is necessary (e.g., due to the way countly metrics work) to  to send this with every view event? (these values are static, doing it  once [during startup here](https://github.com/ipfs/ipfs-companion/blob/6430cf5dd92164c8887f138fe64fd4d6bf4a18bf/add-on/src/background/background.js#L16) would tell us the same)

PR #1158, opened 2023-02-15
fix(release-automation): :construction_worker: should no longer be needed

whizzzkid:
Removing bootstrap shas.


SgtPooki (approved):
still confused about the tags, but if we only have to manually create tags, i think that's a significant improvement.

Issue #1162, opened 2023-02-18
Redirect to local gateway not working


undyingwraith:
**Describe the bug**
All i get is the error page of my local node not being reachable (it is). Even clicking on redirect to public gateway results in the same thing.

**To Reproduce**
1. Navigate to 'ipfs.tech'
2. Observe the error page

**Expected behavior**
The site to load through my local gateway (ipfs.tech.localhost:8080)

**Desktop (please complete the following information):**
 - OS: Win 10
 - Browser: Firefox
 - Version: go-ipfs v0.13.0

**Additional context**
It works fine if i disable the integration and navigate to 'ipfs.tech.localhost:8080' directly.

Im currently copying a lot of files from the disk the ipfs repo is on, pausing it doesnt seem to influence the behaviour.

*Im currently not really well reachable so it might take some time till i can respond. If you need any additional info just ask.*

PR #1161, opened 2023-02-16
fix(release-automation): :green_heart: stupid hack to add tag in CI

whizzzkid:
I opened this issue: https://github.com/google-github-actions/release-please-action/issues/719

Not sure when will that be fixed, in the mean time, this ***might*** work.

PR #1125, opened 2022-12-23
feat: recover client if localhost is unreachable

whizzzkid:
Addresses: #1090 

In this PR:
- Recovery mechanism for URLs that are not reachable because the node is offline.
- Adds a button to redirect to public url.
- Refactor
- Recovery Page UX

Screenshot:

Recovery Process:

https://user-images.githubusercontent.com/1895906/210036280-0ff7c960-5148-4331-bb9d-6b262f1a0a07.mp4

Auto-Reload:

https://user-images.githubusercontent.com/1895906/210036416-4a2ccd5a-e2b7-4c16-9b7a-8538bce438d0.mp4

PR #1163, opened 2023-02-22
fix(recovery): :bug: false-positive for non-gateway URLs

whizzzkid:
Closes: #1162 

This is a case I should've thought of when working on: #1125. The reason this happens:

- When `Automatic Mode` is turned off, the redirect actually happens before the recovery mode begins.
- Since the redirect happens to local node, the recovery mode gets triggered, asking to redirect to public gateway.
- The public gateway redirects back to local and results in the endless loop.

Fix: <del>Add check for `!state.redirect`</del> remove false-positives from `sameGateway` check.

I would like to merge #1160 and #1161 before this and create a patch release.

### DIFF
--- a/test/functional/lib/ipfs-request-gateway-redirect.test.js
+++ b/test/functional/lib/ipfs-request-gateway-redirect.test.js
@@ -437,8 +437,15 @@ describe('modifyRequest.onBeforeRequest:', function () {
       state.pubGwURLString = 'https://ipfs.io'
       state.pubGwURL = new URL('https://ipfs.io')
     })
-    it('should present recovery page if node is offline', function () {
+    it('should present recovery page if node is offline and redirect is enabled', function () {
       expect(state.nodeActive).to.be.equal(false)
+      state.redirect = true
+      const request = url2request('https://localhost:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR/foo/bar')
+      expect(modifyRequest.onBeforeRequest(request).redirectUrl).to.equal('chrome-extension://testid/dist/recovery/recovery.html#https%3A%2F%2Fipfs.io%2Fipfs%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%2Ffoo%2Fbar')
+    })
+    it('should present recovery page if node is offline and redirect is disabled', function () {
+      expect(state.nodeActive).to.be.equal(false)
+      state.redirect = false
       const request = url2request('https://localhost:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR/foo/bar')
       expect(modifyRequest.onBeforeRequest(request).redirectUrl).to.equal('chrome-extension://testid/dist/recovery/recovery.html#https%3A%2F%2Fipfs.io%2Fipfs%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%2Ffoo%2Fbar')
     })


add-on/src/lib/ipfs-request.js
```diff
@@ -144,7 +144,11 @@ export function createRequestModifier (getState, dnslinkResolver, ipfsPathValida
 
       // When local IPFS node is unreachable , show recovery page where user can redirect
       // to public gateway.
-      if (!state.nodeActive && request.type === 'main_frame' && sameGateway(request.url, state.gwURL)) {
+      if (!state.nodeActive &&
+        !state.redirect &&
+        request.type === 'main_frame' &&
+        sameGateway(request.url, state.gwURL)
+      ) {
```
autonome, line 151: Please add a comment for each of the cases tested in that statement, so future generations can understand what's going on and why.

add-on/src/lib/ipfs-request.js
```diff
@@ -144,7 +144,11 @@ export function createRequestModifier (getState, dnslinkResolver, ipfsPathValida
 
       // When local IPFS node is unreachable , show recovery page where user can redirect
       // to public gateway.
-      if (!state.nodeActive && request.type === 'main_frame' && sameGateway(request.url, state.gwURL)) {
+      if (!state.nodeActive && //                  node is not active
+        !state.redirect && //                      this is not a redirect request
```
lidel: nit: this is a global flag that controls if companion is redirecting requests to local gateway, or not.
This was not fixing the problem, and the comment was only confusing the reader – I've restored old version and applied proper fix in `sameGateway`.

lidel (changes requested):
The problem with Companion is that you need to think with portals ;-) 

I dont believe this is fixing problem from https://twitter.com/unicomp21/status/1626244123102679041

We can clearly detect when someone opens URL equal to the Gateway `http://localhost:8080/` – I'll push fix shortly.

test/functional/lib/ipfs-path.test.js
```diff
@@ -184,11 +184,23 @@ describe('ipfs-path.js', function () {
       const gw = 'http://localhost:8080'
       expect(sameGateway(url, gw)).to.equal(true)
     })
-    it('should return true on 127.0.0.1/0.0.0.0 host match', function () {
+    it('should return true on localhost subdomain host match', function () {
+      const url = 'http://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi.ipfs.localhost:8080/foo/bar'
```
lidel, line 188: :information_source: we were missing subdomain test, so I've added it

test/functional/lib/ipfs-path.test.js
```diff
@@ -184,11 +184,23 @@ describe('ipfs-path.js', function () {
...
+    it('should return true on RPC webui path 127.0.0.1/0.0.0.0 host match', function () {
+      // this is misconfiguration, but handling it in sameGateway ensures users who do this dont waste our time debugging
       const url = 'http://0.0.0.0:5001/webui'
       const api = 'http://127.0.0.1:5001'
       expect(sameGateway(url, api)).to.equal(true)
     })
+    it('should return true on RPC /api/v0 path 127.0.0.1/0.0.0.0 host match', function () {
+      // this is misconfiguration, but handling it in sameGateway ensures users who do this dont waste our time debugging
+      const url = 'http://0.0.0.0:5001/api/v0/id'
+      const api = 'http://127.0.0.1:5001'
```
lidel, line 201: :information_source:  this saves us a day of debugging when user puts wrong port in wrong field in their config, and then tweets about it ;-)

test/functional/lib/ipfs-request-gateway-redirect.test.js
```diff
@@ -431,17 +431,38 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
+      const request = url2request('https://localhost:8080/')
+      expect(modifyRequest.onBeforeRequest(request)).to.equal(undefined)
+    })
+    it('should not show recovery page if extension is disabled', function () {
+      // allows user to quickly avoid anything similar to https://github.com/ipfs/ipfs-companion/issues/1162
+      state.active = false
+      expect(state.nodeActive).to.be.equal(false)
+      const request = url2request('https://localhost:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR/foo/bar')
+      expect(modifyRequest.onBeforeRequest(request)).to.equal(undefined)
+    })
```
lidel, line 465: :information_source:  I think this covers all edge cases, and ensures recovery logic does not depend on redirect state (which was why it was initially difficult to reproduce the bug from https://github.com/ipfs/ipfs-companion/issues/1162)

lidel (approved):
@whizzzkid pushed alternative fix for `sameGateway` check and some tests.
This approach should be more reliable.

add-on/src/lib/ipfs-request.js
```diff
@@ -144,7 +144,11 @@ export function createRequestModifier (getState, dnslinkResolver, ipfsPathValida
...
+        request.type === 'main_frame' && //        this is a request for a root document
+        sameGateway(request.url, state.gwURL) //   this is a request to the local gateway
```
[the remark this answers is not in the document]
lidel: `state` is global, not per request. it informs redirect logic what are current settings and state of RPC. 
(bit of a legacy thing, from the XUL time (before v2) when it was less expensive to cache options in one place, but it is still useful when writing tests, to simulate specific state of world).

you probably saw us setting it in tests per different requests to simultate specific  flows, and assumed it is per request, but it is is not :)

Issue #1169, opened 2023-03-01
Implement MV3 Checklist


whizzzkid:
This issue tracks implementation of: https://developer.chrome.com/docs/extensions/mv3/mv3-migration/

ETA: 2023-03-30

### Tasks
- [x] Migrate `manifest_version: 2` -> `manifest_version: 3`
- [x] Migrate `background.page` -> `background.service_worker`
- [x] https://github.com/ipfs/ipfs-companion/issues/1171
  - [x] Implement a new build step to build `service_worker.js` instead of `background.html`
  - [x] Remove instances of `window`.
  - [x] Migrate instances of `localStorage` -> `browser.storage`.
- [x] Migrate `web_accessible_resources` -> url specific resources.
- [x] Migrate `content_security_policy` -> `content_security_policy.extension_pages`
- [x] Migrate `browser_action`
- [x] Migrate away from `eval`
- [x] Migrate `chrome.contextMenus.create` -> `chrome.contextMenus.onClicked`
- [x] `fetch` with cors enabled: https://github.com/ipfs/kubo/pull/8690
- [x] redirects with `Access-Control-Allow-Origin`: https://github.com/ipfs/kubo/issues/9728

PR #1170, opened 2023-03-03
feat(mv3): Manifest V3 Migration Checklist

whizzzkid:
Closes: #1169

### DIFF
--- a/add-on/src/lib/ipfs-companion.js
+++ b/add-on/src/lib/ipfs-companion.js
@@ -517,7 +517,7 @@ export default async function init (windowedContext = false) {
       // Try SVG first -- Firefox supports it natively
       await browser.action.setIcon(iconDefinition)
       if (browser.runtime.lastError.message === 'Icon invalid.') {
-        throw new Error('Icon invalid.')
+        throw browser.runtime.lastError 
       }
     } catch (error) {
       // Fallback!


add-on/manifest.common.json
```diff
@@ -29,15 +30,24 @@
...
+        "icons/ipfs-logo-on.svg",
+        "icons/ipfs-logo-off.svg",
+        "dist/recovery/recovery.css",
+        "dist/recovery/recovery.html",
+        "dist/recovery/recovery.js"
+      ],
+      "matches": [
+        "<all_urls>"
+      ]
+    }
```
SgtPooki, line 47: I took a look at https://developer.chrome.com/docs/extensions/mv3/manifest/web_accessible_resources/ and the first line 

> Web-accessible resources are files inside an extension that can be accessed by web pages or other extensions. Extensions typically use this feature to expose images or other assets that need to be loaded in web pages, but any asset included in an extension's bundle can be made web accessible.

It appears that this is needed so that webpages can test if companion is being used, and if so, they can load recovery.html/js/css. Is that correct? are there other usecases where this is needed?

add-on/src/lib/context-menus.js
```diff
@@ -71,29 +71,13 @@ export function createContextMenus (
   getState, _runtime, ipfsPathValidator, { onAddFromContext, onCopyRawCid, onCopyAddressAtPublicGw }) {
   try {
     const createSubmenu = (id, contextType, menuBuilder) => {
-      browser.contextMenus.create({
-        id,
-        title: browser.i18n.getMessage(id),
-        documentUrlPatterns: ['<all_urls>'],
-        contexts: [contextType]
-      })
+      browser.contextMenus.onClicked.addListener((...args) => console.log(args))
```
SgtPooki, line 74: do we still need this?

add-on/src/popup/browser-action/store.js
```diff
@@ -205,6 +207,7 @@ export default (state, emitter) => {
     const prev = state.active
     state.active = !prev
     if (!state.active) {
+      endSession()
```
SgtPooki, line 210: if we're only calling startSession and endSession inside this DOMContentLoaded for browser-action now, then I don't think we're going to be tracking sessions properly anymore. Is that blocked by https://github.com/ipfs-shipyard/ignite-metrics/issues/103 ?

add-on/src/lib/ipfs-companion.js
```diff
@@ -473,7 +459,7 @@ export default async function init () {
   // -------------------------------------------------------------------
 
   async function updateBrowserActionBadge () {
-    if (typeof browser.browserAction.setBadgeBackgroundColor === 'undefined') {
+    if (typeof browser.action.setBadgeBackgroundColor === 'undefined') {
       // Firefox for Android does not have this UI, so we just skip it
       return
     }
```
SgtPooki, line 465: It looks like this is supported in firefox for android now, https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/setBadgeBackgroundColor#browser_compatibility, so we can remove the content, but we may still want to double check for the existence of this method

add-on/src/lib/ipfs-companion.js
```diff
@@ -525,14 +511,17 @@ export default async function init () {
     let iconDefinition = { path: iconPath }
     try {
       // Try SVG first -- Firefox supports it natively
-      await browser.browserAction.setIcon(iconDefinition)
+      await browser.action.setIcon(iconDefinition)
+      if (browser.runtime.lastError.message === 'Icon invalid.') {
+        throw new Error('Icon invalid.')
```
SgtPooki, line 516: can we just do 
```suggestion
      if (browser.runtime.lastError.message === 'Icon invalid.') {
        throw browser.runtime.lastError 
```
?

SgtPooki (changes requested):
mostly comments. minor changes requested / proposed

SgtPooki (approved):
LGTM.. some of the test files I skimmed over.. that's a lot of awaiting..

PR #1172, opened 2023-03-07
feat(telemetry): Refactor Metrics Tracking from background service_worker

whizzzkid:
Relates to: #1169 

Since MV3 does not have background page context and countly sdk does not build for webworker context. This PR aims to refactor that.

- Initialization is based on cookies, so start/end session is now handled on the options page.
- View Tracking happens on the pages with window context.
- Opened: https://github.com/Countly/countly-sdk-web/issues/382

This merges into the parent PR which includes other MV3 checklist items.


SgtPooki (approved):
Understood that we don't need this unless https://github.com/ipfs-shipyard/ignite-metrics/issues/103 isn't done in time, but code lgtm and I want to unblock you

PR #1179, opened 2023-03-21
feat(mv3): ✨ XHR to Fetch Migration

whizzzkid:
Fixes: https://developer.chrome.com/docs/extensions/migrating/to-service-workers/#replace-xmlhttprequest

The base is set to: https://github.com/ipfs/ipfs-companion/tree/fix/webpack which merges into RC.


SgtPooki (approved):

add-on/src/lib/ipfs-request.js
```diff
@@ -138,14 +138,14 @@ export function createRequestModifier (getState, dnslinkResolver, ipfsPathValida
     // browser.webRequest.onBeforeRequest
     // This event is triggered when a request is about to be made, and before headers are available.
     // This is a good place to listen if you want to cancel or redirect the request.
-    onBeforeRequest (request) {
+    async onBeforeRequest (request) {
```
lidel, line 141: :information_source:  Very happy to see this possible in Chromium :))

`async` here was possible from Firefox 52 onwards, but Chromium did not support promises for years, and we were stuck with all the blocking XHR code here.

lidel (approved):
Just dropping :heart:  as I know how time-consuming this refactors like this one are.

Issue #1156, opened 2023-02-15
Implement blocking via `declarativeNetRequest` API


whizzzkid:
Parent Issue: https://github.com/ipfs/ipfs-companion/issues/1152
~~Blocked By: https://github.com/ipfs/ipfs-companion/issues/1155~~

ETA: 2023-05-05

```[tasklist]
### Tasks
- [x] Implement `declarativeNetRequest` to update URL blocking based on populated LFU.
- [x] Route blocked requests conventionally.
```

SgtPooki:
@BigLep this is the issue tracking the user-value benefits of ipfs-companion once mv3 is done

BigLep:
@whizzzkid : what is meant by "Route blocked requests conventionally"?  Is that the end of the done criteria?

PR #1181, opened 2023-03-23
feat(mv3): blocking by observing

whizzzkid:
Closes: #1156 

This PR introduces the MV3 block and redirect using [`declarativeNetRequest`](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/)

Also:
- Added TS support, now we can build using TS code.
- block and observe from code.
- Redirect requests based on origin.
- Recovery logic fixes.

### DIFF
--- a/add-on/src/lib/ipfs-request.js
+++ b/add-on/src/lib/ipfs-request.js
@@ -1,4 +1,3 @@
-// @ts-nocheck
 'use strict'
 /* eslint-env browser, webextensions */
 


add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  // We can now construct the regex filter and substitution.
+  const regexSubstitution = redirectUrl.slice(0, redirectUrl.length - commonIdx + 1) + '\\1'
+  // We need to escape the characters that are allowed in the URL, but not in the regex.
+  const regexFilterFirst = `${originUrl.slice(0, originUrl.length - commonIdx + 1).replace(ALLOWED_CHARS_URL_REGEX, '\\$1')}`
+  // We need to match the rest of the URL, so we can use a wildcard.
+  const regexFilter = `^${regexFilterFirst}(.*)$`
+
+  return { regexSubstitution, regexFilter }
+}
```
SgtPooki, line 46: we need some heavy tests added around this function

add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -0,0 +1,124 @@
...
+
+// We need to check if the browser supports the declarativeNetRequest API.
+export const supportsBlock = !(browser.declarativeNetRequest?.MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES === 5000)
```
SgtPooki, line 49: do we need to check that the value is hardcoded to 5000? can we check that the value is defined to something instead? it should be more future proof

SgtPooki (changes requested):
mostly minor changes requested

add-on/src/lib/redirect-handler/blockOrObserve.ts
SgtPooki: My original comment here still applies. If chrome updates their value from 5000, this will fail. Do we want to have to keep an eye on their value, or can we make this check more robust?

SgtPooki (approved):

PR #1186, opened 2023-04-04
feat(mv3): adding dynamicNetRequest rule reconciliation logic + Firefox Builds

whizzzkid:
- Closes: #1191 
- Closes: #1156

### In this PR:
- [x] Added firefox background event page asset step.
- [x] Added reconciliation logic, for when the background page/service worker goes to sleep and wakes up, it needs to populate in-memory rules and reconcile stale rules.
- [x] Implemented a mock for `declarativeNetRequest` because https://github.com/acvetkov/sinon-chrome/issues/110
  - This implementation actually emulates the full API behaviour available in the browsers.
- [x] Implemented a few tests, working on more.
- [x] Killed the debug patch because it was breaking windows builds and it works like: https://github.com/debug-js/debug/issues/916
- [x] Refactoring, lots of it!

### Testing:
- After `npm i` rituals, build:
  - Chromium: `npm run build && npm run bundle:chromium`
  - Firefox: `npm run build && npm run bundle:firefox`
- Make config changes in `kubo`/`ipfs-desktop` to allow connections from the extension URLs.
- Chrome (new `declarativeNetRequest` API)
  - Just drag and drop the extension.
  - Visit any ipfs-enabled page, e.g: ipfs.io/ipns/en.wikipedia-on-ipfs.org or https://awesome.ipfs.io/
- Firefox (old blocking request)
  - You'll need firefox developer edition, https://www.mozilla.org/en-US/firefox/developer/
  - Firefox issues a temp id, use that id to configure kubo.
  - Make sure to turn on `Access your data for all websites` under optional permissions. This is a new thing in MV3 as `<all_urls>` from `permissions` are gone, instead `host_permissions` are now declaring that access to `<all_urls>`. Users will be notified on update and install regarding access to all websites and users will enable it then.
  - This is how it looks like:
  - Screenshot 2023-05-04 at 4 55 17 PM

### DIFF
--- a/test/functional/lib/redirect-handler/blockOrObserve.test.ts
+++ b/test/functional/lib/redirect-handler/blockOrObserve.test.ts
@@ -50,7 +50,6 @@ describe('lib/redirect-handler/blockOrObserve', () => {
       expect(isLocalHost('http://localhost')).to.be.true
       expect(isLocalHost('https://google.com')).to.be.false
       expect(isLocalHost('https://ipfs.io')).to.be.false
-      expect(isLocalHost('ipfs://QmHash')).to.be.false
     })
   })
 


SgtPooki:
https://github.com/ipfs/ipfs-companion/pull/1181 is merged now, is this still blocked, or just parked for now?

add-on/manifest.firefox-beta.json
```diff
@@ -2,7 +2,7 @@
   "browser_specific_settings": {
     "gecko": {
       "id": "ipfs-companion-dev-build@ci.ipfs.team",
-      "update_url": "https://ipfs-shipyard.github.io/ipfs-companion/ci/firefox/update.json"
```
SgtPooki, line 5: do we want to remove the update_url?

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -0,0 +1,100 @@
...
+})
+
+describe('lib/redirect-handler/blockOrObserve', () => {
+  before(function () {
+    browserMock.runtime.id = 'testid'
+  })
+
+  describe('isLocalHost', () => {
+    it('should return true for localhost', () => {
+      expect(isLocalHost('http://localhost:8080')).to.be.true
```
SgtPooki, line 41: we should add a test to ensure regex termination is working properly

```suggestion
      expect(isLocalHost('http://localhost')).to.be.true
      expect(isLocalHost('http://localhost:8080')).to.be.true
```

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -0,0 +1,100 @@
...
+      expect(isLocalHost('http://localhost:8080/ipfs/QmHash')).to.be.true
+      expect(isLocalHost('http://127.0.0.1:8080/ipns/QmHash')).to.be.true
+      expect(isLocalHost('http://[::1]:8080/ipfs/QmHash')).to.be.true
+    })
```
SgtPooki, line 45: How do we know it doesn't return true for everything? =P

```suggestion
    })
    it('should return false for non-localhost', () => {
      expect(isLocalHost('http://dev.localhost:8080')).to.be.false
      expect(isLocalHost('http://ipfs.tech')).to.be.false
    })
    it('should return false for unsupported protocols', () => {
      expect(isLocalHost('ipfs://QmHash')).to.be.false
      expect(isLocalHost('https://localhost')).to.be.false
    })
    
```

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    it('Should add redirect rules for local gateway', () => {
+      addRuleToDynamicRuleSet({
+        originUrl: 'https://ipfs.io/ipns/en.wikipedia-on-ipfs.org',
+        redirectUrl: 'http://localhost:8080/ipns/en.wikipedia-on-ipfs.org'
+      })
+      expect(browserMock.declarativeNetRequest.updateDynamicRules.called).to.be.true
+      const [{ addRules, removeRuleIds }] = browserMock.declarativeNetRequest.updateDynamicRules.firstCall.args
+      // this is needed for the reconciliation logic to work. This is tested separately.
```
SgtPooki, line 90: > This is tested separately.

Can you point me (or future comment reader) to where?

webpack.config.js
```diff
@@ -165,6 +165,20 @@ const bgConfig = merge(commonConfig, {
...
+const bgFirefoxConfig = merge(commonConfig, {
+  name: 'background-firefox',
+  entry: {
+    backgroundPage: './add-on/src/background/background.js'
+  },
+  output: {
+    filename: '[name].firefox.bundle.js'
+  }
+})
+
```
SgtPooki, line 181: why do we need a separate bundle for firefox?

SgtPooki (changes requested):
Looks really good! I left a few comments for things that could be improved slightly

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      expect(isLocalHost('http://ipfs.tech')).to.be.false
+      expect(isLocalHost('http://localhost:8080')).to.be.true
+      expect(isLocalHost('http://localhost:8080')).to.be.true
+      expect(isLocalHost('http://localhost:8080')).to.be.true
+      expect(isLocalHost('http://localhost:8080/ipfs/QmHash')).to.be.true
+      expect(isLocalHost('http://localhost:8080/ipfs/QmHash')).to.be.true
+      expect(isLocalHost('http://localhost')).to.be.true
+      expect(isLocalHost('https://google.com')).to.be.false
+      expect(isLocalHost('https://ipfs.io')).to.be.false
+      expect(isLocalHost('ipfs://QmHash')).to.be.false
```
SgtPooki, line 53: nit: technically violating the test name

SgtPooki (approved):

Issue #1199, opened 2023-05-16
MV3 Beta 3.0.0.1092 @ Brave/Chrome: Turning off companion/Enabled for button does not remove redirection


whizzzkid:
Context: https://github.com/ipfs/ipfs-companion/pull/1182#issuecomment-1546306896

- Turning off IPFS integration (main switch) does not remove redirection rules.
- Disabling for current tab also does not remove redirection rules.

PR #1200, opened 2023-05-17
feat(mv3): Handle State Changes

whizzzkid:
Closes: #1199 

In this PR:
- handling logic for state changes from on -> off and vice-versa.
- handling logic for option changes like gateway url.


SgtPooki (commented):
we should have a test for this

PR #1210, opened 2023-05-26
fix(mv3): 👔 Adding better regex replace to remove infinite redirects.

### LINKED ISSUE #1209: [MV3 Beta Bug] Local Redirects Should not generate infinite redirect regex
Under certain circumstances a regex pattern can exist which catches all the URLs and redirects requests till browser crashes.

It's handled here: https://github.com/ipfs/ipfs-companion/blob/rc/3.0-mv3/add-on/src/lib/redirect-handler/blockOrObserve.ts#L129-L133

however this does not work for `http` urls and only for `https`

whizzzkid:
Fixes #1209 

In this PR:
- A weird edge case causes redirects from `http` to ipns causes infinite redirects because it creates a catch-all rule which looks like `^https?\\:\\/((?:[^\\.]|$).*)$` and that redirects everything to new pattern so the result is `http://localhos:8080/http://localhost:8080...`
- Adding more tests made me realize the rules already in memory were not being flushed properly so fixed that.

### DIFF
--- a/add-on/src/lib/redirect-handler/blockOrObserve.ts
+++ b/add-on/src/lib/redirect-handler/blockOrObserve.ts
@@ -349,7 +349,7 @@ export function addRuleToDynamicRuleSetGenerator (
       )
 
       // refresh the tab to apply the new rule.
-      const tabs = await browser.tabs.query({url: `${originUrl}*`})
+      const tabs = await browser.tabs.query({ url: `${originUrl}*` })
       await Promise.all(tabs.map(async tab => await browser.tabs.reload(tab.id)))
     }
 


test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -1,11 +1,11 @@
-import { expect } from 'chai'
-import { before, describe, it } from 'mocha'
+import {expect} from 'chai'
+import {before, describe, it} from 'mocha'
```
SgtPooki, line 2: Arent these style changes the opposite of aegir/standardjs?

SgtPooki (approved):

Issue #1212, opened 2023-05-31
[MV3 Beta Bug] Context Menus Don't Work


whizzzkid:
While migrating the contextMenus were not working as expected, so a placeholder function was added, this needs to be fixed.

PR #1213, opened 2023-05-31
feat(mv3): Adding ContextMenus MV3 Style

whizzzkid:
Closes: #1212 

In this PR:
- Adding ContextMenu functionality
- This got lost in the transition to MV3
- The reason this is different now is, there are no more individual event handler, everything now goes through a global event listener or multiple global listners which decide what action needs to happen.

### DIFF
--- a/add-on/src/lib/context-menus/ContextMenus.ts
+++ b/add-on/src/lib/context-menus/ContextMenus.ts
@@ -52,7 +52,11 @@ export class ContextMenus {
    * @param cb
    */
   create (options: browser.Menus.CreateCreatePropertiesType, cb?: listenerCb): void {
-    browser.contextMenus.create(options)
+    try {
+      browser.contextMenus.create(options)
+    } catch (err) {
+      this.log.error?.('ContextMenus.create failed', err)
+    }
     if (cb != null) {
       if (options?.id != null) {
         this.queueListener(options.id, cb)
--- a/test/functional/lib/context-menus/ContextMenus.test.ts
+++ b/test/functional/lib/context-menus/ContextMenus.test.ts
@@ -75,7 +75,7 @@ describe('lib/context-menus/ContextMenus', () => {
     expect(listenerSpy.called).to.be.true
   })
 
-  it('should create and queue multiple listeners for same menuItemId', () => {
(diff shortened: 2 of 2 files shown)


add-on/src/lib/context-menus.js
```diff
@@ -66,30 +67,54 @@ const apiMenuItemIds = new Set([contextMenuCopyRawCid, contextMenuCopyCanonicalA
...
-        (context) => handler(context, contextType)
-      )
+      contextMenus.create({
+        id: itemId,
+        parentId,
+        title: browser.i18n.getMessage(id),
+        contexts: [contextType],
+        documentUrlPatterns: [
+          '*://*/ipfs/*', '*://*/ipns/*',
+          '*://*.ipfs.dweb.link/*', '*://*.ipns.dweb.link/*', // TODO: add any custom public gateway from Preferences
```
SgtPooki, line 112: We should make an issue from this todo if we don't have one

add-on/src/lib/context-menus/ContextMenus.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.log(`ContextMenus Listener queued for ${menuItemId}`)
+  }
+
+  /**
+   * This method creates a context menu item and maps the listener function to it.
+   *
+   * @param options
+   * @param cb
+   */
+  create (options: browser.Menus.CreateCreatePropertiesType, cb?: listenerCb): void {
```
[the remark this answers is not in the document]
SgtPooki: nit: `CreateCreatePropertiesType` name is odd

also, why cb as a second parameter instead of part of the create options?

add-on/src/lib/context-menus/ContextMenus.ts
```diff
@@ -0,0 +1,68 @@
...
+    try {
+      browser.contextMenus.create(options)
+    } catch (err) {
+      this.log.error?.('ContextMenus.create failed', err)
```
SgtPooki, line 58: error is defined definitively in the constructor

```suggestion
      this.log.error('ContextMenus.create failed', err)
```

SgtPooki (approved):

Issue #1204, opened 2023-05-19
[MV3 Beta Bug] Automatic mode does not work automatically on Brave or Chrome


whizzzkid:
Gateways only get invoked when the first-class "IPFS” button in the address bar is hit manually.

BigLep:
@whizzzkid : my feedback on this was under the heading "Experience with fresh install on Brave without using Kubo (neither Brave-managed Kubo or external Kubo)"

I'm intentionally not using Kubo.

https://www.notion.so/pl-strflt/Dogfooding-4560b1a6cd7a494ca5ac5b1bc7d6cd4a?pvs=4#a39d3a036f494c9c8427efd739a1ce19

BigLep:
@whizzzkid : when is Automatic mode supposed to work.  

From the settings menu it says "Automatically switch from your local gateway to your default public gateway if the Kubo RPC is unavailable."

If I don't have a Kubo node running (and thus not Kubo RPC endpoint and no local gateway) visiting sites like ipfs.tech don't redirect to my external gateway (e.g., ipfs.io).  I would expect when visiting ipfs.tech to be redirected to https://ipfs-tech.ipns.dweb.link/.

I can confirm this doesn't appear to work with "IPFS Companion RC-MV3 (Beta @ 9f4981c)" on both Chrome and Brave.

BigLep:
Thanks @whizzzkid .  I was able to ask @lidel some more companion questions during a 1 on 1.

I'm still struggling with this option to make sure I understand what it does.

If I use companion, have "automatic mode" enabled, and visit http://docs.ipfs.tech.ipns.localhost:8080/, I don't get automatically redirected.  I get a redirect indirect I need to interact with (which make sense from a security regard) (see below).  That said, I see this interstitial regardless of whether I have "automatic mode" enabled or not.

I am wanting to make sure I understand what "Use Automatic Mode - Automatically switch from your local gateway to your default public gateway if the Kubo RPC is unavailable" means. When does this "automatic mode" get invoked?  What is the difference in behavior between enabling it vs. disabling it?

<img width="1180" alt="image" src="https://github.com/ipfs/ipfs-companion/assets/85411/47369887-93ef-4364-b779-046b94b0f773">

lidel:
Some historic notes:
-  "Automatic Mode" was introduced to disable redirect to local gateway when it was down, and re-enable it when it gets back online.  
[...]

PR #1223, opened 2023-06-20
fix(recovery): :adhesive_bandage: automatic mode

whizzzkid:
Fixes: #1204

### DIFF
--- a/add-on/src/lib/redirect-handler/blockOrObserve.ts
+++ b/add-on/src/lib/redirect-handler/blockOrObserve.ts
@@ -320,10 +320,13 @@ export function addRuleToDynamicRuleSetGenerator (
     const state = getState()
     const redirectIsOrigin = originUrl === redirectUrl
     const redirectIsLocal = isLocalHost(originUrl) && isLocalHost(redirectUrl)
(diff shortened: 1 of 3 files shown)


add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -320,10 +320,13 @@ export function addRuleToDynamicRuleSetGenerator (
     const state = getState()
     const redirectIsOrigin = originUrl === redirectUrl
     const redirectIsLocal = isLocalHost(originUrl) && isLocalHost(redirectUrl)
-    const badOriginRedirect = originUrl.includes(state.gwURL.host) && !redirectUrl.includes('recovery')
+    const localOriginRedirect = isLocalHost(originUrl) && !isLocalHost(redirectUrl)
+    // if the redirect is to the gateway and the automatic mode is off, we would like to present the recovery page.
+    // in case automatic mode is on, we would like to redirect to the public gateway.
```
lidel, line 325: I don't think the status of automatic mode should impact/bypass the recovery page – we don't want to blindly redirect users to third-party server – see https://github.com/ipfs/ipfs-companion/issues/1204#issuecomment-1599105933

Issue #1226, opened 2023-06-26
Automatic Config Migration MV2 -> Mv3


whizzzkid:
When Mv2 companion will be replaced with mv3 companion, the configs will not automatically migrate. The reason is the `serviceWorker` context does not allow access to mv2 `localStorage` (html5 storage) instead the new way is `chrome.storage.local` (storage specifically designed for extensions) which will set the configs to default on upgrade.

To make the transition smoother for current users:

```[tasklist]
### Tasks
- [x] Show an upgraded screen on the first upgrade from MV2 -> MV3 (this is not liked by our users and was removed in https://github.com/ipfs/ipfs-companion/issues/933)
- [x] Make the screen show migration info or just a notice that it's a one time thing.
- [x] Use the window context from this upgrade screen to access `window.localStorage`
- [x] Post this to the background service worker and save it in `chrome.storage.sync` or `chrome.storage.local`
- [ ] Validate and write tests.
```

PR #1228, opened 2023-07-01
feat(mv3): Config Migration

whizzzkid:
Closes: #1226

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,7 +2,11 @@ name: Close and mark stale issue
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


add-on/src/lib/on-installed.js
```diff
@@ -28,9 +28,23 @@ export async function runPendingOnInstallTasks () {
         url: welcomePage
       })
     case 'onVersionUpdate':
-      if (!displayReleaseNotes) return
+      // temporary disabled due to upgrade to MV3.
```
SgtPooki, line 31: Do we want to double check the version to ensure mv3 update is what's needed?

PR #1230, opened 2023-07-05
Fix(docs): ✏️ Adding/Updating docs

### LINKED ISSUE #1190: Add MV3 Working Architecture
ETA: 2023-05-31

```[tasklist]
### Tasks
- [x] Add working diagrams for request routing.
- [x] Explain and add documentation on how migration works.
```

### LINKED ISSUE #1207: Fix Companion docs
Fix: https://docs.ipfs.tech/how-to/companion-node-types/#external

- [x] Clarify External means kubo.
- [x] Remove Embedded till ready to be replaced via helia.
- [x] Fix: https://github.com/ipfs/ipfs-docs/blob/main/docs/install/ipfs-companion.md

whizzzkid:
- Closes #1207 
- Closes #1190

### DIFF
--- a/docs/MV3.md
+++ b/docs/MV3.md
@@ -48,7 +48,7 @@ The process is asynchronous, the browser allows "observation" of requests to com
 }
 ```
 
-The API only allows for 5000 such rules to exist at the time of writing this, so that value can be exhausted pretty quickly. To save the rules and gain more coverage of such serviceable URLs, we make use of the fact that rules allow for regex substitutions. So instead of adding one rule per redirect, we can dynamically generate a regex pattern to suffice redirection.
+The API only allows for 5000 such rules to exist at the time of writing this, so that value can be exhausted pretty quickly. So instead of adding one rule per redirect, we can dynamically generate a single regex pattern to handle multiple redirections.
 
 For example
 


docs/MV3.md
```diff
@@ -0,0 +1,114 @@
...
+
+![MV2 Working Diagram](assets/mv2-working.png)
+
+The process was simple and synchronous, the request is intercepted and handed over to companion, which then redirects it to the local gateway as needed. MV3 changes this process by introducing a new API called `declarativeNetRequest`. MV3 does not allow synchronous request interception (also called blocking request), which is not secure and can lead to performance issues. Instead, it uses a declarative approach where the extension declares a set of rules that are then used by the browser to intercept requests. This process looks something like:
+
+![MV3 Working Diagram](assets/mv3-working.png)
+
+The process is asynchronous, the browser allows "observation" of requests to companion, which asynchronously determines if the given request is serviceable by IPFS and then dynamically introduces a rule for the browser to perform redirects to the local gateway. A sammple rule looks something like:
```
SgtPooki, line 17:
```suggestion
The process is asynchronous, the browser allows "observation" of requests to companion, which asynchronously determines if the given request is serviceable by IPFS and then dynamically introduces a rule for the browser to perform redirects to the local gateway. A sample rule looks something like:
```

docs/MV3.md
```diff
@@ -0,0 +1,114 @@
...
+      'webbundle',
+      'websocket',
+      'webtransport',
+      'xmlhttprequest'
+    ]
+  }
+}
+```
+
+The API only allows for 5000 such rules to exist at the time of writing this, so that value can be exhausted pretty quickly. To save the rules and gain more coverage of such serviceable URLs, we make use of the fact that rules allow for regex substitutions. So instead of adding one rule per redirect, we can dynamically generate a regex pattern to suffice redirection.
```
SgtPooki, line 51:
```suggestion
The API only allows for 5000 such rules to exist at the time of writing this, so that value can be exhausted pretty quickly. So instead of adding one rule per redirect, we can dynamically generate a single regex pattern to handle multiple redirections.
```

SgtPooki (approved):
I added Some suggestions where text was too verbose or confusing, but otherwise looks good

docs/MV3.md
```diff
@@ -0,0 +1,114 @@
...
+}
+```
+
+This is a single rule that covers all the redirects from `ipfs.io` to `localhost:8080`. This means that we can save a lot of rules and gain more coverage of serviceable URLs. This is the approach we have taken in MV3.
+
+## Gotchas
+
+Since the process is asynchronous, there are a few things that we need to keep in mind:
+
+- There will always be scenarios, where the redirection to public URL happens before companion determines if the URL is serviceable (consider ipns/fqdn resolutions.) To tackle this a refresh mechanism has been put in place that refreshes the page after a few seconds if the URL is serviceable. This is not ideal, but it is the best we can do for now.
```
BigLep, line 107: > There will always be scenarios, where the redirection to public URL happens before companion determines if the URL is serviceable

Adding the word "redirection" seems odd here.
Maybe "where the servicing of public URLs happen..."

docs/MV3.md
```diff
@@ -0,0 +1,114 @@
...
+- Similarly there are scenarios around page recovery, where the page starts redirection, but the kubo node goes offline, in which case companion needs to remove offending rule and replace it with the recovery URL.
+- If the kubo rpc api url changes for some reason, companion needs to update the rules to reflect the new URL. We can preemptively generate new rules based on the new url, but the new node might not be servicing the old content, so we need to be careful about that. Hence, we remove all such rules as offending and begin the process of generating new rules dynamically.
```
BigLep, line 109: Kubo RPC API endpoint

I'm confused by this.  Do we flush the set of rules and rebuild them or do we update the existing rules?  I assume we flush and rebuild, but I'm also not clear on the problem with updating.

docs/MV3.md
```diff
@@ -0,0 +1,114 @@
...
+- Metrics collection becomes harder, as we can no longer reliably determine if the requested resource was serviced by declarative rule introduced by companion. We can only determine if the request was serviceable and we had a rule. We can probably introduce a content script to every page and redirect with a hash value in the URL. That content script can then invoke a companion API to notify what rule id was used to redirect the request. This is a future improvement.
```
BigLep, line 110: This doesn't seem like a high priority to get tighter metrics here, but if there is a backlog item for it, lets link to it.

docs/MV3.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Other Thoughts
+
+Not intercepting requests synchronously has it's quirks, but implementing a dynamic ruleset of possible redirects is even more complicated. Personally we're noticing improvements in second load times as the browser no longer relies on companion to intercept requests and redirect them. Instead declarative rule set allows for redirection to happen at the browser level, which is much faster. This also means that companion is no longer a bottleneck for the browser, which is a good thing.
```
BigLep, line 114: it's → its

What do we mean by "second load times"?

I assume we don't have metrics from how much overhead MV2 blocking incurred?

docs/MV3.md
```diff
@@ -0,0 +1,114 @@
...
+{
+  "action": {
+    "redirect": {
+      "regexSubstitution": "http://localhost:8080\\1"
+    },
+    "type": "redirect"
+  },
+  "condition": {
+    "excludedInitiatorDomains": [],
+    "regexFilter": "^https?\\:\\/\\/ipfs\\.io((?:[^\\.]|$).*)$",
```
BigLep, line 77: Above we talked about "/ipns" urls.  Here our regex passes everything under ipfs.io not just everything under ipfs.io/ipns.  Is that intentional?

BigLep (commented):
Thanks for putting these together.  

Apologies I didn't look before it was merged.

## Diagrams
1. General: Anytime we generate diagrams, we should ideally have the corresponding raw artwork file so it's self-service for someone to made edits.  (This is part of what makes the inline Mermaid diagrams nice since very simple/quick to edit).
2. "Check If Request IPFS"
  - nit: adjust this grammar here?  "Check if IPFS Request"
  - I would expect to see a Yes and No coming out from this box.
  - It would be ideal to link to the logic for "check if IPFS request".  (This is all encapsulated in https://www.npmjs.com/package/is-ipfs right?)
3. For the second diagram, I would expect the arrow to go from "Web Request" to "Observe the Request" (and maybe add the words "non-blocking")

Issue #1235, opened 2023-07-17
Migrate all Redirection tests with MV3 implementations.


whizzzkid:
```[tasklist]
### Tasks
- [x] Migrate test/functional/lib/ipfs-request-dnslink.test.js
- [x] Migrate test/functional/lib/ipfs-request-gateway-redirect.test.js
- [x] Migrate test/functional/lib/ipfs-request-protocol-handlers.test.js
- [x] Optional test/functional/lib/ipfs-request-workarounds.test.js
```

PR #1236, opened 2023-07-18
feat(mv3): Redirection Tests

whizzzkid:
Closes: #1235

In this PR:

- Adding support for MV3 specific interaction in Redirection Cycle.
- Tests:
  - https://github.com/ipfs/ipfs-companion/blob/rc/3.0-mv3/test/functional/lib/ipfs-request-gateway-redirect.test.js added MV3 equivalent
  - https://github.com/ipfs/ipfs-companion/blob/rc/3.0-mv3/test/functional/lib/ipfs-request-protocol-handlers.test.js added MV3 equivalent.
  - https://github.com/ipfs/ipfs-companion/blob/rc/3.0-mv3/test/functional/lib/ipfs-request-dnslink.test.js This only tests if the redirect did not happen and the tab got refreshed after, right now refresh happens momentarily after. https://github.com/ipfs/ipfs-companion/pull/1239 changes that to a message on the bus to tell when to refresh page, the tests will be added along with that.
  - https://github.com/ipfs/ipfs-companion/blob/rc/3.0-mv3/test/functional/lib/ipfs-request-gateway-recover.test.js similar interaction is already testest in redirect testing
  
  
Edit (during the review):
- Simplified setup to run tests both on MV2 and MV3.
- Added better `ensure..` methods to test for both MV2 and MV3 side-effects.
- DRYed and Refactored.

### DIFF
--- a/add-on/src/lib/redirect-handler/blockOrObserve.ts
+++ b/add-on/src/lib/redirect-handler/blockOrObserve.ts
@@ -1,16 +1,24 @@
 import debug from 'debug'
 import browser from 'webextension-polyfill'
 import { CompanionState } from '../../types/companion.js'
+import { IFilter, IRegexFilter, RegexFilter } from './baseRegexFilter.js'
+import { CommonPatternRedirectRegexFilter } from './commonPatternRedirectRegexFilter.js'
+import { NamespaceRedirectRegexFilter } from './namespaceRedirectRegexFilter.js'
+import { SubdomainRedirectRegexFilter } from './subdomainRedirectRegexFilter.js'
 
 // this won't work in webworker context. Needs to be enabled manually
 // https://github.com/debug-js/debug/issues/916
 const log = debug('ipfs-companion:redirect-handler:blockOrObserve')
 log.error = debug('ipfs-companion:redirect-handler:blockOrObserve:error')
 
+export const DEFAULT_NAMESPACES = new Set(['ipfs', 'ipns'])
+
 export const GLOBAL_STATE_CHANGE = 'GLOBAL_STATE_CHANGE'
 export const GLOBAL_STATE_OPTION_CHANGE = 'GLOBAL_STATE_OPTION_CHANGE'
 export const DELETE_RULE_REQUEST = 'DELETE_RULE_REQUEST'
 export const DELETE_RULE_REQUEST_SUCCESS = 'DELETE_RULE_REQUEST_SUCCESS'
+
+// We need to match the rest of the URL, so we can use a wildcard.
 export const RULE_REGEX_ENDING = '((?:[^\\.]|$).*)$'
 
 interface regexFilterMap {
@@ -21,6 +29,7 @@ interface regexFilterMap {
 interface redirectHandlerInput {
   originUrl: string
   redirectUrl: string
+  getPort: (state: CompanionState) => string
 }
 
 type messageToSelfType = typeof GLOBAL_STATE_CHANGE | typeof GLOBAL_STATE_OPTION_CHANGE | typeof DELETE_RULE_REQUEST
@@ -28,6 +37,8 @@ interface messageToSelf {
   type: messageToSelfType
   value?: string | Record<string, unknown>
 }
+
+export const defaultNSRegexStr = `(${[...DEFAULT_NAMESPACES].join('|')})`
 
 // We need to check if the browser supports the declarativeNetRequest API.
 // TODO: replace with check for `Blocking` in `chrome.webRequest.OnBeforeRequestOptions`
@@ -75,11 +86,18 @@ const savedRegexFilters: Map<string, regexFilterMap> = new Map()
 const DEFAULT_LOCAL_RULES: redirectHandlerInput[] = [
   {
     originUrl: 'http://127.0.0.1',
-    redirectUrl: 'http://localhost'
+    redirectUrl: 'http://localhost',
+    getPort: ({ gwURLString }): string => new URL(gwURLString).port
   },
   {
     originUrl: 'http://[::1]',
-    redirectUrl: 'http://localhost'
+    redirectUrl: 'http://localhost',
+    getPort: ({ gwURLString }): string => new URL(gwURLString).port
+  },
+  {
+    originUrl: 'http://localhost',
+    redirectUrl: 'http://127.0.0.1',
+    getPort: ({ apiURL }): string => new URL(apiURL).port
   }
 ]
 
@@ -101,7 +119,7 @@ export function isLocalHost (url: string): boolean {
  * @param str URL string to escape
  * @returns
  */
-function escapeURLRegex (str: string): string {
+export function escapeURLRegex (str: string): string {
   // these characters are allowed in the URL, but not in the regex.
   // eslint-disable-next-line no-useless-escape
   const ALLOWED_CHARS_URL_REGEX = /([:\/\?#\[\]@!$&'\(\ )\*\+,;=\-_\.~])/g
@@ -115,38 +133,24 @@ function escapeURLRegex (str: string): string {
  * @param redirectUrl
  * @returns
  */
-function constructRegexFilter ({ originUrl, redirectUrl }: redirectHandlerInput): {
-  regexSubstitution: string
-  regexFilter: string
-} {
-  // We can traverse the URL from the end, and find the first character that is different.
-  let commonIdx = 1
-  while (commonIdx < Math.min(originUrl.length, redirectUrl.length)) {
-    if (originUrl[originUrl.length - commonIdx] !== redirectUrl[redirectUrl.length - commonIdx]) {
-      break
+function constructRegexFilter ({ originUrl, redirectUrl }: IRegexFilter): IFilter {
+  // the order is very important here, because we want to match the best possible filter.
+  const filtersToTryInOrder: Array<typeof RegexFilter> = [
+    SubdomainRedirectRegexFilter,
+    NamespaceRedirectRegexFilter,
+    CommonPatternRedirectRegexFilter
+  ]
+
+  for (const Filter of filtersToTryInOrder) {
+    const filter = new Filter({ originUrl, redirectUrl })
+    if (filter.canHandle) {
+      return filter.filter
     }
-    commonIdx += 1
   }
 
-  // We can now construct the regex filter and substitution.
-  let regexSubstitution = redirectUrl.slice(0, redirectUrl.length - commonIdx + 1) + '\\1'
-  // We need to escape the characters that are allowed in the URL, but not in the regex.
-  const regexFilterFirst = escapeURLRegex(originUrl.slice(0, originUrl.length - commonIdx + 1))
-  // We need to match the rest of the URL, so we can use a wildcard.
-  const RULE_REGEX_ENDING = '((?:[^\\.]|$).*)$'
-  let regexFilter = `^${regexFilterFirst}${RULE_REGEX_ENDING}`.replace(/https?/ig, 'https?')
-
-  // This method does not parse:
-  // originUrl: "https://awesome.ipfs.io/"
-  // redirectUrl: "http://localhost:8081/ipns/awesome.ipfs.io/"
-  // that ends up with capturing all urls which we do not want.
-  if (regexFilter === `^https?\\:\\/${RULE_REGEX_ENDING}`) {
-    const subdomain = new URL(originUrl).hostname
-    regexFilter = `^https?\\:\\/\\/${escapeURLRegex(subdomain)}${RULE_REGEX_ENDING}`
-    regexSubstitution = regexSubstitution.replace('\\1', `/${subdomain}\\1`)
-  }
-
-  return { regexSubstitution, regexFilter }
+  // this is just to satisfy the compiler, this should never happen. Because CommonPatternRedirectRegexFilter can always
+  // handle.
+  return new CommonPatternRedirectRegexFilter({ originUrl, redirectUrl }).filter
 }
 
 // If the browser supports the declarativeNetRequest API, we can block the request.
@@ -248,10 +252,10 @@ async function reconcileRulesAndRemoveOld (state: CompanionState): Promise<void>
     }
 
     // make sure that the default rules are added.
-    for (const { originUrl, redirectUrl } of DEFAULT_LOCAL_RULES) {
-      const { port } = new URL(state.gwURLString)
-      const regexFilter = `^${escapeURLRegex(`${originUrl}:${port}`)}(.*)$`
-      const regexSubstitution = `${redirectUrl}:${port}\\1`
+    for (const { originUrl, redirectUrl, getPort } of DEFAULT_LOCAL_RULES) {
+      const port = getPort(state)
+      const regexFilter = `^${escapeURLRegex(`${originUrl}:${port}`)}\\/${defaultNSRegexStr}\\/${RULE_REGEX_ENDING}`
+      const regexSubstitution = `${redirectUrl}:${port}/\\1/\\2`
 
       if (!savedRegexFilters.has(regexFilter)) {
         // We need to add the new rule.
--- a/test/functional/lib/ipfs-request-gateway-redirect.test.js
+++ b/test/functional/lib/ipfs-request-gateway-redirect.test.js
@@ -67,8 +67,8 @@ describe(`[${manifestVersion}] gateway-redirect:`, function () {
           modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
           MV2Expectation: 'http://localhost:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest',
           MV3Expectation: {
-            origin: '^https?\\:\\/\\/google\\.com',
-            destination: 'http://localhost:8080'
+            origin: '^https?\\:\\/\\/google\\.com\\/(ipfs|ipns)\\/',
+            destination: 'http://localhost:8080/\\1/\\2'
           }
         })
       })
@@ -137,8 +137,8 @@ describe(`[${manifestVersion}] gateway-redirect:`, function () {
           modifiedRequestCallResp: await modifyRequest.onBeforeRequest(xhrRequest),
           MV2Expectation: 'http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest',
           MV3Expectation: {
-            origin: '^https?\\:\\/\\/google\\.com',
-            destination: 'http://127.0.0.1:8080'
+            origin: '^https?\\:\\/\\/google\\.com\\/(ipfs|ipns)\\/',
+            destination: 'http://127.0.0.1:8080/\\1/\\2'
           }
         })
       })
@@ -150,8 +150,8 @@ describe(`[${manifestVersion}] gateway-redirect:`, function () {
           modifiedRequestCallResp: await modifyRequest.onBeforeRequest(xhrRequest),
           MV2Expectation: 'http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest',
           MV3Expectation: {
-            origin: '^https?\\:\\/\\/google\\.com',
-            destination: 'http://127.0.0.1:8080'
+            origin: '^https?\\:\\/\\/google\\.com\\/(ipfs|ipns)\\/',
+            destination: 'http://127.0.0.1:8080/\\1/\\2'
           }
         })
       })
@@ -163,8 +163,8 @@ describe(`[${manifestVersion}] gateway-redirect:`, function () {
           modifiedRequestCallResp: await modifyRequest.onBeforeRequest(xhrRequest),
           MV2Expectation: 'http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest',
           MV3Expectation: {
-            origin: '^https?\\:\\/\\/google\\.com',
-            destination: 'http://127.0.0.1:8080'
+            origin: '^https?\\:\\/\\/google\\.com\\/(ipfs|ipns)\\/',
+            destination: 'http://127.0.0.1:8080/\\1/\\2'
           }
         })
       })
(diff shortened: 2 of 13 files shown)


add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -24,7 +24,11 @@ interface messageToSelf {
 // We need to check if the browser supports the declarativeNetRequest API.
 // TODO: replace with check for `Blocking` in `chrome.webRequest.OnBeforeRequestOptions`
 // which is currently a bug https://bugs.chromium.org/p/chromium/issues/detail?id=1427952
-export const supportsBlock = !(browser.declarativeNetRequest?.MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES === 5000)
+// this needs to be a function call, because in tests we mock browser.declarativeNetRequest
+// the way sinon ends up stubbing it, it's not directly available in the global scope on import
+// rather it gets replaced dynamically when the module is imported. Which means, we can't
+// just check for the existence of the property, we need to call the browser instance at that point.
+export const supportsBlock = (): boolean => !(browser.declarativeNetRequest?.MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES === 5000)
```
lidel, line 31: nit: this check will break when Google changes limit.

Maybe:

```suggestion
export const noDeclarativeNetRequest = (): boolean => !(typeof browser.declarativeNetRequest === 'function')
```

?

test/functional/lib/ipfs-request-gateway-redirect.test.mv3.js
lidel, line 1: Any reason why this is a separate file?
I would expect  `ipfs-request-gateway-redirect.test.js` to have tests for both MV2 and MV3, we will run into maintenance problems if we don't keep regression tests in sync.

If it is ok, move MV2 expects there and remove `.mv3.js` – wil be way way easier to review. :pray: 

(I did `diff ipfs-request-gateway-redirect.test.js ~/tmp/ipfs-request-gateway-redirect.test.mv3.js` and various tests are not present here, and I dont know if that was intentional, if we test them elsewhere, or what. By having a single file, it easier to ask these questions.  I think it is okay to have some things only tested for MV2 for now until we figure out how to test with MV3, or explicitly mark them as "this makes sense to be only tested for MV2")

test/functional/lib/ipfs-request-protocol-handlers.test.mv3.js
lidel, line 1: Same here, this should be merged into `ipfs-request-protocol-handlers.test.js`

lidel (changes requested):
@whizzzkid thank you, the direction is good, nearly there.

Dropped quick feedback below, ping me when I should make a proper pass on this (I did not look at all tests due to reasons below – will be more efficient to look after these nits are addressed)

test/functional/lib/ipfs-request-gateway-redirect.test.mv3.js
lidel: Hm.. no, we want to keep original MV2 tests as long we support blocking api.

I think having MV3 is separate file is better than not having it, if we don't solve it now, we will pay a higher price in maintenance but that is all. Don't want to block on this, we can figure out merging them later (or maybe it will not be relevant once firefox can switch to declarative API).

As for review or actual tests, I can do manual diff after you simplify code a bit (see note in https://github.com/ipfs/ipfs-companion/pull/1236#discussion_r1269564633) :pray:

SgtPooki: @whizzzkid another way we could do this is via environment variable that is read during the setup.. and we have the same file but just trigger a different browser.global setup, and have the tests pivot on running when that is set.

I can't remember if we tested that once we got the global setup abstracted out and things were working consistently.

test/setup/mocha-setup.js
```diff
@@ -0,0 +1,11 @@
...
+import AbortController from 'abort-controller'
+browser.runtime.id = 'testid'
+global.browser = browser
+global.AbortController = AbortController
+global.chrome = browser
+global.navigator = {
+  clipboard: {
+    writeText: () => {}
+  }
+}
```
SgtPooki, line 11: merge two setup files and wrap execution of either on env-var?

SgtPooki (changes requested):

SgtPooki:
@whizzzkid note that CI tests are emitting an error: https://github.com/ipfs/ipfs-companion/actions/runs/5597804498/job/15162779434?pr=1236#step:11:97

SgtPooki:
Ok. I made some updates, so maintenance of tests should be a lot easier (addressing most of @lidel and my concerns). 

https://github.com/ipfs/ipfs-companion/pull/1247 merges into https://github.com/ipfs/ipfs-companion/pull/1246 which merges into https://github.com/ipfs/ipfs-companion/pull/1236

add-on/src/lib/redirect-handler/blockOrObserve.ts
lidel: oof, given [Google lost interest in MV3](https://groups.google.com/u/0/a/chromium.org/g/chromium-extensions/c/zQ77HkGmK9E) that is not getting fixed any time soon.

Maybe at least remove dependency on specific number?

```suggestion
export const supportsBlock = (): boolean => !(browser.declarativeNetRequest?.MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES > 0)
```

lidel (changes requested):

test/functional/lib/ipfs-request-gateway-redirect.test.js
```diff
@@ -275,43 +460,75 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
         it('should be left untouched if [::1] is used', async function () {
           // https://github.com/ipfs/ipfs-companion/issues/291
           const request = url2request('http://[::1]:5001/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ/')
-          await expectNoRedirect(modifyRequest, request)
+
+          if (isMv3TestingEnabled) {
+          // return this.skip()
+          } else {
+            await expectNoRedirect(modifyRequest, request)
+          }
```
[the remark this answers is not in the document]
SgtPooki: @lidel these were just me getting mv3 testing migrated from the two separate files to the same file, and not wanting to block on fixing all the individual tests that @whizzzkid has more context on.

SgtPooki:
@whizzzkid lmk when this is ready for me to peek at again!

test/functional/lib/ipfs-request-gateway-redirect.test.js
```diff
@@ -216,49 +309,67 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
-          .to.equal(`http://localhost:8080/ipfs/${cid}/`)
+        ensureCallRedirected({
+          modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
+          MV2Expectation: `http://localhost:8080/ipfs/${cid}/`,
+          MV3Expectation: {
+            origin: `^https?\\:\\/\\/${cid}\\.ipfs\\.dweb\\.link`,
+            destination: `http://localhost:8080/ipfs/${cid}`
+          }
+        })
       })
```
SgtPooki, line 334: @lidel @whizzzkid how deep does the maintenance for older versions go? We should be thinking about dropping support for extremely old versions if it's causing any issues. We can also check version numbers at https://github.com/plprobelab/network-measurements/blob/master/reports/2023/calendar-week-29/ipfs/README.md#agent-versions

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -143,104 +105,78 @@ describe('lib/redirect-handler/blockOrObserve', () => {
...
-      })
-      ensureTabRedirected('http://localhost:8081/ipns/docs.ipfs.tech')
-      ensureDeclrativeNetRequetRuleIsAdded({
-        expectedCondition: `^https?\\:\\/\\/docs\\.ipfs\\.tech${LAST_GROUP_REGEX}`,
-        regexSubstitution: 'http://localhost:8081/ipns/docs.ipfs.tech\\1',
-        removedRulesIds: [addRules[0].id]
-      })
-    })
-
-    it('Should remove the old rules when companion is no longer in active state', async () => {
```
SgtPooki, line 215: it looks like this test was removed? was that intentional?

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -143,104 +105,78 @@ describe('lib/redirect-handler/blockOrObserve', () => {
...
+      expect(priority).to.equal(1)
+      expect(action).to.deep.equal({
+        type: 'redirect', redirect: {
+          "regexSubstitution": "http://localhost:8080/ipns/docs.ipfs.tech\\1"
+        }
+      })
+      expect(condition).to.deep.equal(dynamicRulesConditions('^https?\\:\\/\\/docs\\.ipfs\\.tech((?:[^\\.]|$).*)$'))
     })
 
-    it('Should remove the old rule when redirect changes for local gateway', async () => {
```
SgtPooki, line 189: it looks like this test was removed? was that intentional?

SgtPooki (commented):
looks like two tests from blockOrObserve.test.ts were removed? other than that.. it's looking good to me. just some comments and questions.

SgtPooki (approved):

test/functional/lib/ipfs-request-gateway-redirect.test.js
```diff
@@ -216,49 +309,67 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
-        request.responseHeaders = [{ name: 'X-Ipfs-Path', value: fakeXIpfsPathHdrVal }]
-        expect((await modifyRequest.onBeforeRequest(request)).redirectUrl)
-          .to.equal('http://localhost:8080/ipfs/bafybeigfejjsuq5im5c3w3t3krsiytszhfdc4v5myltcg4myv2n2w6jumy/%3Ffilename=test.jpg?arg=val')
+
+        ensureCallRedirected({
+          modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
+          MV2Expectation: 'http://localhost:8080/ipfs/bafybeigfejjsuq5im5c3w3t3krsiytszhfdc4v5myltcg4myv2n2w6jumy/%3Ffilename=test.jpg?arg=val',
+          MV3Expectation: {
+            origin: '^https?\\:\\/\\/bafybeigfejjsuq5im5c3w3t3krsiytszhfdc4v5myltcg4myv2n2w6jumy\\.ipfs\\.dweb\\.link',
+            destination: 'http://localhost:8080/ipfs/bafybeigfejjsuq5im5c3w3t3krsiytszhfdc4v5myltcg4myv2n2w6jumy'
```
lidel, line 357: :warning:  Not blocking this PR, but see https://github.com/ipfs/ipfs-companion/issues/1255 for follow-up work.

test/functional/lib/ipfs-request-gateway-redirect.test.js
```diff
@@ -275,43 +386,63 @@ describe('modifyRequest.onBeforeRequest:', function () {
         // or could produce false-positives such as redirection from localhost:5001/ipfs/path to localhost:8080/ipfs/path
         it('should fix localhost Kubo RPC hostname to IP', async function () {
           const request = url2request('http://localhost:5001/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ/')
-          // expectNoRedirect(modifyRequest, request)
-          expect((await modifyRequest.onBeforeRequest(request)).redirectUrl)
-            .to.equal('http://127.0.0.1:5001/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ/')
+
+          if (!isManifestV3) {
+            // this is set as a default rule in MV3
```
lidel, line 391: Can we add a test for that? `DEFAULT_LOCAL_RULES` seems to be only used in `reconcileRulesAndRemoveOld` so maybe add unit test for that?

(Not blocker, ok to fill issue and do it as followup)

test/functional/lib/ipfs-request-gateway-redirect.test.js
```diff
@@ -275,43 +386,63 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
           const request = url2request('http://0.0.0.0:5001/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ/')
-          expect((await modifyRequest.onBeforeRequest(request)).redirectUrl)
-            .to.equal('http://127.0.0.1:5001/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ/')
+
+          ensureCallRedirected({
+            modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
+            MV2Expectation: 'http://127.0.0.1:5001/ipfs/QmPhnvn747LqwPYMJmQVorMaGbMSgA7mRRoyyZYz3DoZRQ/',
+            MV3Expectation: {
+              origin: '^https?\\:\\/\\/0\\.0\\.0\\.0',
+              destination: 'http://127.0.0.1'
```
lidel, line 419: :warning: iiuc  rules like this are too greedy and will break people's developer environments as they may run other things on different port.

Can we make the rule more specific, and add rule that touches only  the RPC port?

test/functional/lib/ipfs-request-protocol-handlers.test.js
```diff
@@ -61,173 +61,339 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
         it('should be normalized if ipfs://{CID}', async function () {
           const request = url2request('https://dweb.link/ipfs/?uri=ipfs%3A%2F%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%3FargTest%23hashTest')
-          expect((await modifyRequest.onBeforeRequest(request)).redirectUrl).to.equal('https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest')
+
+          ensureCallRedirected({
+            modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
+            MV2Expectation: 'https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest',
+            MV3Expectation: {
+              origin: '^https?\\:\\/\\/dweb\\.link\\/ipfs\\/\\?uri\\=ipfs%3A%2F%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%3FargTest%23',
+              destination: 'https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#'
```
lidel, line 75: Why is `#` included in `destination`? 

I was under the impression that declarative API rules do not include `#fragments`, and that the browser will append `#hashTest`  to the destination on its own.

test/functional/lib/ipfs-request-protocol-handlers.test.js
```diff
@@ -61,173 +61,339 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
         it('should be normalized if ipns://{fqdn}', async function () {
           const request = url2request('https://dweb.link/ipfs/?uri=ipns%3A%2F%2Fipfs.io%3FargTest%23hashTest')
-          expect((await modifyRequest.onBeforeRequest(request)).redirectUrl).to.equal('https://ipfs.io/ipns/ipfs.io?argTest#hashTest')
+
+          ensureCallRedirected({
+            modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
+            MV2Expectation: 'https://ipfs.io/ipns/ipfs.io?argTest#hashTest',
+            MV3Expectation: {
+              origin: '^https?\\:\\/\\/dweb\\.link\\/ipfs\\/\\?uri\\=ipns%3A%2F%2Fipfs\\.io%3FargTest%23',
+              destination: 'https://ipfs.io/ipns/ipfs.io?argTest#'
```
lidel, line 92: :warning:  Not blocking this PR, but see https://github.com/ipfs/ipfs-companion/issues/1254 for follow-up work.

test/functional/lib/ipfs-request-protocol-handlers.test.js
```diff
@@ -61,173 +61,339 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
         it('should be normalized if dweb:/ipfs/{CID}', async function () {
           const request = url2request('https://duckduckgo.com/?q=dweb%3A%2Fipfs%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%3Farg%3Dfoo%26bar%3Dbuzz%23hash&ia=software')
-          expect((await modifyRequest.onBeforeRequest(request)).redirectUrl).to.equal('https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?arg=foo&bar=buzz#hash')
+
+          ensureCallRedirected({
+            modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
+            MV2Expectation: 'https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?arg=foo&bar=buzz#hash',
+            MV3Expectation: {
+              origin: '^https?\\:\\/\\/duckduckgo\\.com\\/\\?q\\=dweb%3A%2Fipfs%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%3Farg%3Dfoo%26bar%3Dbuzz%23hash\\&ia\\=software',
+              destination: 'https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?arg=foo&bar=buzz#hash'
```
lidel, line 279: Test includes `#hash` here but i've seen skipped  `#hashTest` in some previous ones? 
This inconsistency looks like a bug that needs to be fixed in this PR.

- if declarative API takes care of `#fragments` then we should not include them in created rules.
- if it does not handle fragments, we need to preserve them somehow

test/functional/lib/ipfs-request-workarounds.test.js
```diff
@@ -44,8 +53,14 @@ describe('modifyRequest processing', function () {
...
         initiator: 'https://some-website.example.com' // Chromium
       }
-      expect((await modifyRequest.onBeforeRequest(request)).redirectUrl)
-        .to.equal(`http://127.0.0.1:8080/ipfs/${cid}`)
+      ensureCallRedirected({
+        modifiedRequestCallResp: await modifyRequest.onBeforeRequest(request),
+        MV2Expectation: `http://127.0.0.1:8080/ipfs/${cid}`,
+        MV3Expectation: {
+          origin: '^https?\\:\\/\\/ipfs\\.io',
+          destination: 'http://127.0.0.1:8080'
```
lidel, line 61: :warning:  Not blocking this PR, but please see https://github.com/ipfs/ipfs-companion/issues/1253 for follow-up work.

lidel (changes requested):
Thanks, made another pass,  and I think there are only a few  blockers:

- handling of `#fragment` (some rules preserve it, some do not?)
- greedy localhost and 0.0.0.0 rules without port
- cosmetic: improve UX of `ensureCallRedirected`

Details inline.

For the rest, I've filled separate issues for follow-up work that is not blocking this PR.

test/helpers/mv3-test-helper.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param modifyRequest - Request Modifier
+ * @param request - Request to be modified
+ */
+export async function ensureNoRedirect (modifyRequest, request): Promise<void> {
+  await Promise.all([
+    modifyRequest.onBeforeRequest(request),
+    modifyRequest.onHeadersReceived(request)
+  ].map(async (resp) => {
+    await ensureRequestUntouched(await resp)
+  }))
```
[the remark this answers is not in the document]
lidel: `onBeforeRequest` and `onHeadersReceived` are expected to be called on a request in the [specific order](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest). 

I remember we had some edge cases where we woudl detect something in one stage, and act on it in other stage. One I remember is `ignoredRequests`, but there may be more (or we may add them in future). Better to do this sequentially.

test/functional/lib/ipfs-request-protocol-handlers.test.js
lidel: I think this is a bug in test.

The #fragment is not sent to the server when a URL is requested, as it's meant for client-side use, so we should remove it from `origin`:

```suggestion
              origin: '^https?\\:\\/\\/dweb\\.link\\/ipfs\\/\\?uri\\=ipfs%3A%2F%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%3FargTes,
              destination: 'https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest'
```

test/functional/lib/ipfs-request-gateway-redirect.test.js
```diff
@@ -121,22 +132,54 @@ describe('modifyRequest.onBeforeRequest:', function () {
...
         runtime.isFirefox = true
         const xhrRequest = { url: 'https://google.com/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest', type: 'xmlhttprequest', originUrl: 'https://google.com/' }
-        expect((await modifyRequest.onBeforeRequest(xhrRequest)).redirectUrl).to.equal('http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest')
+
+        ensureCallRedirected({
+          modifiedRequestCallResp: await modifyRequest.onBeforeRequest(xhrRequest),
+          MV2Expectation: 'http://127.0.0.1:8080/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR?argTest#hashTest',
+          MV3Expectation: {
+            origin: '^https?\\:\\/\\/google\\.com',
+            destination: 'http://127.0.0.1:8080'
```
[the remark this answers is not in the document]
lidel: Sgtm.

PR #1239, opened 2023-07-19
feat(mv3): Faster Redirects During The First Page Load in main_frame

### LINKED ISSUE #1238: [MV3 Beta Bug] Refreshes on Initial Page Load
Sometime an unnecessary refresh happens during the initial page load. A common symptom is entering a url that can be served over ipns, the page refreshes and nothing shows. This happens because there is a race-condition that tries to reload the page when the page is still loading/redirecting the desired page. This should not happen, instead of refresh a better approach will be to update the url on the given tab to be the destination url.

Edit: The URLs that are being observed and being resolved by DNSLink, show up depending on the resolver speed, in the mean time the page could've loaded over the network, causing a change in URL after.

whizzzkid:
In this PR:

- Fixes #1238 
- Improves the first time load performance for `main_frame` which makes it better than refreshing it.
- Refactored Tests

### DIFF
--- a/test/functional/lib/redirect-handler/blockOrObserve.test.ts
+++ b/test/functional/lib/redirect-handler/blockOrObserve.test.ts
@@ -51,23 +51,32 @@ function ensureTabRedirected (url): void {
  * @param regexSubstitution
  */
 function ensureDeclrativeNetRequetRuleIsAdded ({
+  addRuleLength = 1,
+  callIndex = 0,
   expectedCondition,
   regexSubstitution,
-  removedRulesIds = []
+  removedRulesIds = [],
 }: {
+  addRuleLength?: number
+  callIndex?: number
   expectedCondition: string
   regexSubstitution: string
   removedRulesIds?: number[]
 }): void {
+  if (callIndex < 0) {
+    callIndex = browserMock.declarativeNetRequest.updateDynamicRules.getCalls().length + callIndex
+  }
   expect(browserMock.declarativeNetRequest.updateDynamicRules.called).to.be.true
-  const [{ addRules, removeRuleIds }] = browserMock.declarativeNetRequest.updateDynamicRules.firstCall.args
+  const [{ addRules, removeRuleIds }] = browserMock.declarativeNetRequest.updateDynamicRules.getCall(callIndex).args
   expect(removeRuleIds).to.deep.equal(removedRulesIds)
-  expect(addRules).to.have.lengthOf(1)
-  const [{ id, priority, action, condition }] = addRules
-  expect(id).to.be.a('number')
-  expect(priority).to.equal(1)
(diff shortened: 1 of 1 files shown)


test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -30,9 +30,41 @@ const dynamicRulesConditions = (regexFilter) => ({
...
+
+/**
+ * Ensures that the tab is redirected to the given url on the first request.
+ *
+ * @param url
+ */
+function ensureTabRedirected (url): void {
+  expect(browserMock.tabs.query.called).to.be.true
+  expect(browserMock.tabs.update.called).to.be.true
+  expect(browserMock.tabs.update.firstCall.args).to.deep.equal([TEST_TAB_ID, { url }])
```
SgtPooki, line 44: should we accept a `TEST_TAB_ID` argument? I imagine we will need to test multiple open tabs at some point?

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -30,9 +30,41 @@ const dynamicRulesConditions = (regexFilter) => ({
...
+/**
+ * Ensures that the declarativeNetRequest API is called with the expected rule.
+ * @param expectedCondition
+ * @param regexSubstitution
+ */
+function ensureDeclrativeNetRequetRuleIsAdded ({ expectedCondition, regexSubstitution }): void {
+  expect(browserMock.declarativeNetRequest.updateDynamicRules.called).to.be.true
+  const [{ addRules, removeRuleIds }] = browserMock.declarativeNetRequest.updateDynamicRules.firstCall.args
+  expect(removeRuleIds).to.deep.equal([])
+  expect(addRules).to.have.lengthOf(1)
```
SgtPooki, line 56: is there ever a chance when adding a rule that a rule will also be removed?

Can we add these to the arguments for this helper function?

Also, can we move this helper function outside of this test file?

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -68,6 +100,7 @@ describe('lib/redirect-handler/blockOrObserve', () => {
       sinonSandbox.restore()
       browserMock.tabs.query.resetHistory()
       browserMock.tabs.reload.resetHistory()
+      browserMock.tabs.update.resetHistory()
```
SgtPooki, line 103: if we're using a spy/mock/stub that sinonSandbox is aware of, it should be able to clear them all out for us.

SgtPooki (approved):
Approved, but some suggested improvements.

SgtPooki (approved):

PR #1240, opened 2023-07-20
feat(mv3): ✨ Introduces Redirect Rule Management

whizzzkid:
In this PR:
- lots of refactoring
- Ability to delete individual or all rules.

Final UI:

image

Functionality (UI not updated):

https://github.com/ipfs/ipfs-companion/assets/1895906/35a3d67c-baa1-40f0-ab60-849197a4ce5f

### DIFF
--- a/add-on/src/options/forms/redirect-rule-form.js
+++ b/add-on/src/options/forms/redirect-rule-form.js
@@ -22,8 +22,8 @@ function ruleItem (emit) {
    */
   return function ({ id, origin, target }) {
     return html`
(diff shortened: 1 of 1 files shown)


SgtPooki:
in the demo video, it looks like en-wikipedia is there after you remove ipfs.io, and then when you go to ipfs.io, the en-wikipedia one is gone? why is that?

SgtPooki (changes requested):
some nits and a comment about the demo video have odd results?

SgtPooki (approved):

PR #1256, opened 2023-08-04
fix(rules): Better Redirect Rules

whizzzkid:
fixes: https://github.com/ipfs/ipfs-companion/issues/1254
closes: https://github.com/ipfs/ipfs-companion/issues/1253

Builds on work already in: #1236 

In this PR:

- This was a non-trivial effort in designing a dynamic regular expression generator based on origin and redirect URL patterns.
- This makes the rules less greedy, `https://host/<namespace>/.*` instead of `https://host/.*`
- The rules are also now dynamic `https://<cid>.<namespace>.subdomain.host.tld/<path>` can now map to `http://localhost:8080/<namespace>/<cid>/<path>` just based on regex rules redirect.

Test Runs: https://github.com/ipfs/ipfs-companion/pull/1259
Follow up PR: https://github.com/ipfs/ipfs-companion/pull/1261

### DIFF
--- a/add-on/src/lib/redirect-handler/subdomainRedirectRegexFilter.ts
+++ b/add-on/src/lib/redirect-handler/subdomainRedirectRegexFilter.ts
@@ -1,4 +1,3 @@
-import isIPFS from 'is-ipfs'
 import { IRegexFilter, RegexFilter } from './baseRegexFilter.js'
 import { DEFAULT_NAMESPACES, RULE_REGEX_ENDING, defaultNSRegexStr, escapeURLRegex } from './blockOrObserve.js'
 


add-on/src/types/global.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'is-ipfs' {
```
SgtPooki, line 1: Created https://github.com/ipfs-shipyard/is-ipfs/issues/88

add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -108,6 +113,16 @@ function escapeURLRegex (str: string): string {
...
 
+/**
+ * Compute the namespace from the URL.
+ *
+ * @param url string
+ */
+function computeNamespaceFromUrl (url: string): string {
+  const { pathname } = new URL(url)
+  return (/\/([^/]+)\//i.exec(pathname)?.[1] ?? '').toLowerCase()
+}
```
SgtPooki, line 124: Can we describe a little more what this is doing?

add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -119,6 +134,84 @@ function constructRegexFilter ({ originUrl, redirectUrl }: redirectHandlerInput)
...
+          regexSubstitution = regexSubstitution.replace(pathWithSearch, '/\\3') // replace path
+        } else {
+          regexSubstitution += '\\3'
+        }
+
+        break
+      }
+      // till we find a namespace or CID, we keep adding subdomains to the staticUrl.
+      staticUrl.unshift(subdomainPart as string)
+    }
```
SgtPooki, line 183: can we split this big while chunk out into another function?

add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -119,6 +134,84 @@ function constructRegexFilter ({ originUrl, redirectUrl }: redirectHandlerInput)
...
+  const originNS = computeNamespaceFromUrl(originUrl)
+  if (!DEFAULT_NAMESPACES.has(originNS) && DEFAULT_NAMESPACES.has(redirectNS)) {
+    // A redirect like https://github.com/ipfs/ipfs-companion/issues/1255
+    regexFilter = `^${escapeURLRegex(regexFilter)}`.replace(/https?/ig, 'https?')
+    const origRegexFilter = regexFilter
+
+    const [tld, root, ...subdomain] = originURL.hostname.split('.').reverse()
+    const staticUrl = [root, tld]
+    while (subdomain.length > 0) {
+      const subdomainPart = subdomain.shift()
```
SgtPooki, line 150: I believe `subdomain` is probably type `(string|undefined)[]` and needs a check to confirm it's not null, despite the fact that we confirm length in the while condition. We can explicitly set type of `subdomainPart` to string to avoid `as string` twice below.

Or, A potentially more type-safe approach could be to do something like

```
let subdomainPart = subdomain.shift()
while(subdomainPart != null) {
// ... code
 subdomainPart = subdomain.shift()
}

File: add-on/src/lib/redirect-handler/blockOrObserve.ts [hunk 5]
```diff
@@ -119,6 +134,84 @@ function constructRegexFilter ({ originUrl, redirectUrl }: redirectHandlerInput)
...
+    // A redirect like
+    // https://ipfs.io/ipfs/QmZMxU -> http://localhost:8080/ipfs/QmZMxU
+    const [originFirst, originLast] = originUrl.split(`/${originNS}/`)
+    const defaultNSRegexStr = `(${[...DEFAULT_NAMESPACES].join('|')})`
+    regexFilter = `^${escapeURLRegex(originFirst)}\\/${defaultNSRegexStr}\\/${RULE_REGEX_ENDING}`
+      .replace(/https?/ig, 'https?')
+    regexSubstitution = redirectUrl
+      .replace(`/${redirectNS}/`, '/\\1/')
+      .replace(originLast, '\\2')
+    return { regexSubstitution, regexFilter }
```
SgtPooki, line 212: It would be great to encapsulate the different return values throughout this function so we could name them and break up this function a bit more.

SgtPooki (changes requested):
A few questions and a few suggestions for improvement

lidel (changes requested):

SgtPooki (approved):
good with this given #1261 will be coming on top, but there is one thing i think could be improved in that PR

add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -119,6 +134,84 @@ function constructRegexFilter ({ originUrl, redirectUrl }: redirectHandlerInput)
...
+      if (isIPFS.cid(subdomainPart as string)) {
+        // We didn't find a namespace, but we found a CID
+        // e.g. https://bafybeib3bzis4mejzsnzsb65od3rnv5ffit7vsllratddjkgfgq4wiamqu.on.fleek.co
+        regexFilter = `${commonStaticUrlStart}(.*?)${commonStaticUrlEnd}`
+        regexSubstitution = redirectUrl
+          .replace(subdomainPart as string, '\\1') // replace CID
+          .replace(new RegExp(`${originURL.pathname}?$`), '\\2') // replace path
+
+        break
+      }
```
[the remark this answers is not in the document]
lidel: Yes, if we dont know the namespace, we can't make informed decision about the intention behind the CID, so better to not introduce "guesswork". 
Subdomain gateway convention was created for this very reason, to ensure there is no guesswork.
(For example, if you have CID with libp2p-key codec, should it fetch the Public key, or resolve IPNS?)

add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -119,6 +134,84 @@ function constructRegexFilter ({ originUrl, redirectUrl }: redirectHandlerInput)
...
+      return { regexSubstitution, regexFilter }
+    } else {
+      // we didn't find a valid regexFilter, so we can return the default.
+      regexFilter = originUrl
+    }
+  }
+
+  // if the namespaces are the same, we can generate simpler regex.
+  // The only value that needs special handling is the `uri` param.
+  // TODO: Remove this check, as `uri` param is deprecated.
```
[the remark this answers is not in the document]
lidel: Ah, interesting edge case. The 422 error on  `https://dweb.link/ipfs/?uri=ipfs%3A%2F%2FQmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR%3FargTest%23hashTest` is a  bug in nginx config somewhere in our infra, we track it  n https://github.com/protocol/bifrost-infra/issues/2604

`?uri` is not deprecated, remove the TODO

PR #1260, opened 2023-08-14
feat(mv3): Better Default Rules

whizzzkid:
Closes: https://github.com/ipfs/ipfs-companion/issues/1253

Builds on existing work already in review: https://github.com/ipfs/ipfs-companion/pull/1256

In this PR:

- Introduced robust default rules.
- Added tests for the same.
- Made default localhost rules stricter as per namespace.


lidel (approved):
sgtm, thanks!

Issue #1251, opened 2023-08-02
MV3 Beta 3.0.0.1114: regression in Brave UX


lidel:
## Problem in MV3 Beta 3.0.0.1114

When built-in node is used in Brave:

Companion disables "My Node" button:

> ![2023-08-02_20-11](https://github.com/ipfs/ipfs-companion/assets/157609/63788e2e-7b13-407b-878d-12cacf0b3697)

And redirect happens to `http://` URI instead of native `ipfs://` and `ipns://` :

> ![2023-08-02_20-11_1](https://github.com/ipfs/ipfs-companion/assets/157609/7d44e62e-272f-4c87-99f7-3c640310b7ed)

## Expected behavior 

- Opening WebUI after clicking "My node" should work.
- When user is redirected to native URI when built-in Brave node type is used:
  >  ![2023-08-02_20-12](https://github.com/ipfs/ipfs-companion/assets/157609/a0a3f6ae-7eb0-46e1-9a2a-d6d95f270b98)

lidel:
@whizzzkid thanks for looking into this.  It feels like a blocker for publishing to stable channel, some details, and ideas below.

### On why we may want to fix this before mV3 ships



Brave uses the package we publish to Chrome Web Store, so by shipping MV3 without fixing this, we introduce an unnecessary regression in Brave, which aims to be the state of the art for native handlers like  `ipfs://`.

We would be undoing years of progress: we have a browser which supports native handler, but we make it act like regular Google Chrome or Firefox. We also alienate existing users who had native protocol for years now, and if someone was using something like wallet over it, they may lose access to it unless they type `ipfs://cid` manually.  

Not the end of the world, but imo easier to fix now, 
rather than to deal with fallout and responding to issues here and in brave.

### Potential implementations paths

I don't know if/when a blocking API in MV3 will happen in Brave. 
Not  a stable foundation we should rely upon going forward.

See if we can use declarative API.  
If not,  perhaps we could  fix the address bar  in Brave  in less invasive ways. 

One idea,  introduce an additional listener in Brave that looks at `chrome.webNavigation` events of the top level documents (URL in address bar), and when a resource with `http://` URL of the local gateway is loaded, update the tab and replace it with `ipfs://` equivalent (so only ipfs:// shows in browsing history).  
Bit hacky, but reload will happen so fast end user will not notice. The benefit of this approach is future-proof, and will  work no matter how broken declarative API in MV3 is :)

PR #1270, opened 2023-09-08
fix(brave): Fix Brave UX

whizzzkid:
Fixes: #1251 

In this PR:
- Addine redirect pattern to help with brave specific redirects using `ipfs://` and `ipns://` protocols
- Brave has had it's own regressions and bugs, the redirect gets added to the rules but is not handled correctly in brave.
- This should be the last PR to make it into: https://github.com/ipfs/ipfs-companion/pull/1182

### DIFF
--- a/add-on/src/options/page.js
+++ b/add-on/src/options/page.js
@@ -115,7 +115,7 @@ export default function optionsPage (state, emit) {
   ${resetForm({
     onOptionsReset
   })}
-  ${supportsBlock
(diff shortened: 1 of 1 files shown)


SgtPooki (approved):
I'm good with this but wondering if we should have more future proof naming. Instead of "isBrave" should we  call it "isNativeIpfsProtocolEnabled" ?

Or is there more to it than that?

PR #1273, opened 2023-09-11
feat(mv3): :heavy_plus_sign: Adding hash function to generate predict…

whizzzkid:
Fixes: https://github.com/ipfs/ipfs-companion/pull/1182#discussion_r1270980344

Introduces: npmjs.com/package/fast-hash-code

### DIFF
--- a/add-on/src/lib/redirect-handler/blockOrObserve.ts
+++ b/add-on/src/lib/redirect-handler/blockOrObserve.ts
@@ -1,4 +1,5 @@
 import debug from 'debug'
+import { fastHashCode } from 'fast-hash-code'
 import browser from 'webextension-polyfill'
 import { CompanionState } from '../../types/companion.js'
 import { IFilter, IRegexFilter, RegexFilter } from './baseRegexFilter.js'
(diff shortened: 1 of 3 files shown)


add-on/src/lib/redirect-handler/blockOrObserve.ts
```diff
@@ -281,8 +282,10 @@ function saveAndGenerateRule (
   regexSubstitution: string,
   excludedInitiatorDomains: string[] = []
 ): browser.DeclarativeNetRequest.Rule {
-  // We need to generate a random ID for the rule.
-  const id = Math.floor(Math.random() * 29999)
+  // We need to generate a positive number as an id.
+  const id = fastHashCode(`${regexFilter}:${regexSubstitution}:${excludedInitiatorDomains.join(':')}`, {
```
SgtPooki, line 286: This seems reasonable, but the link provided by `fast-hash-code` to the implementation reasoning is no longer available. It makes me wonder about collisions. `webCrypto.subtle.digest` should do what we want and is available in service workers, is there a reason we didn't use that?

does `id` need to be a number?

SgtPooki (approved):
Approved with concern

Issue #1274, opened 2023-09-13
[MV3 Beta Bugs] browserAction updates broken in Firefox 


lidel:
> @whizzzkid found this while reviewing https://github.com/ipfs/ipfs-companion/pull/1182 (will be filling standalone issues as I go)

Tested  MV3 from https://github.com/ipfs/ipfs-companion/pull/1182 in latest Firefox nightly (119.0a1 (2023-09-13)):

```console
$ npm run dev-build && npm run firefox:nightly
```

Onboarding It works as expected, After I grant permission to companion it connects to RPC just fine, but the browserAction icon remains "offline" due to error below:

> ![2023-09-13_19-38](https://github.com/ipfs/ipfs-companion/assets/157609/cc02f9ba-6133-4b0b-ad29-a9a440e20a19)
> ![2023-09-13_19-36](https://github.com/ipfs/ipfs-companion/assets/157609/f057f6dd-e063-409f-b42e-01b71e9c07e3)

PR #1275, opened 2023-09-14
fix(mv3): FF Browser Action

whizzzkid:
Closes: #1274 

Broken in: #1266

Cause: MV3 APIs accessed in MV2 Scope

Test:
- there is no easy/automated way to test this, I spent some time looking into writing an e2e test, but nope.


lidel (approved):
lgtm, manual test for this is ok. thanks!

PR #1285, opened 2023-09-21
fix(brave): not updating tab to the new redirect url in some cases.

whizzzkid:
Closes: https://github.com/ipfs/ipfs-companion/issues/1283

RCA:
- The timing in brave makes it hard to determine what's happening in the browser.
- When redirecting from brave internal node, companion cannot find the tab containing that URL.
- Checking repeatedly fixes the issue mostly.
- Added tests.

Demo: the double refresh is slightly visible.

https://github.com/ipfs/ipfs-companion/assets/1895906/17254a4d-23cc-4534-85e4-e121879382d4

### DIFF
--- a/test/functional/lib/redirect-handler/blockOrObserve.test.ts
+++ b/test/functional/lib/redirect-handler/blockOrObserve.test.ts
@@ -2,6 +2,7 @@ import { expect } from 'chai';
 import { before, describe, it } from 'mocha';
 import sinon from 'sinon';
 import browserMock from 'sinon-chrome';
+import { MAX_RETRIES_TO_UPDATE_TAB } from './../../../../add-on/src/lib/redirect-handler/blockOrObserve';
 
 import { optionDefaults } from '../../../../add-on/src/lib/options.js';
 import { RULE_REGEX_ENDING, addRuleToDynamicRuleSetGenerator, cleanupRules, isLocalHost } from '../../../../add-on/src/lib/redirect-handler/blockOrObserve';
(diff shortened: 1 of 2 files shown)


lidel (approved):
Usually `setTimeout`  smells fishy, but sometimes this is the only way to get UX we want, especially in Brave when there is  additional IPFS code  in play on the browser side that executes async (like the purple button scenario).

Retrying for half a second when there are no tabs sounds like a sane failsafe:  In the past, I've experienced similar ordering discrepancies between Firefox and Chromium, and they appeared and disappeared between browser releases, so this may also help with smoothing the UX there.

test/functional/lib/redirect-handler/blockOrObserve.test.ts
```diff
@@ -184,6 +183,30 @@ describe('lib/redirect-handler/blockOrObserve', () => {
...
+
+    it('Should wait for the tab to exist before updating tab from originUrl to redirectUrl for the first time', async () => {
+      const clock = sinon.useFakeTimers();
+      browserMock.tabs.query.onCall(0).resolves([])
+      browserMock.tabs.query.onCall(1).resolves([{ id: 40 }])
+      await addRuleToDynamicRuleSet({
+        originUrl: 'https://ipfs.io/ipns/en.wikipedia-on-ipfs.org',
+        redirectUrl: 'http://localhost:8080/ipns/en.wikipedia-on-ipfs.org'
+      })
+      await clock.tickAsync(400)
```
lidel, line 204: nit: could we base this on `MAX_RETRIES_TO_UPDATE_TAB` somehow, so the reader can figure out why we sleep here?

SgtPooki (approved):
Sgtm

PR #1296, opened 2023-09-25
fix(telemetry): :arrow_down: Reduce event syncs

whizzzkid:
5mins -> 60mins


lidel (approved):
lgtm, but if we have a local persistence, is there a reason why we don't bump it even higher, and flush events to courtly only during extension start + once a day?

SgtPooki:
Agree 100% with Lidel on persistence, and also reducing event emission to weekly.

Also, we should confirm in countly that the events from 1 particular user are actually being sent at the interval we expect (could be a bug?)

PR #1302, opened 2023-11-08
fix: ipfs.dns → ipfs.resolve

lidel:
This PR aims to fix webui preload in Kubo 0.24.

Context: `ipfs.dns` has been deprecated for a while, and the way we use it here got broken in Kubo 0.24: https://github.com/ipfs/kubo/pull/10212

### DIFF
--- a/add-on/src/lib/precache.js
+++ b/add-on/src/lib/precache.js
@@ -15,7 +15,7 @@ export async function precache (ipfs, state) {
   try {
     let cid, name
     if (state.useLatestWebUI) { // resolve DNSLink
-      cid = await ipfs.dns('webui.ipfs.io', { recursive: true })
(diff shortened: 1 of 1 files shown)


SgtPooki (approved):
SGTM, were you able to test this with a local build?

lidel:
@SgtPooki  yes, it calls `/api/v0/resolve` correctly

lidel:
@SgtPooki pushed fix, `/api/v0/resolve` is now called with `/ipns` path, and that returns resolved one that is used for preload:

> 

But it does not really matter, even if that failed in Kubo 0.24, the fallback based on pathname detection still works and preloads ok:

>

SgtPooki (approved):

PR #1327, opened 2025-02-27
fix: remove countly

### LINKED ISSUE #1141: Telemetry: click count per menu option
Currently, the telemetry  (#1117 ) has low value.
IIUC only "views" are tracked, and that means only "optiona" and  "quick import" features are tracked (because they own HTML page), but nothing else.   

No way to reason about feature usefullness without having the full picture.
We need to know which features accessible via browser action menu are the mostly used.

If we are gathering telemetry, let's make it worth the effort.
We need to how many times each button was clicked. Every clickable item on the below screenshot should bump its click count:

> Screenshot 2023-01-30 at 12-09-33 ipfs_ipfs-companion Browser extension that simplifies access to IPFS resources on the web

### LINKED ISSUE #1315: Remove countly.ipfs.tech telemetry
This is a placeholder issue: we won't have paid Countly after Feb 18th, 2024
and iiuc should remove / disable code that sends requests to countly.ipts.tech.

There is a proposal in https://github.com/ipfs-shipyard/ignite-metrics/issues/133 to update ignite-metrics to use Plausible instead and then have this repo update to use the new ignite-metrics.

Regardless of the chosen path, the Countly shutdown will happen first, independent of this work.  This means that for a time, metrics will be POSTed to countly but not stored.

cc https://github.com/ipfs/ipfs-webui/issues/2198 https://github.com/ipfs/ipfs-desktop/issues/2743

2color:
## Summary
This PR removes all Countly tracking functionality from the project while preserving user preferences and maintaining a consistent UI (in case we want to add back a different implementation) . The telemetry collection has been completely disabled, with tracking functions replaced by no-op implementations that only log debug messages.

Fixes https://github.com/ipfs/ipfs-companion/issues/1315
Closes #1141

## Changes
- Replaced tracking functions with no-op implementations that only log debug messages
- Preserved original storage keys for backward compatibility with existing user preferences
- Updated the telemetry UI to clearly indicate that tracking has been removed
- Maintained the telemetry toggle in UI 
- Updated documentation to reflect the removal of telemetry
- Removed custom Countly type definitions
- Removed the @ipfs-shipyard/ignite-metrics dependency


lidel (approved):
lgtm, good idea about hiding telemetry section, pushed a fix. 

if looks fine, feel free to merge, then we can ship this to see how long chrome web store publishing takes 

@SgtPooki  do you still have permissions, or should I do the release?
I think @2color  does not have account for that yet

2color:
Can confirm this is working including the hiding of the telemetry section. Thanks @lidel

SgtPooki:
> @SgtPooki do you still have permissions, or should I do the release?

I believe I have permissions. I can start release this monday unless you want me to try to start it today? If started today it would likely be out in FF/chrome tomorrow through tuesday.

lidel:
Let's not ship on Friday, Monday is fine :) :+1:

PR #1339, opened 2025-09-03
fix: skip DNSLink resolution when offline, deduplicate inflights

lidel:
### fix: remove ipfs.io fallback in DNSLink resolution

- prevent privacy DNS leak by not sending requests to external services
- removed hardcoded fallback to https://ipfs.io/ when offline
- return `undefined` when offline to avoid caching false negatives
- closes https://github.com/ipfs/ipfs-companion/issues/1337

### fix: deduplicate concurrent DNSLink lookups for same domain
- added in-flight request tracking with pendingLookups Map
- prevents multiple simultaneous API calls for the same FQDN
- reduces redundant requests from 2-4 down to 1 per domain


SgtPooki (approved):
lgtm, but two comments :)

PR #1350, opened 2025-09-21
fix:add error page for incorrect ipfs:// usage with domain names

### LINKED ISSUE #1316: ipfs://example.com in Chromium should produce error
## Problem

Seems we have some magic in Chromium implementation of `ipfs://` and `ipns://`  that makes 
`ipfs://docs.ipfs.tech` correctly redirect to /ipns/ namespace.

While this sounds like useful, it perpetuates incorrect URIs in the wild: 
I personally had to reach out to well-meaning authors of some blogposts in the community to fix `ipfs//example.com` into `ipns://example.com` because both work in Chromium with Companion, and its easy to miss  one letter difference.

## Solution

Produce intermediate error page which explains that `ipfs://` should be only used for immutable identifiers and mutable pointers like cryptographic IPNS rcords and DNSLink domain names should use `ipns://`.

It should have big green button "Continue to ipns://example.com →"  that allows user to get to the correct destination with minimal disruption.

SAHU-01:
## Description

This PR fixes #1316, by adding an educational error page that appears when users incorrectly use `ipfs://` protocol with domain names instead of `ipns://`.

## Problem

Users often confuse `ipfs://` and `ipns://` protocols, using `ipfs://example.com` when they should use `ipns://example.com`. This creates confusion and perpetuates incorrect URIs in documentation and blog posts.

## Solution  

- Added detection for domain names used with `ipfs://` protocol
- Created an educational error page explaining the difference
- Styled to match IPFS WebUI design system
- Provides "Continue to ipns://..." button for seamless redirection

## Testing

Tested on Chromium with the following URLs:

- [x] `ipfs://example.com` → Shows error page → Redirects to `ipns://example.com`
- [x] `ipfs://docs.ipfs.tech` → Shows error page → Redirects to `ipns://docs.ipfs.tech`  
- [x] `ipfs://ipfs.io` → Shows error page → Redirects to `ipns://ipfs.io`
- [x] `ipfs://QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco` → Works normally (valid CID)

## Screenshots
image
image
image
image

https://github.com/user-attachments/assets/03f3f7b8-1c7b-4155-a4c3-a59631fe9bbc

## Considerations:

- The 500 error for `example.com` is expected as it doesn't have DNSLink records
- The regex error in console doesn't affect functionality, it's a Chromium limitation with complex redirect rules
- I tried matching the theme of exsisting pages, still changes could be made here to match the overall UI

Reviewers: @lidel

### DIFF
--- a/add-on/src/landing-pages/protocol-error/protocol-error.css
+++ b/add-on/src/landing-pages/protocol-error/protocol-error.css
@@ -1,0 +1,196 @@
+body {
+  font-family: 'Inter', -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Helvetica Neue', sans-serif;
+  background: #0b3a53;
+  background: linear-gradient(135deg, #0b3a53 0%, #1d4362 100%);
+  color: #ffffff;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-height: 100vh;
+  margin: 0;
+  padding: 20px;
+}
+
+.container {
+  max-width: 600px;
+  background: rgba(255, 255, 255, 0.05);
+  backdrop-filter: blur(10px);
+  border: 1px solid rgba(105, 196, 205, 0.2);
+  border-radius: 8px;
+  padding: 48px;
+  box-shadow: 0 8px 32px rgba(0, 0, 0, 0.3);
+}
+
+.warning-icon {
+  font-size: 48px;
+  text-align: center;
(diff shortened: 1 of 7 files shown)


lidel:
thank you for this poc
superseded by #1380  (which solves multiple problems and has unified ui)

